PR #8, opened 2018-07-06
Revamp TestingLanguage

kfichter:
Changes:
- Moves contracts to `contracts`
- Small bugfix to `PlasmaRLP`
- Refactors `ChildChain` into a simple append-only chain
- Moves everything else to `plasma_contracts`
- Refactors `TestingLanguage` to account for changes to `ChildChain`
- Refactors root chain test suite


plasma_core/child_chain.py
```diff
@@ -0,0 +1,119 @@
...
+        self.next_child_block = self.child_block_interval
+        self.next_deposit_block = 1
+
+    def add_block(self, block):
+        # Is the block being added to the head?
+        is_next_child_block = block.number == self.next_child_block
+        if is_next_child_block or block.number == self.next_deposit_block:
+            # Validate the block.
+            try:
+                self.__apply_block(block)
```
boolafish: calling `apply_block` actually would have side effect on changing the statement of block/tx. If it changes the input state here but return False might/might not lead to error on who calls this function, and would be super hard to understand what changes the state.
Probably verification and apply should be totally separate functions, and only the `True` case should call apply function. How do you think?

testlang/testlang.py
```diff
@@ -0,0 +1,161 @@
...
+        key = getattr(ethtester, 'k{0}'.format(i))
+        accounts.append(EthereumAccount(address_to_hex(address), key))
+    return accounts
+
+
+class PlasmaExit(object):
+    """Represents a Plasma exit.
+    Attributes:
+        owner (str): Address of the exit's owner.
+        amount (int): How much value is being exited.
```
boolafish: lack `token` in comment

testlang/testlang.py
```diff
@@ -0,0 +1,161 @@
...
+        exit_info = self.root_chain.exits(utxo_id)
+        return PlasmaExit(*exit_info)
+
+    def get_balance(self, account):
+        return self.ethtester.chain.head_state.get_balance(account.address)
+
+    def forward_timestamp(self, amount):
+        self.ethtester.chain.head_state.timestamp += amount
+
+    @property
```
boolafish: Minor: just personal preference to have property to top

kfichter:
@boolafish Thanks for the review! Addressed your comments + added some better documentation while I was at it.

PR #10, opened 2018-07-23
Add ERC20 support

paulperegud:


contracts/RootChain.sol
```diff
@@ -263,10 +282,13 @@ contract RootChain {
         while (exitable_at < block.timestamp) {
             currentExit = exits[utxoPos];
 
-            // FIXME: handle ERC-20 transfer
-            require(address(0) == _token);
+            if (_token == address(0)) {
+                currentExit.owner.transfer(currentExit.amount);
+            }
+            else {
+                require(ERC20(_token).transfer(currentExit.owner, currentExit.amount));
```
paulperegud: Radosław Zagórowicz has pointed out that there is a possibility of re-entrance here. It seems that it can be exploited even for honest token, providing it is implementing ERC223 transfer semantic and calls external contract.

contracts/RootChain.sol
```diff
@@ -135,15 +147,22 @@ contract RootChain {
...
         require(currentDepositBlock < CHILD_BLOCK_INTERVAL);
 
-        bytes32 root = keccak256(msg.sender, address(0), msg.value);
-        uint256 depositBlock = getDepositBlock();
-        childChain[depositBlock] = ChildBlock({
-            root: root,
-            timestamp: block.timestamp
-        });
-        currentDepositBlock = currentDepositBlock.add(1);
+        emitDepositBlock(msg.sender, address(0), msg.value);
```
kfichter: I would rename this function. It looks like a typo of `emit DepositBlock`.

contracts/RootChain.sol
```diff
@@ -135,15 +147,22 @@ contract RootChain {
...
+    }
+
 
-        emit Deposit(msg.sender, depositBlock, address(0), msg.value);
+    /**
+     * @dev Deposits approved amount of ERC20 token. Approve must be called first. Note: does not check if token was added.
```
kfichter: Checking the token was added seems important. Was there a specific reason not to?

contracts/RootChain.sol
kfichter: In this same line - if the `require` statement fails (e.x. token `transfer` function just reverts), then the loop fails forever. We need to be calling the function directly (see https://truffleframework.com/tutorials/testing-for-throws-in-solidity-tests) so a failed transfer doesn't block the loop permanently.

contracts/RootChain.sol
```diff
@@ -105,6 +107,16 @@ contract RootChain {
      * Public Functions
      */
 
+    // @dev Allows anyone to add new token to Plasma chain
+    // @param token The address of the ERC20 token
+    function addToken(address _token)
+        public
+    {
+        require(exitsQueues[_token] == address(0));
```
kfichter: Might as well use `hasToken` here.

contracts/RootChain.sol
```diff
@@ -342,6 +376,28 @@ contract RootChain {
...
+     * @param _amount The amount deposited.
+     */
+    function emitDepositBlock(address _owner, address _token, uint256 _amount)
+        private
+    {
+        bytes32 root = keccak256(_owner, _token, _amount);
+        uint256 depositBlock = getDepositBlock();
+        childChain[depositBlock] = ChildBlock({
+            root: root,
+                    timestamp: block.timestamp
```
kfichter: Small note to fix the indents here.

contracts/RootChain.sol
paulperegud: Token must be added before any exists are being made, otherwise `addExitToQueue` will throw. Checking if token was added here incures additional gas cost, since we can't really remove check from `addExitToQueue`. So question is - does it provide additional safety? I can't find good arguments for that.

paulperegud:
Done. @kfichter please take one more look.

contracts/RootChain.sol
kfichter: Yeah, it's not ideal but it's fine in the short term. There are some solutions for this case (isolating the funds and allowing ordered withdrawals) but they tend to be relatively annoying to implement.

contracts/RootChain.sol
paulperegud: done

contracts/RootChain.sol
paulperegud: done

contracts/RootChain.sol
paulperegud: done

kfichter:
LGTM :+1:

PR #14, opened 2018-08-21
Add support for Python 3.7

kfichter:
Bumps Web3.py version to 4.5.0 for Python 3.7 support. Small change to gitignore to handle multiple `env*` folders.


kfichter:
Merged.

PR #15, opened 2018-08-21
Break out builder for contracts into separate package

paulperegud:
Main idea - to use single building tool for both contract tests and production node.

In PR for use in production plasma chain, also is likely to be used in fee-burner.

package: https://pypi.org/project/py-solc-simple/
source: https://github.com/paulperegud/py_solc_simple/

### DIFF
--- a/plasma_core/utils/deployer.py
+++ b/plasma_core/utils/deployer.py
@@ -1,102 +1,12 @@
-import json
-import os
-from solc import compile_standard
 from web3.contract import ConciseContract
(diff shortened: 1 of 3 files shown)


boolafish:
shouldn't the builder package be maintained under omisego as well?

paulperegud:
Good point.

Repo moved here: https://github.com/omisego/py-solc-simple

DavidKnott:
@paulperegud LGTM

PR #19, opened 2018-08-28
Rework finalize exits

### LINKED ISSUE #18: finalizeExits does not work for long queues
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior
`finalizeExits` must be provided with enough gas to clear the queue. Queue can grow long enough so clearing it will not fit in block gas limit. This is an attack vector.

## Expected Behavior
Being able to clear only part of the queue would be nice.

## Suggested Fix A (born in discussion with @pdobacz )

One idea is to have finalizeExits proceed the following way:
1/ finalize the first exit in the queue
2/ calculate the gas used for (1/)
3/ goto 1/ only if remaining gas allowance is > than result of (2/)

A (probably) necessary variation is to keep track of the maximum gas paid for a single exit finalized, and gracefully stop processing exits, in case when there are still exits in the queue, but there isn't enough gas for the pesimistic scenario.

## Suggested Fix B (born in discussion with @pdobacz )

Alternatively this could be done as follows. We modify API:

`finalizeExits(uint256 _top_utxoPos, uint256 _exits_to_process)`

1/ `finalizeExits` checks if passed `_top_utxoPos` is the one on top of the queue
2/ loop processes only `_exits_to_process` exits
3/ sender does local gas estimation and attaches more than enough gas to execute those particular exits

`_top_utxoPos` check makes sure that in case of the tx race, gas losses will be minimal. Explicit number of exits to process simplifies gas estimation.

Drawback of this solution is that gas estimation is just that. Changed state of the token might enfluence amount of gas needed to perform the transfer (possible even for normal tokens because of gas refund mechanic), leading to loss of all of the gas attached to tx. This risk is minimisied by attaching more gas than needed to tx.

On the bright side - there are no reasons to suspect that finalizeExits will be a function called all of the time. Cost of not calling it is a time cost of money. Races should be rare and the one losing the race saves money, since somebody else paid for processing part of the queue.

paulperegud:
Fixes #18 . Implements "Suggested solution B".


kfichter:
I'm still skeptical of the utility of `_topUtxoPos`. It seems like it might necessitate significant coordination to clear the queue in a mass exit scenario. Seems easier to let users spam the queue by default. Maybe we can get the best of both worlds and have `_topUtxoPos == 0` skip the check.

paulperegud:
Will add `_topUtxoPos == 0` skip tomorrow.

paulperegud:
Just a note - in case of mass exit there will be lots of calls to startExit and way smaller amount of finalizeExits calls. Assuming our protocol works correctly, money in mature exits are safe and cost of not calling finalizeExits is time cost of money.

tests/contracts/root_chain/test_process_exits.py
```diff
@@ -0,0 +1,101 @@
...
+    testlang.finalize_exits(token.address, spend_id, 100)
+
+    plasma_exit = testlang.get_standard_exit(spend_id)
+    assert plasma_exit.token == encode_hex(token.address)
+    assert plasma_exit.owner == NULL_ADDRESS_HEX
+    assert standard_exit.amount == amount
+    assert token.balanceOf(owner.address) == pre_balance + amount
+
+
+def test_finalize_exits_partial_queue_processing(testlang):
```
pdobacz: on the paranoid side of things you could also test:
 - behavior of `topUtxoPos` and `count` when the queue is empty, esp. when `topUtxoPos == 0`
 - that short-circuiting the process using `topUtxoPos` actually saves gas, so that there is a require not an assert there (this one not sure)
 - what if `count` is larger than size of priority queue
 - `count` is zero
 - `count` is > 1

paulperegud: :+1:

paulperegud:
Things to test:
* finalizeExits(_, _, 0) - how it handles the gas
* check how finalizeExits on empty queue handles the gas

contracts/RootChain.sol
```diff
@@ -271,15 +271,17 @@ contract RootChain {
...
      */
-    function finalizeExits(address _token)
+    function finalizeExits(address _token, uint256 _topUtxoPos, uint256 _exitsToProcess)
         public
     {
         uint256 utxoPos;
         uint256 exitable_at;
+        uint256 _exitsLeft = _exitsToProcess;
         (utxoPos, exitable_at) = getNextExit(_token);
+        require(_topUtxoPos == utxoPos || _topUtxoPos == 0);
```
kfichter: May be useful to add some clarifying comments here if not planned for later.

kfichter:
Rest LGTM, we should probably have some cleanup later.

contracts/RootChain.sol
paulperegud: Added explanation in function docs.

paulperegud:
by @pgebal :
> add a test for a failure of exiting utxo that is already exited (finalized)

PR #20, opened 2018-08-28
tests: priority queue gas costs

paulperegud:
Add test that checks priority queue gas cost behavior for 3 scenarios: increasing inserts, decreasing inserts and shuffled inserts.

Depends on #17


paulperegud:
Those tests will be incorporated into #17

PR #24, opened 2018-09-28
On finalizeExit - check if exit was successfully challenged

paulperegud:
Challenging the exit does not remove it from priority queue. Don't send ETH or token when such challenged exit is being processed. @kbak

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -285,14 +285,17 @@ contract RootChain {
             currentExit = exits[utxoPos];
 
             queue.delMin();
-            delete exits[utxoPos].owner;
 
-            if (_token == address(0)) {
-                currentExit.owner.transfer(currentExit.amount);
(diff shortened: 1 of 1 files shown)


contracts/RootChain.sol
```diff
@@ -285,14 +285,17 @@ contract RootChain {
...
+            // Send funds only if exit was not successfully challenged.
+            if (exits[utxoPos].owner != address(0)) {
+                if (_token == address(0)) {
+                    currentExit.owner.transfer(currentExit.amount);
+                }
+                else {
+                    require(ERC20(_token).transfer(currentExit.owner, currentExit.amount));
+                }
             }
+            delete exits[utxoPos].owner;
```
kfichter: Does this cost gas if `exits[utxoPos].owner == address(0)`? If so, maybe makes sense to move inside if statement.

paulperegud:
Closing, since dc-audit branch will no longer be used.

PR #26, opened 2018-10-04
tests for some transitions between states of the exit

### LINKED ISSUE #23: add more tests
From @pdobacz:

> a batch of longer-running tests that test whether gas is always under control for submitBlock, startExit, challengeExit and finalizeExit, no matter how much we put into the contract, like we did for HonteStaking.sol

> Another, slightly paranoid thing I'd consider is adding assertions on pre/post-conditions to actions executed using testlang, like "if a successful deposit is done, then contract's balance increases, depositor's - decreases", "if a successful finalizeExit is done, then balances are updated properly". Likewise, I'd assert proper event firing in post conditions. This is something we did for HonteStaking too.

From @pgebal:
> fix `leaves=[]` in FixedMerkle
> ~add a test for a failure of exiting utxo that is already exited (finalized)~ will do in #19

paulperegud:
closes #23

### DIFF
--- a/testlang/testlang.py
+++ b/testlang/testlang.py
@@ -113,7 +113,10 @@ class TestingLanguage(object):
                                  NULL_ADDRESS, 0)
 
         blknum = self.root_chain.getDepositBlock()
+        pre_balance = self.get_balance(self.root_chain)
         self.root_chain.deposit(value=amount, sender=owner.key)
+        balance = self.get_balance(self.root_chain)
+        assert balance == pre_balance + amount
 
         block = Block(transaction_set=[deposit_tx], number=blknum)
         self.child_chain.add_block(block)
@@ -142,7 +145,10 @@ class TestingLanguage(object):
         token.approve(self.root_chain.address, amount, sender=owner.key)
         self.ethtester.chain.mine()
         blknum = self.root_chain.getDepositBlock()
+        pre_balance = self.get_balance(self.root_chain, token)
         self.root_chain.depositFrom(owner.address, token.address, amount, sender=owner.key)
+        balance = self.get_balance(self.root_chain, token)
+        assert balance == pre_balance + amount
 
         block = Block(transaction_set=[deposit_tx], number=blknum)
         self.child_chain.add_block(block)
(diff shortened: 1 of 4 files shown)


tests/contracts/root_chain/test_process_exits.py
```diff
@@ -150,15 +150,28 @@ def test_finalize_exits_tx_race(testlang):
...
+    testlang.start_standard_exit(utxo4.owner, utxo4.spend_id)
+
+    testlang.forward_timestamp(2 * WEEK + 1)
+    testlang.finalize_exits(NULL_ADDRESS, utxo1.spend_id, 1)
+
     testlang.finalize_exits(NULL_ADDRESS, utxo2.spend_id, 3)
-    gas2 = testlang.ethtester.chain.last_gas_used()
-    assert gas1 < 4000
-    assert gas2 > 30000
+    three_exits_gas = testlang.ethtester.chain.last_gas_used()
```
DavidKnott: @paulperegud Doesn't `last_gas_used` only get the gas used from the last transaction as opposed to the last three exits?

testlang/testlang.py
```diff
@@ -338,17 +345,24 @@ def get_standard_exit(self, utxo_id):
...
             int: The account's balance.
         """
-
-        return self.ethtester.chain.head_state.get_balance(account.address)
+        if token == NULL_ADDRESS:
+            return self.ethtester.chain.head_state.get_balance(account.address)
+        if hasattr(token, "balanceOf"):
+            return token.balanceOf(account.address)
+        else:
+            token_contract = conftest.watch_contract(self.ethtester, 'MintableToken', token)
```
DavidKnott: What's special about `MintableToken` does it not have `balanceOf`?

paulperegud: This is misleading. Token can be an address or a wrapped contract.

tests/contracts/root_chain/test_process_exits.py
paulperegud: Those three exits are being done as single Ethereum transaction. See third parameter in finalize_exits call.

testlang/testlang.py
paulperegud: Would love to be able to change that.

PR #27, opened 2018-10-05
Add and test plasma core

DavidKnott:
Adds `PlasmaCore.sol` which will be used by More Viable Plasma.
`transaction_v2.py` will eventually replace `transaction.py` but it'll be a breaking change so it'll be done in another PR.
Tests `PlasmaCore.sol`
@paulperegud

### DIFF
--- a/contracts/PlasmaCore.sol
+++ b/contracts/PlasmaCore.sol
@@ -1,0 +1,197 @@
+pragma solidity ^0.4.0;
+
+import "./ByteUtils.sol";
+import "./RLP.sol";
+
+
+/**
+ * @title PlasmaCore
+ * @dev Utilities for working with and decoding Plasma MVP transactions.
+ */
+library PlasmaCore {
+    using ByteUtils for bytes;
+    using RLP for bytes;
+    using RLP for RLP.RLPItem;
+
+
+    /*
+     * Storage
+     */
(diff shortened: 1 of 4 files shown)


contracts/PlasmaCore.sol
```diff
@@ -0,0 +1,197 @@
...
+    struct TransactionInput {
+        uint256 blknum;
+        uint256 txindex;
+        uint256 oindex;
+    }
+
+    struct TransactionOutput {
+        address owner;
+        uint256 amount;
+    }
```
paulperegud: Shouldn't we add `address token;` here to cover ERC20 usage?

plasma_core/transaction_v2.py
```diff
@@ -0,0 +1,88 @@
...
+        self.blknum = blknum
+        self.txindex = txindex
+        self.oindex = oindex
+
+    @property
+    def identifier(self):
+        return encode_utxo_id(self.blknum, self.txindex, self.oindex)
+
+
+class TransactionOutput(rlp.Serializable):
```
paulperegud: Just a side note - pyRLP has changed its API in 1.0.0, breaking things. Should we merge this one first or move to pyRLP 1.0.0 first? What do you prefer?

PR #28, opened 2018-10-09
Switch to rlp 1.0.0

paulperegud:
This switch will be required if we are to stay with pyethereum tester. Alternatively, we can just go with Ganache and ignore pyrlp 1.0.0.

### DIFF
--- a/plasma_core/block.py
+++ b/plasma_core/block.py
@@ -1,5 +1,5 @@
-import rlp
-from rlp.sedes import binary, CountableList, big_endian_int
+from rlp.sedes import CountableList, big_endian_int
+from rlp.sedes.serializable import Serializable
 from ethereum import utils
 from plasma_core.utils.signatures import sign, get_signer
 from plasma_core.utils.merkle.fixed_merkle import FixedMerkle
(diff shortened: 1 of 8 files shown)


plasma_core/transaction.py
```diff
@@ -1,11 +1,31 @@
 import rlp
 from rlp.sedes import big_endian_int, binary
+from rlp.sedes.serializable import Serializable
 from ethereum import utils
 from plasma_core.constants import NULL_SIGNATURE
 from plasma_core.utils.signatures import sign, get_signer
 
 
-class Transaction(rlp.Serializable):
+class _UnsignedTransaction(Serializable):
```
DavidKnott: Why the underscore in the class name?

plasma_core/block.py
```diff
@@ -1,20 +1,21 @@
...
 from ethereum import utils
 from plasma_core.utils.signatures import sign, get_signer
 from plasma_core.utils.merkle.fixed_merkle import FixedMerkle
 from plasma_core.transaction import Transaction
 from plasma_core.constants import NULL_SIGNATURE
 
 
-class Block(rlp.Serializable):
-
+class UnsignedBlockRLP(Serializable):
```
DavidKnott: I think we'll need an RLP with block and and signatures down the road, what do you think of creating the classes `Block` and `SignedBlock` and having SignedBlock inherit from `Block`?

setup.py
```diff
@@ -24,9 +24,9 @@
     keywords='plasma contracts ethereum development solidity',
     packages=find_packages(exclude=['contrib', 'docs', 'tests']),
     install_requires=[
-        'ethereum==2.3.0',
-        'web3==4.5.0',
-        'rlp==0.6.0',
+        'ethereum==2.3.2',
+        'web3==4.7.2',
+        'rlp==1.0.3',
```
DavidKnott: Thanks for doing this!

PR #29, opened 2018-10-09
add long running chain test

paulperegud:
To run it pass `--runslow` flag


paulperegud:
Going to add double-spends and challenges to this test.

PR #30, opened 2018-10-16
prevent exploitation of `deposit` and `depositFrom` via re-entrance attack

paulperegud:
Otherwise attacking token can be used to produce deposit block with blknum ending in 000. This can be exploited by attacker to create a situation when operator is spending the deposit which later disappears. Depending on implementation, this has a potential of invalidating of chain in eyes of client nodes.

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -394,6 +394,12 @@ contract RootChain {
     function writeDepositBlock(address _owner, address _token, uint256 _amount)
         private
     {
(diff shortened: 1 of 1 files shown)


paulperegud:
Exact scenario is in #161138274

kfichter:
Interesting. This seems like the correct fix assuming 999 deposits per block, although I guess you could also "fix" by increasing the number of intermediate deposit blocks to some number high enough that this will never happen.

LGTM :+1:

PR #31, opened 2018-10-17
Drop confirmation sigs

paulperegud:
This PR removes confirmation signatures from contract and tests. Some remains of confirmation signatures might still be in testlang and core_plasma.

We need this functionality since child chain runs without confirmation signatures.


contracts/Validate.sol
```diff
@@ -9,25 +9,20 @@ import "./ECRecovery.sol";
...
+        require(sigs.length % 65 == 0 && sigs.length <= 130);
 
-        bool check1 = true;
-        bool check2 = true;
-
-        check1 = ECRecovery.recover(txHash, sig1) == ECRecovery.recover(confirmationHash, confSig1);
-        if (blknum2 > 0) {
-            bytes memory confSig2 = ByteUtils.slice(sigs, 195, 65);
-            check2 = ECRecovery.recover(txHash, sig2) == ECRecovery.recover(confirmationHash, confSig2);
+        if (oindex > 0) {
```
pdobacz: wouldn't more explicit `if (oindex == 1) else if (oindex == 0) else assert false` consition sequence be more defensive?

contracts/RootChain.sol
```diff
@@ -205,7 +205,7 @@ contract RootChain {
      * @param _utxoPos The position of the exiting utxo in the format of blknum * 1000000000 + index * 10000 + oindex.
      * @param _txBytes The transaction being exited in RLP bytes format.
      * @param _proof Proof of the exiting transactions inclusion for the block specified by utxoPos.
-     * @param _sigs Both transaction signatures and confirmations signatures used to verify that the exiting transaction has been confirmed.
+     * @param _sigs Transaction signatures are needed to check tx inclusion.
```
kfichter: I think it's fine to remove `_sigs` entirely, assuming we also remove it from the Merkle hash: 

`bytes32 merkleHash = keccak256(keccak256(_txBytes), ByteUtils.slice(_sigs, 0, 130));`
=>
`bytes32 merkleHash = keccak256(_txBytes);`

Note that this also means making a change in the client to account for this. I believe only reason `_sigs` was part of the Merkle hash anyway was to account for confirmation signatures.

contracts/RootChain.sol
```diff
@@ -239,28 +238,29 @@ contract RootChain {
...
-     * @param _confirmationSig The confirmation signature for the transaction used to challenge.
      */
     function challengeExit(
         uint256 _cUtxoPos,
         uint256 _eUtxoIndex,
         bytes _txBytes,
         bytes _proof,
-        bytes _sigs,
-        bytes _confirmationSig
+        bytes _sigs
```
kfichter: Seems easier to have the challenger provide a single signature and check that `ECRecovery.recover(keccak256(_txBytes), _sig) == owner)`. Gets rid of a lot of extra logic about the `oindex` and removes the need for `Validate.checkSigs` entirely.

contracts/RootChain.sol
paulperegud: This change is mostly a stop-gap solution before more-vp. I would prefer to avoid unnecessary work now.

contracts/RootChain.sol
paulperegud: This would require changes in client. Lets punt this to more-vp stage.

paulperegud:
Since you haven't find any mistakes, I'm rebasing and merging.

PR #34, opened 2018-10-20
[WIP] Identify use of implicit Solidity behavior when accessing unset values

saurabh-srivastava:
[ This PR is a WIP. Not ready for merge yet. ]

Use of unset memory is dangerous. To ensure future maintainability, lets add explicit comments where such functionality is critical to operation.

Examples include:
* Use of unset array locations -- reverts. E.g., `getMin` of PriorityQueue reverts on empty queue. `finalizeExits` in RootChain.sol relies on that to abort early.
* Solidity maps are complete, and return 0-initialized when unknown keys are accessed. Used to check emptiness in `exits` map in RootChain.sol.

### DIFF
--- a/contracts/PriorityQueue.sol
+++ b/contracts/PriorityQueue.sol
@@ -58,6 +58,10 @@ contract PriorityQueue {
         }
     }
 
+    /**
+     * @dev Determines the next exit to be processed.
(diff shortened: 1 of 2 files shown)


boolafish:
been out of date. Also, we've moved to ALD.

PR #38, opened 2018-11-09
Rework priority queue as library

paulperegud:
Our main contract no longer deploys - it reaches gas limit. Breaking up PriorityQueue into library that is deployed independently will help decrease RootChain's deployment gas cost.

TODO: actually deploy library in separate transaction. This probably requires changes in our building process.


kfichter:
LGTM. Good call.

PR #39, opened 2018-11-09
add proper long running test

paulperegud:
This test can run overnight (16 hours), doing 2000 iterations of deposits, spends, and exits. See readme for information on how to run it.

### DIFF
--- a/tests/contracts/root_chain/test_long_run.py
+++ b/tests/contracts/root_chain/test_long_run.py
@@ -13,7 +13,7 @@ def pick(left):
     return (subtract(left, [item]), item)
 
 
-@pytest.mark.slow()
+# @pytest.mark.slow()
 def test_slow(testlang):
     utxos = []
     random.seed(a=0)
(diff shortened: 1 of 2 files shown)


tests/contracts/root_chain/test_long_run.py
```diff
@@ -26,46 +26,61 @@ def test_slow(testlang):
...
 
+        print("done exiting")
         testlang.ethtester.chain.mine()
         # 4. attempt to finalize
         testlang.finalize_exits(NULL_ADDRESS, 0, 3)
         max_gas = max(max_gas, testlang.ethtester.chain.last_gas_used())
         print("max_gas is {}".format(max_gas))
+        print("done finalizing")
 
     assert max_gas < 200000
```
pdobacz: maybe we should track `max_gas` for every contract function separately, assert separately and have tighter limit. So that excessive gas usage in a cheaper kind of transaction doesn't get silenced by `200000` being the limit suitable for an expensive one.

paulperegud:
Closing for now. Branch stays until we decide on what to do with amnesiac operation of changing timestamp on a block.

PR #45, opened 2018-12-10
tests: require explicit token in outputs

paulperegud:
Otherwise amount could be passed to contract as token. That introduced hard
to debug errors in test execution. Also, explicit > implicit.


paulperegud:
Merged as a part of #46, closing.

Issue #52, opened 2019-01-02
function respondToNonCanonicalChallenge should emit an event


pik694:
function 'respondToNonCanonicalChallenge' should emit an event of some type so observing processes could react

paulperegud:
Good catch! Will you be willing to do a PR?

pik694:
Then get prepared to do a CR :wink:

PR #53, opened 2019-01-03
feat: add missing event to respondToNonCanonicalChallenge function

pik694:
solves #52

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -102,6 +102,14 @@ contract RootChain {
         address token
     );
 
+    event ExitFinalized(
+        uint256 indexed utxoPos
+    );
+
+    event ExitChallenged(
+        uint256 indexed utxoPos
+    );
+
     event InFlightExitStarted(
         address indexed initiator,
         bytes32 txHash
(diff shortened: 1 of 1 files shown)


contracts/RootChain.sol
```diff
@@ -119,6 +119,12 @@ contract RootChain {
         uint256 challengeTxPosition
     );
 
+    event InFlightExitChallengeResponded(
+        address indexed challenger,
```
paulperegud: `indexed` marker is not needed for `challenger` field.

contracts/RootChain.sol
```diff
@@ -608,6 +614,9 @@ contract RootChain {
 
         // Reset the flag so only the outputs are exitable.
         inFlightExit.exitStartTimestamp = clearFlag(inFlightExit.exitStartTimestamp);
+
+        emit InFlightExitChallengeResponded(msg.sender, keccak256(_inFlightTx), _inFlightTxId);
```
paulperegud: `_inFlightTxId` is not a best name for value that describes tx position.

PR #57, opened 2019-01-25
bump `requests` to 2.20.0

paulperegud:
Reason: https://nvd.nist.gov/vuln/detail/CVE-2018-18074

Not really applicable, but the cost of the fix is low.


paulperegud:
Fixed in https://github.com/omisego/plasma-contracts/pull/62, closing.

PR #59, opened 2019-01-28
style: provide consistent names

pik694:
outputId is now UTXO position, which makes a clear distinction
between, what was previously called,  outputId and outputIndex


pik694:
Just added one last occurrence of `transactionId`

PR #62, opened 2019-02-12
Omg 242 metadata field in the transaction

pthomalla:


tests/contracts/rlp/test_plasma_core.py
```diff
@@ -32,18 +32,6 @@ def test_get_output(plasma_core_test):
     assert plasma_core_test.getOutput(tx.encoded, 1) == [null, null, 0]
 
 
-def test_decode_mallability(testlang, plasma_core_test):
```
paulperegud: Why removing this test? It checks if one can add multiple IFEs using the same correct transaction.

testlang/testlang.py
```diff
@@ -176,6 +176,9 @@ def spend_utxo(self, input_ids, keys, outputs=[], force_invalid=False):
 
     def start_standard_exit(self, output_id, key, bond=None):
         output_tx = self.child_chain.get_transaction(output_id)
+        self.start_standard_exit_transaction(output_id, output_tx, key, bond)
+
+    def start_standard_exit_transaction(self, output_id, output_tx, key, bond=None):
```
paulperegud: start_standard_exit_with_tx_body?

PR #72, opened 2019-03-06
bug: don't emit ExitFinalized for challenged exits

### LINKED ISSUE #69: Don't emit ExitFinalized() event for challenged exits in finalization loop
Successfully challenged exits stay in priority queue (since it does not support random access deletion) and are cleared in `finalizeExits` loop. Event should be emitted if and only if funds are being transferred.

paulperegud:
fixes #69

### DIFF
--- a/tests/contracts/root_chain/test_process_exits.py
+++ b/tests/contracts/root_chain/test_process_exits.py
@@ -274,11 +274,26 @@ def test_finalize_challenged_exit_will_not_send_funds(testlang):
     testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
 
     pre_balance = testlang.get_balance(testlang.root_chain)
+    testlang.process_exits(NULL_ADDRESS, 0, 1, sender=finalizer.key)
+    post_balance = testlang.get_balance(testlang.root_chain)
+    assert post_balance == pre_balance
+
+
+def test_finalize_challenged_exit_does_not_emit_events(testlang):
+    owner, finalizer, amount = testlang.accounts[0], testlang.accounts[0], 100
+    deposit_id = testlang.deposit(owner, amount)
+    spend_id = testlang.spend_utxo([deposit_id], [owner.key], [(owner.address, NULL_ADDRESS, 100)])
+
+    testlang.start_standard_exit(spend_id, owner.key)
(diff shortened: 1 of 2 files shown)


tests/contracts/root_chain/test_process_exits.py
```diff
@@ -274,7 +274,9 @@ def test_finalize_challenged_exit_will_not_send_funds(testlang):
     testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
 
     pre_balance = testlang.get_balance(testlang.root_chain)
+    testlang.flush_events()
     testlang.process_exits(NULL_ADDRESS, 0, 1, sender=finalizer.key)
+    assert [] == testlang.flush_events()
```
boolafish: I'd like to challenge a bit here: Is the logic of emitting Event really something we'd like to test? (well...I can understand there are cases that clients need to rely on Event to get the data they need then this might be important, not really sure about the case here though?)

However, if we like to test this, might be easier for people to understand if we test that feature on a separate test.
eg. `test_finalize_challenged_exit_will_not_emit_event`. It is easy to read this test in this PR but might be really hard for people to understand why are we asserting on empty event in the future.

paulperegud: +1

tests/contracts/root_chain/test_process_exits.py
```diff
@@ -279,6 +279,24 @@ def test_finalize_challenged_exit_will_not_send_funds(testlang):
...
+def test_finalize_challenged_exit_does_not_emit_events(testlang):
+    owner, finalizer, amount = testlang.accounts[0], testlang.accounts[0], 100
+    deposit_id = testlang.deposit(owner, amount)
+    spend_id = testlang.spend_utxo([deposit_id], [owner.key], [(owner.address, NULL_ADDRESS, 100)])
+
+    testlang.start_standard_exit(spend_id, owner.key)
+    doublespend_id = testlang.spend_utxo([spend_id], [owner.key], [(owner.address, NULL_ADDRESS, 100)])
+
+    testlang.challenge_standard_exit(spend_id, doublespend_id)
+    assert testlang.get_standard_exit(spend_id) == [NULL_ADDRESS_HEX, NULL_ADDRESS_HEX, 0]
```
boolafish: Minor: I personally would try to only assert what I want to test. But not really a blocker of this merge.

Issue #73, opened 2019-03-07
Processing in-flight exits that are not piggybacked returns with error


pgebal:
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior
When there is an in-flight exit that is not piggybacked by anyone, processExits function call returns with error.

## Steps to Reproduce

1. Checkout ce4a23d75b8c5f9cc4526f50bde4968908fc258d
2. Comment out line 200 in test_start_in_flight_exit.py
3. Run tests

boolafish:
It seems like the reason is that in `startInflightExit` we didn't put the in-flight-exit into the queue.
`_enqueueExit(...)` is only called on piggyback.

paulperegud:
> It seems like the reason is that in `startInflightExit` we didn't put the in-flight-exit into the queue.
> `_enqueueExit(...)` is only called on piggyback.

This is behavior as expected. IFE may contain many tokens and token queues need to be processed independently because of security reasons (attacking tokens, badly implemented tokens).

PR #74, opened 2019-03-08
fix: process ife without piggybacked would fail

boolafish:
### Note
This commit fixes the bug that an in-flight exit without piggybacked would
cause "processExits" to fail. This is because in in-flight exits, they are
put into priority queue only when they are piggybacked. And "processExits"
would crash on an empty queue.

Fix the issue by enabling "processExits" on an empty queue.

- issue: https://github.com/omisego/plasma-contracts/issues/73

### Test
Followed the `reproduce` step from the issue. Remove the piggyback and test works.

### Discussion
1. For the issue #73, It might possibly be a "feature" instead of "bug" though. As it was by design to fail when the queue is empty.  Would like to hear opinion shall we fix that or keep that as a feature.
1. No matter we fix it by enabling empty queue or not fix it, it would lead to a possibility of an unfinalized in-flight exit. Would this cause any problem? (I can't think of any but I might miss sth)

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -803,14 +803,20 @@ contract RootChain {
      */
     function processExits(address _token, uint192 _topExitId, uint256 _exitsToProcess)
         public
+        returns (uint256)
     {
         uint64 exitableTimestamp;
         uint192 exitId;
         bool inFlight;
+        uint256 processedExitNum = 0;
+
+        PriorityQueue queue = PriorityQueue(exitsQueues[_token]);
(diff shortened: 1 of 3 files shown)


paulperegud:
Lets leave it as it is. In tests we always know how many exits are there to process. In production - caller also has to know how many exits he wants to process and what the state of the queue is. Calling processExits on empty queue is always a programming error.

boolafish:
As discussed, the client should always have an understanding on the status of the queue (at least not calling it when they have nothing to exit). Adding this change would increase gas call as well.

Close this PR.

PR #75, opened 2019-03-08
docs: explicitly state that processExit fails when exit queue for token is empty

pgebal:


paulperegud:
@pgebal what is the status of this one?

PR #82, opened 2019-03-25
Omg 399 illegality of transaction with holes 2

pnowosie:


contracts/PlasmaCore.sol
```diff
@@ -72,13 +73,23 @@ library PlasmaCore {
                 txindex: input[1].toUint(),
                 oindex: input[2].toUint()
             });
+            
+            // check for empty inputs - disallow gaps
+            if (decodedTx.inputs[i].blknum == 0) emptySeen[0] = true;
```
paulperegud: We are not consistent on definition of what constitutes gap.

https://github.com/omisego/elixir-omg/blob/84c407396f6da0b57514cc4871c41efe7dfc8cf0/apps/omg/lib/state/transaction.ex#L230

contracts/PlasmaCore.sol
```diff
@@ -72,13 +73,23 @@ library PlasmaCore {
...
                 owner: output[0].toAddress(),
                 token: output[1].toAddress(),
                 amount: output[2].toUint()
             });
+
+            // check for empty outputs - disallow gaps
+            if (decodedTx.outputs[i].owner == 0 
+             && decodedTx.outputs[i].token == 0 
+             && decodedTx.outputs[i].amount == 0) emptySeen[1] = true;
+            else require(emptySeen[1] == false, "Gaps in outputs not allowed ");
```
paulperegud: "are not allowed"?

contracts/PlasmaCore.sol
pnowosie: To summarize internal discussion: input {0, 0, 0} is incorrect as any other from block 0, however it's allowed because it's how we represent empty (not provided) value.
And analogically we apply the same for outputs: all zeros means empty.

Issue #80, opened 2019-03-19
exit id for deposit is not unique 


pthomalla:
## Issue Type 

```
[0] bug report
[ ] feature request
```

## Current Behavior

Create two deposit with this same data (`owner`, `token`, `amount`) generate the same deposit bytes.
Then `getStandardExitId` isn't unique, preventing exits 

## Expected Behavior

Change processDeposit to create unique `deposit` bytes.

- Command that caused error:  `function getStandardExitId(bytes32 _txhash, uint8 _oindex)` and  `function _processDeposit(bytes _depositTx, PlasmaCore.Transaction memory decodedTx)`

pik694:
We thought about several solutions:
1. Instead of creating a deposit transaction with empty inputs, we could create a transaction
were the first input's `blknum` would be `nextDepositBlock` shifted left 128-bits. Such an approach would mean that in the protocol (or just a tinny part of it) we would be accepting spending inputs from the future. This may lead to problems in the future.
2. Instead of computing `exit_id` from `hash(txbytes)`, we could compute it from `hash(txbytes <> utxopos)`. Unfortunately this would not work, as we do not always have the `txoPos` (ex. in-flight transaction), therefore we would not be able to compute `exit_id` of IFE output.
3. The third solution, and the implemented one, is that we make a clear distinction between `exit_id` of deposit and any other transaction. If the transaction is an deposit the `exit_id` is computed from `hash(txbytes <> utxopos)`, which will always be unique, as blknums will be unique.  `Exit_id`s of other transactions remain being computed from `hash(tx_bytes)` only. The only problem is the fact that if we started an IFE from a deposit, we could not check whether someone tried to also SE the deposit, but this is not a problem, since we cannot start an IFE from the deposit - `sum of inputs < sum of outputs`.

boolafish:
Thanks for the explanation of possible approaches! It is really helpful!

For the first approach, I 'd like to clarify:
> we would be accepting spending inputs from the future

Why is it "future"? Though it is `nextDepositBlock`, but that is the block where that certain deposit tx would be at. If the deposit succeeds, it will be "present" not "future". If it fails, it doesn't matter.

BTW, what are possible problems in the future that you've seen in using the first approach though?

I am asking this because the first approach looks more consistent within all exit-id creation, so I'd like to clarify the reason to prefer 3 over 1.

boolafish:
[...]

PR #84, opened 2019-03-26
OMG-443 exit id for deposit is not unique

pik694:


pik694:
Refers #80

contracts/RootChain.sol
```diff
@@ -845,29 +847,46 @@ contract RootChain {
 
     /**
      * @dev Given UTXO's position, returns its exit ID.
```
paulperegud: This line is misleading - UTXO position is only one of two args.

paulperegud:
Hmm, one more remark. We should detect if tx is an deposit in consistent way. When determining exit priority, we check if `blknum rem 1000 != 0`.

pik694:
Closes #80

PR #87, opened 2019-03-29
chore: use more descriptive names for functions manipulating bit flags of IFEs

pgebal:
…s on in-flight exits

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -28,7 +28,6 @@ contract RootChain {
     using PlasmaCore for uint192;
     using PlasmaCore for uint256;
 
-
     /*
      * Storage
      */
@@ -85,7 +84,6 @@ contract RootChain {
         address token;
         uint256 amount;
     }
-
 
     /*
      * Events
@@ -167,15 +165,12 @@ contract RootChain {
         _;
     }
 
-
     /*
-     * Constructor. Empty because see `function init()`
+     * Empty, check `function init()`
      */
-
     constructor()
         public
     {
-
     }
 
     /*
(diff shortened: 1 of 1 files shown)


contracts/RootChain.sol
```diff
@@ -976,48 +965,105 @@ contract RootChain {
         return _value.bitSet(255) || _value.bitSet(254);
     }
 
-    function setFlag(uint256 _value)
-        public
+    function clearFlag(uint256 _value)
```
paulperegud: unused?

contracts/RootChain.sol
```diff
@@ -976,48 +965,105 @@ contract RootChain {
...
+    {
+        return _ife.exitStartTimestamp.bitSet(255) || _ife.exitStartTimestamp.bitSet(254);
+    }
+
+    function isFinalized(InFlightExit storage _ife)
+        view
+        private
+        returns (bool)
+    {
+        return _ife.exitMap.bitSet(255) || _ife.exitMap.bitSet(254);
```
paulperegud: This one should check only bit 255. Bit 254 is never set.

contracts/RootChain.sol
```diff
@@ -1333,7 +1374,7 @@ contract RootChain {
         _inFlightExit.bondOwner.transfer(inFlightExitBond);
 
         // Flag as finalized
-        _inFlightExit.exitMap = setFlag(_inFlightExit.exitMap);
+        finalize(_inFlightExit);
```
paulperegud: Inconsistent naming:
`finalize, isFinalized`
`piggyback, isPiggybacked`
vs
`setExited, isExited`

Issue #88, opened 2019-04-02
Inputs of deposit tx are not checked


pik694:
When processing a deposit we check if only the first output is set, but we do not check inputs at all.

Not checking inputs  (asserting all are zeroed) of the deposit may lead to some unexpected behaviour, ex. making an IFE canonical by creating the deposit tx that looks the same as an IFE transaction.

Things that can't be done using this potential vulnerability are:
1) preventing the exit by creating deposit that "spends" exited utxo <-> we are checking the signature, not the inclusion proof
2) ...

PR #89, opened 2019-04-03
fix: inputs of deposit not checked

boolafish:
Current contract does not check the input of deposit tx
should be empty. As a result, currently user can fake any input
data to the deposit tx. This may lead to some unexpected behaviour,
eg. making an IFE exit_id not unique by mimicing another existing
transaction with the same input. We did not separate deposit exit id
from regular tx for IFE because it is assumed that deposit tx cannot IFE
due to sum(inputs) would be 0 < sum(outputs). However, if fake input
is possible, then the assumption would be wrong. 

This commit add check to deposit tx inputs on `blknum` to be 0,
which should be enough as block start with 1.

issue: https://github.com/omisego/plasma-contracts/issues/88

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -292,9 +292,14 @@ contract RootChain {
         // deposit with blknum ending with 000.
         require(nextDepositBlock < CHILD_BLOCK_INTERVAL);
 
-        // Check that all but first outputs are 0.
-        for (uint i = 1; i < MAX_INPUTS; i++) {
-            require(decodedTx.outputs[i].amount == 0);
+        for (uint i = 0; i < MAX_INPUTS; i++) {
+            // all inputs should be empty
+            require(decodedTx.inputs[i].blknum == 0);
+
+            // only first output should have value
+            if (i >= 1) {
+                require(decodedTx.outputs[i].amount == 0);
+            }
         }
 
         // Calculate the block root.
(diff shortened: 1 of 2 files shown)


boolafish:
Discussion: 
1. is checking `blknum` enough? Shall we actually check all fields?
2. as asked in slack, why double checking `require(nextDepositBlock < CHILD_BLOCK_INTERVAL);`. If not necessary I can sneak in the change to remove duplicated ones.

contracts/RootChain.sol
```diff
@@ -292,9 +292,14 @@ contract RootChain {
         // deposit with blknum ending with 000.
         require(nextDepositBlock < CHILD_BLOCK_INTERVAL);
 
-        // Check that all but first outputs are 0.
-        for (uint i = 1; i < MAX_INPUTS; i++) {
-            require(decodedTx.outputs[i].amount == 0);
+        for (uint i = 0; i < MAX_INPUTS; i++) {
+            // all inputs should be empty
+            require(decodedTx.inputs[i].blknum == 0);
```
pik694: checking only `blknum` is safe as we start counting blocks from 1, so even if other fields would not equal to 0, the deposit transaction would not be spending any valid `utxo`

pik694: please note that the deposit transaction itself is not signed, so it cannot be used to challenge any other transaction (shared inputs)

boolafish: ouch, didn't look clear enough from the issue lol and misunderstood the intention. I will change my PR description and commit message before merge

pik694: Sorry, the issue was rather a quick note. I meant to document it better, but you took it earlier ;)

bg1156:
Great going folks, keep it coming...I really do appreciate your hard work.
byron111

PR #93, opened 2019-04-17
Abstract contract design draft

boolafish:
**Disclaimer: This is for discussion use only. Not aiming for merge. Also, any code is not tested and is basically pseudo code for review/discuss purpose.**

### Basic Gaol
- Able to support ODEX tx
- Able to upgrade our ODEX in the future
- Try to be flexible for adding new features on top of plasma

### High Level Design Concept
In UTXO blockchain, Tx is a record of state transition. Now we bind each "tx type" to an exit game logic. In our use case of expending to ODEX, it would be: basic morevp tx, funding tx, settlement tx, and withdraw tx. 

Tx can be some data like: 
```
{
    txType: [:type]
    inputs: [{UTXO_position}, {UTXO_position}...],
    outputs: [{token, amount, outputType}, {token, amount, outputType}, ...],
    data: [] // we can put proof and other data here. eg. signature, settlement orders....
}
```

And then we would have a mapping of `txType` <--> `exit game`. 

Also, there's an extra `outputType`, this is to limit what type of output can be used as tx input. For instance, in "settlement tx", we might not want the input to use output from "morevp tx". Instead, we want to make suer we are only using "funding tx" output.

BTW, I did not put `owner` in output because in the case of ODEX funding tx output, the ownership is like between exchange and user. Probably it would be better to structure that info in the `data` field?

With this concept, now we have the flexibility of adding new type of tx to the chain! (But it has to be really carefully designed and reviewed. Any invalid tx type can break the whole chain as we are not separating them)

### Changes to Plasma Layer
1. To support this, we first need a place to register exit games. see `ExitGameRegistrator.sol` for this.
1. Plasma layer handles the deposit, submit block, priority queue, and token transfer (eth or erc20).
1. Plasma layer proxies the exit game logic to the exit game contract. (see `runExitGame ` in `ExitGameController`)
1. Exit game contract decides whether the exit can be finalized and also how the exit would end up sending funds to user. (eg. SE sends funds of tx output, IFE sends funds of all piggybacked input/output, ODEX sends fund to user instead of exchange....etc)
1. To make sure we can easily upgrade the "exit game contract" and the plasma contract, all the storage is put in a centralized place `PlasmaStorage.sol`
1. Exit game contract would have access to the "GeneralizedStorage" variables. They can use that to save the required info.

#### Success flow
1. Alice start exit
1. Plasma Layer proxy that to the Exit Game contract
1. Exit Game contract checks the exit, and safe the essential exit data (eg, amount, token, exit_to) in Plasma Layer's `GeneralizedStorage`. 
1. Exit Game enqueue the exit to priority queue
1. Sb. calls finalize exit
1. Plasma Layer loops through the priority queue, get the exit out, ask the related Exit Game contract to process the exit.
1. Exit Game contract would know how to exit. eg. SE exit from output, IFE exit from piggybacked, ODEX exit to user and not exchange....etc
1. Exit calls PlasmaLayer's function to perform the real transfer (ETH, ERC20...).

#### Failure flow
1. Alice start exit
1. Plasma Layer proxy that to the Exit Game contract
1. Exit Game contract checks the exit, and safe the essential exit data (eg, amount, token, exit_to) in Plasma Layer's `GeneralizedStorage`. 
1. Exit Game enqueue the exit to priority queue
1. sb. challenges that exit
1. Plasma Layer would proxy the challenge to Exit Game contract.
1. Exit Game Contract would safe some variable (in Plasma layer's GeneralizedStorage) to flag that exit failed
1. When finalized exit, it would ask the Exit Game contract to see it can exit or not. As it was flagged failed previously, it would return false.
1. Exit won't be processed.

### FAQ
1. Why not "predicate"?
The hardest thing that blocks us from using predicate is in DEX, we have the situation where we cannot verify state transit in one single function (gas limit). Thus, we need customized interactive game to do that. 
Though we can work on to have some general interface to support that, but give the complexity on the possible DEX exit game and unknown of what we are going to support in the future, it makes more sense to just proxy them.

1. What's some downside of this design?
It introduced some back and forth communication between plasma contract and exit game contract. And all the states is stored in plasma contract, so for exit game to use those storage, it needs an extra call. Each external call would add a basic of 700 gas.

1. What are some known limits that this does not support?
  - If we want to withdraw to another contract by calling the specific smart contract function, exit game contract does not handle that. We would need a upgrade to `PlasmaWallet.sol` to support that.
  - If we want to support new type to ERC token (eg. ERC721), we need to change `PlasmaWallet.sol` as well. 
  - Any new tx type game should be thinking in MVP/MoreVP way.


pdobacz:
Hi, thanks for this. I'm not delving in too deeply, but I have some quick thoughts:

1/ in the other discussion revolving around how to customize transactions via predicates, the "output-type" was required to be more flexible (CC'ing @rodiazet to take a quick look too and/or correct me), as in "ERC721-outputs" vs "ERC20-outputs". "ERC721-outputs" for example would not have `amount`. So maybe sth more along the lines of:

```
{
    txType: [:type]
    inputs: [{UTXO_position}, {UTXO_position}...],
    outputs: [{outputType, outputData}, {outputType, outputData}, ...],
    data: []
```

2/ tagging txs with `funding tx` type would expose the intent to trade to the public, which is what in subsequent iterations we'd want to avoid
3/ it may be worth to separate the `metadata` field as well - currently it has the traits of being "invisible" from POV of all plasma logic, size-limited to 32 bytes, and optional. As such I think it's good to keep it separate from the `data` field (or any other)

boolafish:
[Note]
- For ODEX, we might need to be able to support pure message transaction (for withdraw). Seems like withdraw might need a notification message first and then be able to withdraw (to protect venue). Apparently we need no games for such tx type. So probably it breaks nothing.

- Thanks @paulperegud , some reasonings about "output type": https://github.com/omisego/plasma-contracts/issues/94

boolafish:
@pdobacz 
1/ yeah, I agree that is better and we would need that!
2/ I agree that is not ideal, just meanwhile I couldn't find a way to totally get rid of that. (@paulperegud has an idea of having offline message instead of funding tx. But that would means we would need to add the "offline message tx" into the normal tx exit game. Might work? Not sure if that's better.)
3/ as for `data` field I am thinking about we can structure something like this for normal tx
```
data: {
   metadata: "xxxxxxx",
   signatures: [],
}
```

pdobacz:
> 3/ as for data field I am thinking about we can structure something like this

I'd support my voice to have `metadata` an entirely separate field on the top level, since it has (and should never ever have) any use in "our" code, it's all up to the users to put stuff there, we only include it in the txhash, then forget about it.

boolafish:
yeah, I think that's a good point. I was not so much fan of having them in same level because `data` and `metadata` looks confusing to me, like at first glance, what's the different? I think the following might be better. 

```
{
    type: xxx,
    inputs:[],
    outputs:[],
    proofData: {},
    metadata: xxx,
}
```

boolafish:
#### Update
TL;DR 
now support a predicate registry taking `(TxOutputType, ConsumingTxType)` as key. The predicate would define how a tx output type can be used in certain next consuming tx type.

From the previous discussion with @paulperegud, we sort of find out probably watcher only cares about "how to define a tx output is used?" + "sum of input >= sum of output". They don't really need to care about whether "all txs are format-wise valid". (eg, they might not need to care the trade tx has right orders attached) 

So I update the interface here to support "TxOutputPredicate". (See `TxOutputPredicateInterface.sol` and `NormalOutputToNormalTxPredicate.sol`)

It's with key of two things: `(TxOutputType, ConsumingTxType)`. This means in the future, whenever we want to add new tx type that can consume certain existing tx output type, we don't even need to upgrade old predicate (If we only use `TxOutputType` as key). We can simply deploy new predicate.

#### Regarding funding tx
> 2/ tagging txs with funding tx type would expose the intent to trade to the public, which is what in subsequent iterations we'd want to avoid


After this, I think instead of "funding tx" that takes normal txoutput as input. We can have "funding msg tx" that has no input but points to a UTXO of normal tx. And the trade tx needs to prove the existence of "funding msg tx". So we would have a "normal tx output <> trade tx" predicate that defines the check of such "funding msg tx".

The funding msg tx can be encrypted so no one would know the detail in advance.

One more issue in here is the user can send another normal tx that uses the tx output (utxo) after the "funding msg tx". This would hurt venue as venue needs to re-calculate the proof. So the "funding msg tx" probably should be with a bond. User can transfer away the money but they would lose a bond if they do so.

boolafish:
Added some graph: [click here](https://omisego.atlassian.net/wiki/spaces/RES/pages/35520517/Abstract+Layer+Architecture?atlOrigin=eyJpIjoiNTYyMmI0Y2YyYWNjNGI3YzhlMmU5OTQ2NWYyMGQwNTYiLCJwIjoiYyJ9)

boolafish:
[Update 2019.05.14]

Found a security design bug. The function of "processExit" cannot be in the same contract as ExitGame defines. Since we allow client to proxy functions to ExitGame contract directly, this would allow client to call "processExit" directly.

Separate the function to another contract "ExitProcessor" and is logically coupled with ExitGame (knows how the game work and the condition to exit). Same, add a registry for that.

Have updated the doc and the pseudo code.

kfichter:
Heyo, was asked to provide some review here. Generally, I would construct this in the following way (typescript interfaces for simplicity):

```typescript
interface State {
  data: string // Arbitrary data
  transactionType: string // Reference to the function that checks how the state can be mutated
}

interface StateObject {
  id: string // Unique identifier
  state: State // State data
}

interface Transaction {
   inputs: string[] // List of `StateObject` IDs.
   outputs: StateObject[] // List of outputs
   witnesses: string[] // List of witnesses, one for each input
}
```

So for example, let's imagine we have only have one simple transaction type that allows the `owner` of an output to spend. We'll say the corresponding `transactionType` for this transaction is `0xabc`. A `StateObject` locked with that transaction type might look like this:

```typescript
const obj: StateObject = {
  id: '0x12345', // Arbitrary ID
  state: {
    data: '0x5a0b54d5dc17e0aadc383d2db43b0a0d3e029c4c', // Arbitrary address
    transactionType: '0xabc'
  }
}
```

Now let's construct a transaction for this:

```typescript
const tx: Transaction = {
  inputs: ['0x12345'], // ID of the thing we're spending
  outputs: [
    {
      data: '0xf25746ac8621a7998e0992b9d88e260c117c145f', // Arbitrary address
      transactionType: '0xabc'
    }
  ],
  witnesses:['0x.....'] // Witness, single over the inputs and outputs in this case
}
```

When you're checking the validity of a transaction you'd effectively follow the following process:

2. For each input, pass the full transaction + witness + input index to the "transaction validation function" that corresponds to the input's `transactionType`.

The "transaction validation function" for this specific transaction type might look something like this (very simplified, not at all how this will look in practice but just to give you an idea):

```typescript
function validate(transaction: Transaction, inputIndex: number): boolean {
  const inputObject = getStateObject(transaction.inputs[inputIndex])
  const owner = decodeStateObject(inputObject).data.owner
  const signature = transaction.witnesses[inputIndex]
  
  return ecrecover(transaction.unsigned(), signature) === owner
}
```

Again, you'd want to do this for each of the inputs. If doing so for any input returns "false", then you'd reject the transaction. 

Thoughts? Happy to provide more information if anything is confusing.

boolafish:
Hi @kfichter, thanks for the review! :)

(btw, I just moved the [design doc](https://docs.google.com/document/d/1PSxLnMskjqje4MksmW2msSSg-GtZoBSMNYey9nEDvt8/edit?usp=sharing) public accessible in case there was no access before.)

I could explain a bit on why the design is a bit different from yours:

1. 
> interface State {
  data: string // Arbitrary data
  transactionType: string // Reference to the function that checks how the state can be mutated
}

I've chosen `outputType` instead of `transactionType` on output because we would have the use case of an output can possibly be used in two different kinds of txs (or there can be multiple ways to transit the state). eg, an normal payment transaction output can be used in transaction to provide funding to DEX and, of course, in normal payment tx.

As a result, the `validate` function ("transaction validation function") of yours would be in our predicate which is unique per `(outputType, consumeTxType)` instead of per `txType`. 

2. 
> When you're checking the validity of a transaction you'd effectively follow the following process:

In general, there are two places that would need the check of validity of a transaction: on ethereum games and in backend system (operator/watcher).

We give up the idea of full validation of transaction on ethereum at once because it will hit the gas limit easily, especially we are trying to do batch settlement transaction. As a tradeoff, each transaction has to define their own exit/challenge game functions. The challenge can be more specific to an invalid condition to reduce gas (eg. in DEX tx, order mismatching input/output can be a separate condition and have a specific challenge game for that). We would need to make sure transaction is "valid" or I think you use the word "exitable" in in-flight exit of MoreVp doc. 

(BTW, as a result we recently investigated on [the idea of using MVP instead of MoreVp for DEX](https://github.com/omisego/research/issues/83) to remove such restriction from the beginning so we might be fine now, but I started the abstraction design with the idea to support complex IFE txs.)

However, I do agree those validation function can be used in backend system. And the way backend system does would be quite similar to what as you described.

Meanwhile, it might not be enough to verify just all "transaction validation function" for state/output is valid. To make a transaction valid, there are some condition that is in "transaction" level instead of "state/input/output" level. For example, sum of inputs >= sum of outputs. This is something that needs all inputs and outputs data to be able to be checked. Depends on the business logic of transaction, there might be even more conditions that is on a higher level.

3. 
> interface StateObject {
  id: string // Unique identifier
  state: State // State data
}

I actually like the idea of having `id` per state/output. I will investigate this a bit more to see if there's any difficulties or obvious tradeoff to use `id` instead  of `utxo_pos`.

boolafish:
as this is not aiming for merge, and we got other branch for POC, closing this PR.

PR #99, opened 2019-04-30
Expose eip 712 hash public for tests

pnowosie:
This is needed for https://github.com/omisego/elixir-omg/pull/656

### DIFF
--- a/contracts/SignatureTest.sol
+++ b/contracts/SignatureTest.sol
@@ -1,0 +1,23 @@
+pragma solidity ^0.4.0;
+
+import "./ECRecovery.sol";
+import "./Eip712StructHash.sol";
+
+
+/**
+ * @title SignatureTest
+ * @dev Opens Eip712 structural hash for end-to-end tests
+ */
+contract SignatureTest {
+    /**
+     * @dev Verifies signature was made by the signer.
+     * @return True when signature is valid, false otherwise.
+     */
+    function verify(bytes _tx, bytes _sign, address _signer)
+        public
+        view
+        returns (bool)
+    {
+        return _signer == ECRecovery.recover(Eip712StructHash.hash(_tx), _sign);
+    }
+}
--- a/contracts/SignatureTest.sol
+++ b/contracts/SignatureTest.sol
@@ -6,7 +6,7 @@ import "./Eip712StructHash.sol";
 
 /**
  * @title SignatureTest
- * @dev Opens Eip712 structural hash for end-to-end tests
+ * @dev Opens Eip712 structural hash for dependency conformance tests
  */
 contract SignatureTest {
     /**
--- a/contracts/SignatureTest.sol
+++ b/contracts/SignatureTest.sol
@@ -10,14 +10,14 @@ import "./Eip712StructHash.sol";
  */
 contract SignatureTest {
     /**
-     * @dev Verifies signature was made by the signer.
-     * @return True when signature is valid, false otherwise.
+     * @dev Recovers signer address from signature and EIP-712 structural hash of transaction.
+     * @return Signer address when signature is valid, random bytes otherwise.
      */
-    function verify(bytes _tx, bytes _sign, address _signer)
+    function getSigner(bytes _tx, bytes _sign)
         public
         view
-        returns (bool)
+        returns (address)
     {
-        return _signer == ECRecovery.recover(Eip712StructHash.hash(_tx), _sign);
+        return ECRecovery.recover(Eip712StructHash.hash(_tx), _sign);
     }
 }


pthomalla:
I think it's a bad idea, to discover an internal function for test

pnowosie:
> I think it's a bad idea, to discover an internal function for test

Idea here is to keep functions internal in production code. I expose then in a separate test contract to test dependency conformance.

boolafish:
It looks weird to me as well though....why we would need to expose function for "end-to-end" test? 
By "end-to-end" usually it should be pretty blackbox testing, isn't it?

pnowosie:
> By "end-to-end" usually it should be pretty blackbox testing, isn't it?

Strictly said it's `dependency conformance` test we have eip-712 hashing dependency in elixir and in solidity, we need to make sure they both do the same. Do I address you concern @boolafish :question:

boolafish:
Okay, got your point.

However, I feel this is a bit overkill and would be better to just write unit tests in both implementations (BTW, it seems like in elixir, those functions are just test fixture, do we really care?). I am saying unit test because,  If I have the right understanding, the function here is pretty much pure computation and we should ideally have enough confident it is correct by just unit test. 

I guess what we want is both to be "correct" and naturally be the same. Also, take a bit extreme example, if we are writing a `add(x, y)` function, would you test dependency this way?

pdobacz:
my cents:

Generally we definitely avoid exposing any "test-only" interfaces, guts and all that, I'm trying to be super careful that this never happens and treat "testing via public API" even dogmatically I'd say.

That being said I think this context here is special:

Errors that arise from things like incorrect signatures/merkle proofs/hashing/encoding in Solidity contracts are notoriously hard to debug, at least with the current tooling we use (I'm not aware of anything that would revolutionize here, yet). This prompted me to suggest we build such suite of compliance test, tests of signature being the first.

We do have end-to-end, public-API based tests that test whether the contract behaves as expected when prodded with inputs coming from our Elixir code - those are the `:integration` tests. And this is the black-box testing you're speaking to here, and this is the, let's say "main test" we're looking at, when figuring out whether we can talk to the contract.

Those `:integration` tests however take ages to execute, so we keep only a minimum amount of those, testing selected "happy paths" only.

Suppose we didn't have those compliance tests like this one, and one of those high level `:integration` tests failed - it would take considerable effort to rule out problems with the signing mechanism.

Other example (and my current plan for these compliance tests to evolve) is the merkle hashing/proving code - I'd love to be able to always reliably tell that this is executed identically on both Solidity and Elixir end, but doing so extensively via our public APIs would be tedious.

> just write unit tests in both implementations

I think such unit tests would never tell you whether both implementations work the same - they could both do the job as specified but still produce different signatures. Simplest example would be when one of the EIP-712 implementations would miss one round of `keccak256`-ing or `abiEncodePacked`-ing _somewhere_.

On the other hand, in a way, this is a unit test. It's just labeled `:integration` because it uses an eth-node to do the EVM work, but like you said it is pure computation. We're asserting that two disparate implementations of an algo give bit by bit identical results, under multiple varying inputs.

> Also, take a bit extreme example, if we were writing a add(x, y) function, would you test dependency this way

I don't agree with the analogy implied - `add` is way more natural, debuggable than a stack of `keccaks`,`encodings` etc where all data looks like `/dev/random` and your test feedback is basically just `transaction ok | transaction failed`

boolafish:
> Errors that arise from things like incorrect signatures/merkle proofs/hashing/encoding in Solidity contracts are notoriously hard to debug

I agree, that is also why I am pushing so hard to have unit test in contract. Currently we don't really have unit test hitting the `Eip712StructHash.sol`. To do so, it means we would need some mock/test contract that exposes the function, but I think the trad-off is good and also this practice is heavily used by openzepplin.

> Simplest example would be when one of the EIP-712 implementations would miss one round of keccak256-ing or abiEncodePacked-ing somewhere.

Doesn't this means your test is basically written wrong? If what your test is expecting/asserting is something we are confident to be the "true" value, and both implementation has the test suite that makes us confident they are both correct, how would this happen?

boolafish:
I think I do see your pain point for hard to pinpoint the failing part. My 2 cent would be let's try to improve our unit tests on contracts first. Ideally this should even block integration test from running. 

However, if this still obviously does not help the situation, or it's just really too hard to write good quality solidity test codes then we go that path of "testing both implements are same".

contracts/SignatureTest.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.4.0;
+
+import "./ECRecovery.sol";
+import "./Eip712StructHash.sol";
+
+
+/**
+ * @title SignatureTest
+ * @dev Opens Eip712 structural hash for end-to-end tests
```
boolafish: Minor: probably we'd like to re-phrase end-to-end here as well

contracts/SignatureTest.sol
```diff
@@ -0,0 +1,23 @@
...
+ */
+contract SignatureTest {
```
boolafish: Personally I would prefer we expose more directly on the `Eip712StructHash.hash`.
If we definitely want `ECRecovery.recover`, we can expose that as well.

Reason is that if we start to do test in the same style of [open-zeppelin](https://github.com/OpenZeppelin/openzeppelin-solidity/tree/master/contracts/mocks), those can be used and naturally exposed. Also, this make this agnostic of the omg-elixir repo.

pnowosie: But then I'd need to make 2 calls instead of 1 to this test-only contract. Also exposing these functions make impact on production code (RootChain contract dependencies) I'm not really feel competent. Postponing for this :pr:

boolafish: yeah, I come from a point if we are to exposing them anyway and more from a point of this repo instead of omg-elixir. But not important now and with current situation might not exactly being better as a whole.

PR #100, opened 2019-05-09
Add current contract address to struct hashing 'eip_domain'

pnowosie:
With the assumption that on different Ethereum networks different deployer addresses will be used, this :pr: adds replay protection for plasma transactions.


pdobacz:
added a31056c37ce1a78c63e0e756fe65a3322c841481 with the conformance of tests to varying verifying contract address when signing transactions.

I'll let you take over now. If the commit is ok consider squashing into the main commit

pnowosie:
Your changes @pdobacz are fully related to the test system, so I like separation of commits (won't squash).
Thanks a lot :exclamation:

PR #101, opened 2019-05-13
Plasma framework implementation

pgebal:
Not for merging.


plasma_framework/ExitGameController.sol
```diff
@@ -0,0 +1,55 @@
...
+import "./TxModel.sol";
+
+contract ExitGameController is ExitGameRegistry {
+    struct Exit {
+        ExitGame exitGameContract;
+        TxModel.Tx tx;
+        bytes exitData;
+    }
+
+    // NOTE: exitId must be unique among all exit games
```
boolafish: I am curious why is this a case? Does the nonce + priority thing does not work?

pgebal: Piotr left me a note that if is necessary for IFE being aware of SE on the same output.

boolafish: hmm...a SE and IFE would exists the same time in same exit game contract if it is MoreVp protocol. For instance, if we are having a MoreVP ExitGame contract, the contract would have both. So `exitId` needs to be unique per ExitGame contract. 

But if we are having a DEX ExitGame, it should be totally fine to use the same exit id as in MoreVp ExitGame contract since ideally the two contracts should not know each other.

Probably my doc was a bit mis-directing in some description that leads to this confusion.

boolafish: I re-read the SE and IFE interaction doc again: https://github.com/omisego/elixir-omg/blob/master/docs/standard_vs_in_flight_exits_interaction.md

The concern is IFE exit from input might collide with SE from previous transaction's output.
So actually the current solution using `standardExitId(txbytes, utxo_pos)` for 
`2. Mallory does SE on input, finalized -> Mallory submits IFE` 
would not work in this framework. However, instead of blocking the IFE, we can just add `nonCanicalChallenge` with the condition of input finalized (to be seen as used).

Thanks for pointing this out!

paulperegud: > However, instead of blocking the IFE, we can just add `nonCanicalChallenge` with the condition of input finalized (to be seen as used).

Yep. Whole conflict can be resolved by adding additional challenges. We chose to change the way standardExitId is computed because it allowed to contain required changes to contract.

boolafish: I re-think about the SE vs IFE interaction. If we have a way to make sure one output can only exit once, I think SE and IFE do not need to interact together at all. We can just let SE and IFE both able to finalize the exit but only the first exit would be processed with fund transfer, the second exit would simply check such output finalized and do nothing.

However, one issue is that we cannot simply/naively use `utxo_pos` as identifier to block output double-exit since an in-flight tx does not has block number yet. One solution is to put a nonce as `id` on each output.

paulperegud: > We can just let SE and IFE both able to finalize the exit but only the first exit would be processed with fund transfer, the second exit would simply check such output finalized and do nothing.

I like this one. Looks like a way to simplify the contract.

> One solution is to put a nonce as id on each output.

If I can chose the same nonce as you and next exit first, am I preventing you from performing an exit?

We've used hash of tx concatenated with output oindex. Check out getStandardExitId implementation.

boolafish: > EDIT: what about the case where IFE input was already exited via SE?

haha, yeah you are right, this is not covered. 
btw, Blocking from the start might have an attack that one can first start SE of input, holding the in-flight tx until very close to end of the finalization of SE. And then start IFE. Now, IFE would not be blocked as the SE is not finalized but there is really limited time for others to challenge the SE with the newly disclosed in-flight tx.

> If I can chose the same nonce as you and next exit first, am I preventing you from performing an exit?

ouch, that's so right. Especially in MoreVp we cannot be protected by priority since a later generated nonce can have higher priority input.

> We've used hash of tx concatenated with output oindex. 

This one looks really good!

boolafish:
closing this.

PR #122, opened 2019-06-11
115 deploy with truffle

kevsul:
Added Truffle support for deploying contracts

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Plasma Contracts
 
-Root chain contracts for Plasma MVP, work in progress.
+Root chain contracts for Plasma M(ore)VP, work in progress.
 
 ## Contents
 This version of the contract implements [Plasma MVP](https://ethresear.ch/t/minimal-viable-plasma/426) (Buterin, Poon, Knott). This implementation is a PoA scheme with one operator and multiple watchers (users). Detailed description of our child chain design is in [Tesuji document](https://github.com/omisego/elixir-omg/blob/master/docs/tesuji_blockchain_design.md).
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -50,3 +50,54 @@ Running slow (overnight) tests:
...
+```
+npm install
+```
+
+
+### Deploying
+Deploying the contracts requires three accounts:
+1. `DEPLOYER` The account that actually deploys the contracts
+2. `AUTHORITY` The Authority account calls `RootChain.init()` and is the account used by the Child chain (or operator). By default a new `AUTHORITY` account is created when deploying. If you prefer you can use an existing `AUTHORITY` account, but it must not have made any transaction prior to calling `RootChain.init()` i.e. its nonce must be 0.
+3. `FAUCET` If a new `AUTHORITY` account is created, the `FAUCET` account is used to fund it. This account can be the same as the `DEPLOYER` account.
```
paulperegud: I can't imagine a case where we would need faucet to be different from deployer. Why not simplify?

truffle.js
```diff
@@ -0,0 +1,53 @@
...
+      settings: {
+        optimizer: {
+          enabled: true,
+          runs: 1
+        }
+      }
+    }
+  },
+  networks: {
+    development: {
```
boolafish: This would override the behavior for running `truffle test`. Without `development` it would automatically spin up a blockchain for you and with this we need to spin up the blockchain ourselves that matches the configuration here.

Do we really need to config this now?

package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/omisego/plasma-contracts.git"
+  },
+  "author": "",
+  "license": "ISC",
```
boolafish: perhaps `Apache-2.0`?

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "plasma-contracts",
+  "version": "1.0.0",
+  "description": "Root chain contracts for Plasma MVP, work in progress.",
```
boolafish: How about: MVP --> M(ore)VP

README.md
```diff
@@ -50,3 +50,54 @@ Running slow (overnight) tests:
...
+
+
+Normally you will deploy the contracts using an Ethereum client that you run yourself, such as Geth or Parity. However, you can also use a provider such as Infura. In this case you'll need to know the private keys for the `DEPLOYER` and `AUTHORITY` accounts. See `truffle.js` for an example.
```
boolafish: It seems like truffle changed the default config file name from `truffle.js` to `truffle-config.js`. Does your truffle auto generate `truffle.js`? Mine provides me with truffle-config.js though

- https://ethereum.stackexchange.com/a/62728
- truffle doc: https://www.trufflesuite.com/docs/truffle/reference/configuration

truffle.js
kevsul: Ah ok. I think it's useful for deploying to a local eth network.
I'll change the name of it to `local` to avoid overriding `truffle test` behaviour

README.md
kevsul: Yeah, `truffle-config.js` is a better name. I'll rename it.

package.json
kevsul: I didn't explicitly add this - npm must pick it up automatically from the README!
I'll change it in both places 😄

README.md
```diff
@@ -50,3 +50,51 @@ Running slow (overnight) tests:
...
+ - `DEPLOYER_PASSPHRASE` Passphrase of the `DEPLOYER` account.
+ - `AUTHORITY_PASSPHRASE` Passphrase of the `AUTHORITY` account.
+ - `AUTHORITY_ADDRESS_INITIAL_AMOUNT` The amount the fund the `AUTHORITY` account with (in wei). Defaults to 1 ETH.
+ - `USE_EXISTING_AUTHORITY_ADDRESS` Set to `true` if you want to use an existing `AUTHORITY` account instead of creating a new one. You must also set `AUTHORITY_ADDRESS`
+
+
+
+Run truffle, passing in the network e.g.
+```
+npx truffle migrate --network development
```
boolafish: development --> local

Issue #121, opened 2019-06-11
[abstract] Implement BlockController


boolafish:
parent story: #105 
POC code: https://github.com/omisego/plasma-contracts/blob/abstract_poc/abstract_poc/contracts/src/framework/PlasmaBlockController.sol

PR #127, opened 2019-06-13
feat: BlockController for plasma framework

boolafish:
### Note
This commit implements and tests the BlockController for
plasma framework. It would be used to submit block. One
function for child chain block and another for deposit block.

- close issue: https://github.com/omisego/plasma-contracts/issues/121

### Test
```
Contract: VaultRegistry
    onlyFromVault
      ✓ accepts call when called by registered vault contract (46ms)
      ✓ reverts when not called by registered vault contract

 Contract: BlockController
    constructor
      ✓ nextChildBlock is set to "childBlockInterval" (101ms)
      ✓ nextDepositBlock set to 1 (87ms)
      ✓ childBlockInterval is set as the inserted value (73ms)
    submitBlock
      ✓ saves the child chain block root to contract (136ms)
      ✓ updates "nextChildBlock" with a jump of "childBlockInterval" (156ms)
      ✓ resets "nextDepositBlock" back to 1 (174ms)
      ✓ emits "BlockSubmitted" event (108ms)
      ✓ reverts when not called by operator (113ms)
    submitDepositBlock
      ✓ saves the deposit block root to contract (139ms)
      ✓ adds 1 to nextDepositBlock (140ms)
      ✓ does not change nextChildBlock (149ms)
      ✓ reverts when exceed max deposit amount (childBlockInterval) between two child chain blocks (272ms)
      ✓ reverts when not called by registered vault (104ms)
```


plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
```diff
@@ -1,11 +1,16 @@
 pragma solidity ^0.5.0;
 
-import "../modifiers/Operated.sol";
+import "../utils/Operated.sol";
```
boolafish: I gave up the having modifiers folder because other modifier would quite likely to live in registries (`onlyFromVault` and `onlyFromExitGame`). Unless the modifier does not need to access to some variable. 

(previously tried to use "getter" interface to solve that but turn out would be around 2k more expensive on gas if access variable via `this.var()` instead of `var` directly)

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -0,0 +1,58 @@
+pragma solidity ^0.5.0;
+
+import "./models/BlockModel.sol";
+import "./registries/VaultRegistry.sol";
+import "./utils/Operated.sol";
+
+contract BlockController is Operated, VaultRegistry {
```
boolafish: I notice that @pik694 implemented `transactionIncluded` in the [POC code](https://github.com/omisego/plasma-contracts/blob/abstract_poc/abstract_poc/contracts/src/framework/PlasmaBlockController.sol#L59), I was not aware that checking inclusion proof in exit game need some extra data from plasma framework layer.

However, I think it might be better if we separate the stateful part (get block root with block number) and move the stateless part of that into a lib instead of having an expose api doing the job. Feel like we potentially need that block hash anyway (eg. checking confirm sig). As a result, I did not add that function to the implementation here.

plasma_framework/contracts/mocks/framework/DummyVault.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    function setVaultRegistry(address _contract) public {
+        vaultRegistry = VaultRegistryMock(_contract);
+    }
+
+    function checkOnlyFromVault() public {
+        vaultRegistry.checkOnlyFromVault();
+    }
+
+    function setBlockController(address _contract) public {
```
pgebal: We don't we provide block controller and  vault registry in constructor?

plasma_framework/test/src/framework/registries/VaultRegistry.test.js
```diff
@@ -1,67 +1,88 @@
...
+const DummyVault = artifacts.require('DummyVault');
 
 const { BN, expectRevert } = require('openzeppelin-test-helpers');
 const { expect } = require('chai');
 
 contract('VaultRegistry', ([operator, other]) => {
     beforeEach(async () => {
         this.registry = await VaultRegistry.new();
-        this.dummyVaultAddress = (await VaultRegistry.new()).address;
+        this.dummyVault = (await DummyVault.new());
```
pgebal: We don't need brackets here.

plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
```diff
@@ -1,11 +1,16 @@
...
 
 contract VaultRegistry is Operated {
     mapping(uint256 => address) private _vaults;
     mapping(address => uint256) private _vaultToId;
 
+    modifier onlyFromVault() {
+        require(_vaultToId[msg.sender] != 0, "Not being called by registered vaults");
```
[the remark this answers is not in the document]
boolafish: well....it's rely on `uint` to make it positive. As a result, if a client really tries to register with `-1`, it would still work but just get a super huge number: `115792089237316195423570985008687907853269984665640564039457584007913129639935`.

However, would change the requirement here to be `_vaultToId[msg.sender] > 0`

plasma_framework/contracts/mocks/framework/DummyVault.sol
boolafish: well....mainly because different test needs different instance being set in advance here. And use setter just make it easier whenever we find out we need to put more object here. In production code I would try to avoid setter, but this is more like a test helper code so I will go with easier way.

plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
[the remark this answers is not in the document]
pgebal: I was thinking to put that requirement (> 0), in register... function.

plasma_framework/contracts/mocks/framework/DummyVault.sol
pik694: I agree with @pgebal. Could you please make a small comment in the code, so when it comes to implementing a production version the developer would not just copy the PoC code without rethinking the API.

plasma_framework/contracts/mocks/framework/registries/VaultRegistryMock.sol
```diff
@@ -0,0 +1,15 @@
...
+
+contract VaultRegistryMock is VaultRegistry {
+    bool public vaultCheckPass;
+
+    constructor() public {
+        vaultCheckPass = false;
+    }
+
+    function checkOnlyFromVault() public onlyFromVault {
+        vaultCheckPass = true;
```
pik694: What about emitting an event here? The checks could be made repeatedly.

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -0,0 +1,76 @@
...
+    uint256 private _childBlockInterval;
+    uint256 private _nextChildBlock;
+    uint256 private _nextDepositBlock;
+
+    mapping (uint256 => BlockModel.Block) private _blocks;
```
pik694: why do you want to keep this private?

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -0,0 +1,76 @@
...
+
+    event BlockSubmitted(
+        uint256 blockNumber
```
pik694: I think you could add `root hash` field to this event.

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+    constructor(uint256 _interval) public {
+        _childBlockInterval = _interval;
+        _nextChildBlock = _childBlockInterval;
+        _nextDepositBlock = 1;
+    }
+
+    function blocks(uint256 _blockNumber) public view returns (bytes32 root, uint256 timestamp) {
```
pik694: If you kept `_blocks` public, this getter would not be necessary.

plasma_framework/contracts/mocks/framework/registries/VaultRegistryMock.sol
boolafish: Sounds like a good idea. Would fix this Monday.

plasma_framework/contracts/src/framework/BlockController.sol
boolafish: Sound good idea to me, but I would like to keep the event to be same as existing one first. We could change after our first round of migration with elixir code.

plasma_framework/contracts/src/framework/BlockController.sol
boolafish: I am on the fence for this file actually. 

In general I prefer private so it promises any contract inherits this cannot accidentally set the value. So, for instance, the `operator` field for `Operated.sol` makes perfect sense to be private as it is inherited in several places and in deep layer. However, on the other hand, If it is a top level contract code that nobody inherits, there is almost no benefit to keep private as nothing is going to inherit it.

The case here to me is in a gray area. It is designed to be inherit by `PlasmaFramework` but functional wise, it should be the top layer code here. `PlasmaFramework` should simply only extends several modules to glue things into one contract. In this case, I chose to see how hard it would be to keep private in coding and testing. Since it is not too much more complex I chose private here. But for the case here I am pretty open to make it public.

plasma_framework/contracts/src/framework/BlockController.sol
boolafish: answered up in previous comment, would like to hear your opinion.

plasma_framework/contracts/mocks/framework/registries/VaultRegistryMock.sol
boolafish: fixed

plasma_framework/contracts/src/framework/BlockController.sol
pik694: Calling a method is rather more expensive than just retrieving the stored data, but now I understand your decision I think that for now it can be as it is :)

Issue #116, opened 2019-06-05
new python testing framework


pik694:
## Current stack
Our tests run under `pytest` testing framework. Each run of the tests we run an internal `EVM` provided by `pyethereum`. We compile and link our contracts using `py-solc` which wraps `solc`.

## Motivation for changes
Our main motivation is the fact that `py-solc` does not work with `solidity v0.5`. One other is that `pyethereum` is deprecated.

## Goal
We would like to have a testing framework that is capable of running `solidity v0.5`. The framework should also be future-proof (at least should look promising).

ref #103

pik694:
## Compilation
Our first thought was to use `truffle` as a pretty mature framework to compile the contracts, capable of compiling and deploying contracts - we do not need anything more. 

The first result of testing truffle is that we would not be able to deploy contracts using truffle as this would require us to have an external EVM (ex. `ganache`), but  we would prefer to stick to the internal EVM. Even though if we tried external EVM, our testing framework would be a Frankenstein - running some python code to run `truffle` to run js to compile and link contracts with `solcjs` to send them to `ganache`.

Therefore we have to abandon thoughts of using truffle to deploy contracts; but we can still use truffle to compile contracts, don't we? We do, but  we would have to find some tool to link the deployed libraries to the contracts - we would have to use/write a `solc` wrapper - this means we are no closer to the solution as we are seeking something that will replace our current `solc` wrapper.

pik694:
## Brownie
During the research I found a new framework - `brownie`, which was written to mimic `truffle`. It looks promising, but we would like to use more stable solutions. Nevertheless, it is worth keeping a note for the future.

pik694:
# Web3 with EthereumTesterProvider

## Stack
We have been experimenting with `web3py` and usage of `EthereumTesterProvider`. This provider uses [py-evm](https://github.com/ethereum/py-evm) underneath, which is a reference EVM implementation in python. 

In order to link and deploy contracts we used [pytest-ethereum](https://github.com/ethereum/pytest-ethereum), which provides a convenient `linker` and `deployer`. One minor drawback of using it is that it injects its own fixtures ex. web3. We had to disable those and create our own web3 instance, as we have had to change the default configuration of the EVM (`block_limit`). 
`pytest-ethereum` requires us to provide a package or a manifest at least.
We used [py-ethpm](https://github.com/ethpm/py-ethpm) to create a manifest from compilation results.

[...]

Issue #110, opened 2019-05-29
Solve SE output and IFE input collision without relying on exit id


boolafish:
### Why
Abstract design requires each exit games to be as independent with each others as possible. Current solution for the collision requires the knowledge of "exit id" of input transaction. However, the input transaction's "exit id" would be decided by its own game.  We do not want to be in the situation while designing exit game for new tx we need to take care of all input exit games' logic.

previous discussion: https://github.com/omisego/research/issues/78#issuecomment-492170043

**This is a blocker for abstract layer implementation, we need decision on this**

### Proposal
In non-canonical challenge, an IFE can be challenged by input that is already finalized, and this would have highest priority so that nothing can respond this.

Alternatives, [Kelvin's solution](https://github.com/omisego/plasma-contracts/issues/102#issuecomment-495809967) for #102 might just as well solve this?

### What to do
- [x] Basically, there must have dependency somewhere in MoreVP. Currently it is linked by exit id. We should also analyze is it possible to use `utxo pos` instead of `exit id`?
- [x] Check whether Kelvin's solution for #102 would just as well solve this

boolafish:
For the first point of `exit id` and `utxo pos`:

Okay, I think the problem is more about naming and variable responsibility.
Currently we use `exit id` for two things: 
1. A id to get essential data from `exits` mapping
1. A id to sort of represent output

Also, to check an output is finalized, we re-use the `exits` mapping + the fact that once finalized, the `owner` field would be cleaned up. So an existing `exits` mapping + empty owner indicates finalized output. (relevant code: [here](https://github.com/omisego/plasma-contracts/blob/master/contracts/RootChain.sol#L406))

The main issue for the abstract design is that the id to get essential data should not need to know about how other exit game decides their id for data. The id to represent output for sure would be a global logic.

Since we are having abstracted `vault` (originally named `wallet`), we can have a map there recording finalized outputs (with `utxo_pos` or `output id` as key). We can still check the input is finalized or not on IFE starts.

One thing would be hard to keep is that, current contract [automatically challenges the input SE](https://github.com/omisego/plasma-contracts/blob/master/contracts/RootChain.sol#L412). 

However, I would really like to argue it is more of a best practice to move this out to the responsibility of watcher instead. This simplifies the contract code and safe some gas on IFE starts.

boolafish:
re: Kelvin's solution (checking all inputs on processExit) impact on this:
[...]

PR #130, opened 2019-06-17
Exit game controller

### LINKED ISSUE #126: [abstract] Implement ExitGameController
parent story: #105.
Previous POC code: https://github.com/omisego/plasma-contracts/blob/abstract_poc/abstract_poc/contracts/src/framework/ExitGameController.sol

boolafish:
### Note
- Copy priority queue contracts to here. 
  - Upgrade to solidity 0.5
  - Use open-zeppelin library: Ownable, SafeMath
- Implement ExitGameController and unit tests

close #126 

### Side notes
If using Visual Studio Code, use the following setting to remove import error (for open-zeppelin).

```
"solidity.packageDefaultDependenciesContractsDirectory": "",
"solidity.packageDefaultDependenciesDirectory": "plasma_framework/node_modules"
```

### Test
truffle test
```
Contract: ExitGameController
    hasToken
      ✓ returns true when token already added (177ms)
      ✓ returns false when token not added (52ms)
    addToken
      ✓ rejects when token already added (174ms)
      ✓ generates a priority queue instance to the exitsQueues map (128ms)
      ✓ emits TokenAdded event (110ms)
    enqueue
      ✓ rejects when not called from exit game contract (150ms)
      ✓ rejects when such token has not been added yet (200ms)
      ✓ increases `exitQueueNonce` one (171ms)
      ✓ inserts the new unique priority to the queue (187ms)
      ✓ saves the exit data to map (160ms)
      ✓ can enqueue with the same exit priority multiple times (221ms)
    processExits
      ✓ rejects when such token has not been added yet (165ms)
      ✓ rejects when the exit queue is empty (115ms)
      ✓ rejects when the "top unique priority" mismatch with the specified one (182ms)
      ✓ processes when the "top unique priority" is set to 0 (183ms)
      ✓ processes when the "top unique priority" is set to the exact top one of the queue (289ms)
      ✓ calls the "processExit" function of the exit processor when processes (194ms)
      ✓ deletes the processed exit from queue after processed (236ms)
      ✓ deletes the exit data after processed (222ms)
      ✓ processes no more than the "maxExitsToProcess" limit (242ms)
      ✓ does not process exit that is not able to exit yet (162ms)
      ✓ stops to process when queue becomes empty (209ms)

Contract: ExitGameRegistry
    onlyFromExitGame
      ✓ accepts call when called by registered exit game contract (47ms)
      ✓ reverts when not called by registered exit game contract

Contract: PriorityQueue
    getMin
      ✓ can get the min value (66ms)
      ✓ fails when empty
    currentSize
      ✓ can get current size (40ms)
    insert
      ✓ can insert one value (60ms)
      ✓ can insert multiple values with order (122ms)
      ✓ can insert out of order and still get the minimal number (88ms)
      ✓ is not idempotent (85ms)
    delMin
      ✓ can delete when single value in queue (76ms)
      ✓ can delete when multiple values in queue (121ms)
      ✓ can delete all (141ms)
```


plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+
+        it('rejects when such token has not been added yet', async () => {
+            const fakeNonAddedTokenAddress = (await DummyExitGame.new()).address;
+            await expectRevert(
+                this.dummyExitGame.enqueue(0, fakeNonAddedTokenAddress, this.dummyExit),
+                "Such token has not be added to the plasma framework yet"
+            );
+        });
+
+        it('increases `exitQueueNonce` one', async () => {
```
pgebal: one -> once

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -0,0 +1,106 @@
...
+
+import "./models/ExitModel.sol";
+import "./interfaces/ExitProcessor.sol";
+import "./registries/ExitGameRegistry.sol";
+import "./priorityQueue/PriorityQueue.sol";
+
+contract ExitGameController is ExitGameRegistry {
+    uint64 public exitQueueNonce = 1;
+    mapping(uint256 => ExitModel.Exit) public exits;
+    mapping (address => PriorityQueue) public exitsQueues;
```
pgebal: Don't we want to keep those queues private?

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+                exitableAt: 1,
+                exitId: 123,
+            };
+        });
+
+        it('rejects when such token has not been added yet', async () => {
+            const fakeNonAddedTokenAddress = (await DummyExitGame.new()).address;
+            await expectRevert(
+                this.controller.processExits(fakeNonAddedTokenAddress, 0, 1),
+                "Such token has not be added to the plasma framework yet"
```
pgebal: be -> been

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        it('rejects when the exit queue is empty', async () => {
+            await expectRevert(
+                this.controller.processExits(this.dummyToken, 0, 1),
+                "Exit queue is empty"
+            );
+        });
+
+        it('rejects when the "top unique priority" mismatch with the specified one', async () => {
```
pgebal: mistmatch -> mismatches

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+            await this.dummyExitGame.enqueue(0, this.dummyToken, this.dummyExit);
+
+            const tx = await this.controller.processExits(this.dummyToken, 0, 1);
+            await expectEvent.inLogs(tx.logs, 'ProcessedExitsNum', {
+                processedNum: new BN(1),
+                token: this.dummyToken
+            });
+        });
+
+        it('processes when the "top unique priority" is set to the exact top one of the queue', async () => {
```
pgebal: top one of -> top of

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        it('calls the "processExit" function of the exit processor when processes', async () => {
+            await this.dummyExitGame.enqueue(0, this.dummyToken, this.dummyExit);
+
+            await this.controller.processExits(this.dummyToken, 0, 1);
+
+            // 'exitIdProcessed' would be set when processExit is called in the DummyExitGame
+            const exitIdProcessed = await this.dummyExitGame.exitIdProcessed();
+            expect(exitIdProcessed).to.be.bignumber.equal(new BN(this.dummyExit.exitId));
```
pgebal: I would rather test that by expecting an event that exit has been processed.
I think we are going to deep in implementation details here.

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        it('deletes the processed exit from queue after processed', async () => {
+            await this.dummyExitGame.enqueue(0, this.dummyToken, this.dummyExit);
+
+            await this.controller.processExits(this.dummyToken, 0, 1);
+
+            const priorityQueueAddress = await this.controller.exitsQueues(this.dummyToken);
+            const priorityQueue = await PriorityQueue.at(priorityQueueAddress);
+            expect(await priorityQueue.currentSize()).to.be.bignumber.equal(new BN(0));
```
pgebal: Like in previous comment. I think it would be better to test that through controller API. 
Check whether calling `processExits` for the second time fails with expected error message.

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -0,0 +1,268 @@
...
+        });
+
+        it('deletes the exit data after processed', async () => {
+            await this.dummyExitGame.enqueue(0, this.dummyToken, this.dummyExit);
+            const uniquePriority = await this.dummyExitGame.uniqueuPriorityFromEnqueue();
+
+            await this.controller.processExits(this.dummyToken, 0, 1);
+
+            const exit = await this.controller.exits(uniquePriority);
+            expect(exit.exitProcessor).to.equal(constants.ZERO_ADDRESS);
```
pgebal: If I remember correctly, for reason of dependency between standard and in-flight exits we kept some data about exits in contract. Please give it a second thought if it's still the case in our new design. I am not able to do that now. 
On the other hand, we can revert that change later, so do as you wish.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -0,0 +1,106 @@
...
+     * @param _priority The priority of the exit itself
+     * @param _token Token for the exit
+     * @param _exit Exit data that contains the basic information for exit processor the process the exit
+     * @return a unique priority number computed for the exit
+     */
+    function enqueue(uint192 _priority, address _token, ExitModel.Exit memory _exit) public onlyFromExitGame returns (uint256) {
+        require(hasToken(_token), "Such token has not be added to the plasma framework yet");
+
+        PriorityQueue queue = exitsQueues[_token];
+        uint256 uniquePriority = (uint256(_priority) << 64 | exitQueueNonce);
```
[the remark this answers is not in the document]
boolafish: The idea here is that to make each ExitGame contract as independent as each others' logic as possible. With this nonce from the framework level, no matter the priority conflicts in two ExitGame contracts or not, it would work fine to the queue. Otherwise the ExitGame would need to check how all other contracts choose their priority and make sure the logic fits in.

plasma_framework/contracts/mocks/framework/DummyExitGame.sol
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    function setExitGameController(address _contract) public {
+        exitGameController = ExitGameController(_contract);
+    }
+
+    function enqueue(uint192 _priority, address _token, ExitModel.Exit memory _exit) public {
+        uniqueuPriorityFromEnqueue = exitGameController.enqueue(_priority, _token, _exit);
+    }
+}
```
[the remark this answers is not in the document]
boolafish: Agree. New to VS code so haven't setup auto end of file new line yet.
Just did it: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

plasma_framework/test/src/framework/ExitGameController.test.js
boolafish: There would be two kinds of "exit data", one in framework layer which is this one that only contains pointer for the framework to know which exit processor to call and some minimum info to use for "process exit". 

Another is the real exit data that contains all the exit info that would be stored in ExitGame contract.

For the interaction of SE and IFE, if conflicting with same output of same tx, that should be able to handled within the ExitGame contract as both data exists in the same contract. The one with issue would be SE conflicting IFE input because there is potential cross ExitGame contract data dependency. And that is why this issue: https://github.com/omisego/plasma-contracts/issues/110.

I prefer to solve the conflict by using "Vault" that flags finalized output instead of using exit mapping to represent the finalized output (as proposed in the issue above) since Vault is global in a way anyway and it makes sense that a vault knows what output has been finalized there.  This comes with the tradeoff of gas cost but make each ExitGame contract more independent and don't need to be aware of other contract implementation detail.

plasma_framework/contracts/src/framework/ExitGameController.sol
boolafish: haha, was being lazy and this would be top contract anyway (supposedly only `PlasmaFramework` would inherit this and would be nothing more than a contract that simply inherits multiple contract to get all functions)

updated with private vars.

plasma_framework/contracts/src/framework/priorityQueue/PriorityQueueLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        self.currentSize = 0;
+    }
+
+    function insert(Queue storage self, uint256 k) public {
+        self.heapList.push(k);
+        self.currentSize = self.currentSize.add(1);
+        percUp(self, self.currentSize);
+    }
+
+    function minChild(Queue storage self, uint256 i) public view returns (uint256) {
```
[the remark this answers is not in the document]
boolafish: move to private function as it is used within the contract actually.

plasma_framework/contracts/src/framework/priorityQueue/PriorityQueueLib.sol
```diff
@@ -0,0 +1,101 @@
...
+        address owner;
+        uint256[] heapList;
+        uint256 currentSize;
+    }
+
+    /*
+     *  Public functions
+     */
+
+    function isOwner(Queue storage self) internal view returns (bool) {
```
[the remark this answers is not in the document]
boolafish: nice catch!

plasma_framework/test/src/framework/ExitGameController.test.js
boolafish: The test description here is:
>deletes the processed exit from queue after processed

It feels to me more straight forward by checking priority queue directly here. Check priority queue empty via re-call the function and would fail relies on the knowledge of "the function would fail when the priority queue is empty". Feel like one more step of implementation knowledge to reach there actually. 

How do you think?

plasma_framework/test/src/framework/ExitGameController.test.js
boolafish: Not exactly sure what your recommendation here would be. I make `DummyExitGame` to emit a `ExitFinalizedFromDummyExitGame` event and test that instead of setting exitId variable. Is this what you mean here?

boolafish:
@pgebal 
re: adding to CI

Have we finalized our decision on JS testing? I am sort of waiting the result of https://github.com/omisego/plasma-contracts/issues/116. From my last understanding it seems like people are more and more agreed on using JS because of tooling and to be consistent with deployment tool we are going to use.

@pik694 do you have any idea what's the current state of the python testing? And what's your opinion on the two side since you've probably has most insight on both.

plasma_framework/test/src/framework/ExitGameController.test.js
pgebal: You are right on that. Test matches the description. But I have a feeling that we could test on a higher level. The things we care about is that exits are processed, none is processed twice, they are processed with respect to their priority, limit numbers and arguments values have effect. Priority queue underneath is an implementation detail. But on the other hand maybe that matters in security critical components, like smart contracts. Let's see what @paulperegud thinks of that.

paulperegud: By checking current size we are checking if size has changed, not if value is no longer in the queue. I would just try to `getMin()` next value out of the queue and observe call getting reverted by EVM.

Also - please check the old tests for priority queue. They cover a lot.

plasma_framework/test/src/framework/ExitGameController.test.js
pgebal: My bad, it's ok.

plasma_framework/test/src/framework/ExitGameController.test.js
boolafish: @pgebal I see your point now. I like your description about the things we care about:
> exits are processed, none is processed twice, they are processed with respect to their priority

I changed the test to be testing
> processes with the order of priority and delete the processed exit from queue

So I first put two things to the queue and check after process the top (`getMin`) is the lower priority one + the size decreases.

Issue #131, opened 2019-06-18
Solve security alert on `diff` package version


boolafish:
### Note
There is a [security alert](https://github.com/omisego/plasma-contracts/network/alert/package-lock.json/diff/open) on dependency `diff` package. It is recommended to upgrade to 3.5.0.

However, it is a dependency from truffle. I tried to upgrade the truffle to latest 5.0.22 and no luck. Still using mocha@4.1.0 and this diff@3.3.1.

npm ls diff
```
└─┬ truffle@5.0.21
  └─┬ mocha@4.1.0
    └── diff@3.3.1
```

I've added a comment to truffle issue on mocha version: https://github.com/trufflesuite/truffle/issues/1705

### What can we do
not sure what is best practice in this case now.
1. Override the version: https://stackoverflow.com/a/48524488
   - use [npm-force-resolutions](https://github.com/rogeriochaves/npm-force-resolutions) to update the package.
1. Wait for truffle to upgrade the mocha.

It seems like low security issue as mocha is only used for testing (?)

boolafish:
Anyway, I sent a [PR](https://github.com/omisego/plasma-contracts/pull/132). and as shown in the PR, the solution causes another warning : (

I don't really think dependency of mocha could break any of our stuff but would like to ask whether we should solve this and should we save the way in the PR or.....?

Another thing (not really related to this issue) is that do we want truffle as "dependencies" or "dev-dependencies" (?) cc. @kevsul @jbunce

boolafish:
As from @kevsul 's comment: https://github.com/omisego/plasma-contracts/pull/132#issuecomment-503001105

Should be no impact to us. Dismiss the security alert.

PR #132, opened 2019-06-18
fix: security alert for package 'diff'

boolafish:
### Note
We get a security alert and this commit aims to fix it.
This uses 'npm-force-resolutions' to fix the conflict.

- alert link: https://github.com/omisego/plasma-contracts/network/alert/package-lock.json/diff/open
- npm-force-resolutions: https://github.com/rogeriochaves/npm-force-resolutions
- close issue: https://github.com/omisego/plasma-contracts/issues/131

### My Step to fix
1. npm install -g npm-force-resolutions
1. add `resolutions` to package.json
1. npm-force-resolutions
1. npm install

### Test
```
➜  plasma-contracts git:(fix_security_alert) npm ls diff
plasma-contracts@1.0.0 /Users/boolafish/GitHub/plasma-contracts
└─┬ truffle@5.0.21
  └─┬ mocha@6.1.4 invalid
    └── UNMET DEPENDENCY diff@3.5.0
```

However, as shown, this fix would lead to `invalid` and `UNMET DEPENDENCY` warnings during `npm ls`

**So I don't know is this the right way to do or not.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -176,11 +176,6 @@
       "resolved": "https://registry.npmjs.org/brorand/-/brorand-1.1.0.tgz",
       "integrity": "sha1-EsJe/kCkXjwyPrhnWgoM5XsiNx8="
     },
-    "browser-stdout": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/browser-stdout/-/browser-stdout-1.3.0.tgz",
-      "integrity": "sha1-81HTKWnTL6XXpVZxVCY9korjvR8="
-    },
     "browserify-aes": {
       "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


kevsul:
This vulnerability doesn't affect us. 
It's a potential DOS attack, which is a vulnerability _if_ you're running a service and an attacker can somehow send arbitrary data to be `diff`-ed, but that doesn't apply to this repo.

Usually the correct way to update packages to fix vulnerabilities is to run
```
npm audit fix
```
but it doesn't fix this `diff` issue, so I imagine npm and github's vulnerability lists are out of sync.

Anyway, I don't think it's an issue for us. We can just dismiss the security alert.

boolafish:
Agree, thanks for confirm!

PR #136, opened 2019-06-27
feat: glue components into PlasmaFramework.sol

boolafish:
### Note
This commit combines all existing compoenents into the final
PlasmaFramework code. Also add a missing feature in constructor
of ExitGameController.

### Test
truffle test
```
  Contract: ExitGameController
    constructor
      ✓ inits the queue for ETH
  Contract: PlasmaFramework
    constructor
      ✓ can set the min exit period (182ms)
```

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -20,6 +20,12 @@ contract ExitGameController is ExitGameRegistry {
         address token
     );
 
+    constructor() public {
+        address ethToken = address(0);
+        address queueOwner = address(this);
+        _exitsQueues[ethToken] = new PriorityQueue(queueOwner);
+    }
+
     function exitQueueNonce() public view returns (uint64) {
         return _exitQueueNonce;
     }
@@ -38,6 +44,7 @@ contract ExitGameController is ExitGameRegistry {
 
     /**
      * @notice Add token to the plasma framework and initiate the priority queue.
+     * @notice ETH token is supported by default on deployment.
      * @dev the queue is created as a new contract instance.
      * @param _token Address of the token.
      */
--- a/plasma_framework/contracts/src/framework/PlasmaFramework.sol
+++ b/plasma_framework/contracts/src/framework/PlasmaFramework.sol
@@ -1,0 +1,20 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "./BlockController.sol";
+import "./ExitGameController.sol";
+import "./registries/VaultRegistry.sol";
+import "./registries/ExitGameRegistry.sol";
+import "./utils/Operated.sol";
+
+contract PlasmaFramework is Operated, VaultRegistry, ExitGameRegistry, ExitGameController, BlockController {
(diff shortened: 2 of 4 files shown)


plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -38,6 +44,7 @@ contract ExitGameController is ExitGameRegistry {
 
     /**
      * @notice Add token to the plasma framework and initiate the priority queue.
+     * @notice ETH token is by default supported while deployment.
```
kevsul: I think this reads better as
`ETH token is supported by default on deployment`

plasma_framework/contracts/src/framework/PlasmaFramework.sol
```diff
@@ -0,0 +1,19 @@
...
+import "./BlockController.sol";
+import "./ExitGameController.sol";
+import "./registries/VaultRegistry.sol";
+import "./registries/ExitGameRegistry.sol";
+import "./utils/Operated.sol";
+
+contract PlasmaFramework is Operated, VaultRegistry, ExitGameRegistry, ExitGameController, BlockController {
+    uint256 constant CHILD_BLOCK_INTERVAL = 1000;
+
+    // NOTE: this is the "middle" period. Exit period for fresh utxos we'll double that while IFE phase is half that
```
kevsul: Exit period for fresh utxos _is_ double that

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -20,7 +20,15 @@ contract('ExitGameController', () => {
         this.dummyTxType = 1;
         this.controller.registerExitGame(this.dummyTxType, this.dummyExitGame.address);
 
-        this.dummyToken = constants.ZERO_ADDRESS;
+        // take any random contract address as token
+        this.dummyToken = (await DummyExitGame.new()).address;
+    });
+
+    describe('constructor', () => {
+        it('inits the queue for ETH', async () => {
```
kevsul: Standard practice when using Mocha is to start the test description with 'should...' e.g.
`it('should init the queue for ETH'`
It just reads better, but it's not a big deal if you prefer not to do that.

plasma_framework/test/src/framework/PlasmaFramework.test.js
```diff
@@ -0,0 +1,20 @@
...
+const { expect } = require('chai');
+
+contract('PlasmaFramework', () => {
+    before('link library', async () => {
+        const priorityQueueLib = await PriorityQueueLib.new();
+        await PlasmaFramework.link("PriorityQueueLib", priorityQueueLib.address);
+    });
+
+    describe('constructor', () => {
+        it('can set the min exit period', async () => {
```
kevsul: Again, `it('should set...` if you want to follow the convention

plasma_framework/test/src/framework/ExitGameController.test.js
boolafish: thanks, didn't know about that!

boolafish:
all @kevsul 's comment fixed!

PR #137, opened 2019-06-27
fix: alter visibility for PriorityQueue(Lib)

boolafish:
### Note
Previously PriorityQueueLib was using 'public' instead of 'internal'. This would make a 'delegetecall' instead of a 'jump' under the hood. And would be more expensive on gas.

Also, changes 'public' to 'external' for PriorityQueue as it is not needed to be public and external sometimes can safe a bit gas if input is complex.

### Test
truffle test


pdobacz:
hm, I recall that the `internal` were removed to not have the calls inlined and have them be resolved during linking. This was done to reduce the gas cost of the calling contract deployment (or rather - split it into smaller chunks).

C.f. https://github.com/omisego/plasma-contracts/pull/38/commits/e63fc5e689486ef6600012517550c7fe79f1fc18 and https://github.com/ethereum/solidity/issues/3985

(disclaimer: I'm not 100% sure this has been the right move and is still necessary, but that was the point why `PriorityQueueLib` existed in the 1st place. We did that so that `RootChain.sol` deployment fit in the block gas limit)

boolafish:
@pdobacz thanks! was not aware that was intended to be public.

I agree and believe this would have more deployment cost as I assume the reason it can use `jump` must be it is pulling the code together. However, as the new design would split contracts into multiple parts, I am so far **hoping** that can fit into deployment gas cost. Also, the framework layer itself is pretty light (I think). Most complex code would still be in ExitGame.

I would still like to give it a try if we are able to deploy with this as this saves gas on every action. But yeah, if it just does not work at the end then I will revert then.

pdobacz:
> However, as the new design would split contracts into multiple parts, I am so far hoping that can fit into deployment gas cost.

Gotcha! if it does, then one can (possibly) ditch the libraries altogether. Monitor the deployment gas and we'll see how it goes.

boolafish:
I take in @pdobacz 's suggestion of merging lib back and make it a single file. I guess even we hit gas limit our first attempt would be using `init()` instead of constructor. We have really low chance to need to put the PriorityQueueLib separately I guess so might just ditch the lib now.

@pgebal can you help me review the new git commit?
I also found I have some unclean porting on the `owner` to be using `Ownable` from open zeppelin. I clean those up again.

plasma_framework/contracts/src/framework/utils/PriorityQueue.sol
```diff
@@ -1,53 +1,69 @@
...
 
-    function delMin(Queue storage self) internal returns (uint256) {
-        uint256 retVal = self.heapList[1];
-        self.heapList[1] = self.heapList[self.currentSize];
-        delete self.heapList[self.currentSize];
-        self.currentSize = self.currentSize.sub(1);
-        percDown(self, 1);
-        self.heapList.length = self.heapList.length.sub(1);
+    /**
+     * @notice Deletes the top element of the heap and shifts everything up by the owner.
```
pgebal: I think `Deletes the smallest element from the queue` would be enough.

plasma_framework/contracts/src/framework/utils/PriorityQueue.sol
```diff
@@ -1,53 +1,69 @@
...
+        queue.currentSize = queue.currentSize.sub(1);
+        percDown(queue, 1);
+        queue.heapList.length = queue.heapList.length.sub(1);
         return retVal;
     }
 
-    function getCurrentSize(Queue storage self) internal view returns (uint256) {
-        return self.currentSize;
+    /**
+     * @notice Returns the top element of the heap.
```
pgebal: Returns the smallest element from the queue.
The comment you put would be more suitable for `@dev` annotation.

pgebal:
There is a typo in the last commit message. defauly -> default

plasma_framework/contracts/src/framework/utils/PriorityQueue.sol
boolafish: change the @notice to your sentence.
I did not add @dev to keep consistency with delMin and popping from top of heap is pretty much a common sense I guess.

PR #138, opened 2019-06-27
Replace py solc with py solcx

pik694:
Upgrades `py-solc`simple`, which effectively replaces `py-solc` with `py-solc-x` which is capable of compiling Solidity v0.5


setup.py
```diff
@@ -27,6 +27,14 @@
         'ethereum==2.3.0',
         'web3==4.8.2',
         'rlp==0.6.0',
-        'py-solc-simple==0.0.13'
-    ]
+        'py-solc-simple==0.0.14',
+        'eip712-structs==1.1.0'
```
boolafish: curious, why we don't need this before?

pik694: We did, but previously it was mentioned only in the requirements.txt file

PR #139, opened 2019-06-28
Porting libraries: Merkle, UtxoPos

boolafish:
### Note
- port `Merkle.sol` and add related test (test code thanks to @pik694 's POC code)
- Although there is a [Merkle lib in open-zepplin](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/bd13be9174943f3fa18896df4ac269f411143c7d/contracts/cryptography/MerkleProof.sol), but their code require the leaves to be sorted :(. So I don't think we can use that.
- add new lib `UtxoPosLib.sol` to handle utxo pos related functions. Originally in `PlasmaCore.sol`.

### Side note
found open zeppelin's latest `ERCRecover` (they renamed `ECDSA`), so we don't need to port that :)
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/bd13be9174943f3fa18896df4ac269f411143c7d/contracts/cryptography/ECDSA.sol
### Test
truffle test
```
Contract: Merkle
    checkMembership
      ✓ should return true when proven included (41ms)
      ✓ should return false when not able to prove included
      ✓ should reject call when proof data size is incorrect

  Contract: UtxoPosLib
    blockNum
      ✓ should parse the correct block number
    txIndex
      ✓ should parse the correct tx index
    outputIndex
      ✓ should parse the correct output index
    txPos
      ✓ should parse the correct tx position
```

### DIFF
--- a/plasma_framework/test/src/utils/Merkle.test.js
+++ b/plasma_framework/test/src/utils/Merkle.test.js
@@ -1,0 +1,49 @@
+const Merkle = artifacts.require('MerkleWrapper');
+
+const { expectRevert } = require('openzeppelin-test-helpers');
+const { expect } = require('chai');
+
+const { MerkleTree } = require('../../helpers/merkle.js');
+
+contract('Merkle', () => {
+    before('setup merkle contract and tree value', async () => {
+        this.merkleContract = await Merkle.new();
+        this.leaves = ['leaf 1', 'leaf 2', 'leaf 3'];
+        this.merkleTree = new MerkleTree(this.leaves);
+    });
+
+    describe('checkMembership', () => {
+        it('should return true when proven included', async () => {
+            const leafIndex = 0;
+            const leafData = web3.utils.sha3(this.leaves[leafIndex]);
+            const rootHash = this.merkleTree.root;
+            const proof = this.merkleTree.getInclusionProof(this.leaves[leafIndex]);
+
+            const result = await this.merkleContract.checkMembership(leafData, leafIndex, rootHash, proof);
+            expect(result).to.be.true;
+        });
+
+        it('should return false when not able to prove included', async () => {
+            const leafIndex = 0;
+            const leafData = web3.utils.sha3(this.leaves[leafIndex]);
+            const fakeRootHash = web3.utils.sha3('random root hash');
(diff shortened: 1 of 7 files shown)


plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -0,0 +1,49 @@
...
+        it('should reject call when proof data size is incorrect', async () => {
+            const leafIndex = 0;
+            const leafData = web3.utils.sha3(this.leaves[leafIndex]);
+            const rootHash = this.merkleTree.root;
+            const proof = this.merkleTree.getInclusionProof(this.leaves[leafIndex]);
+            const wrongSizeProof = proof + '13212';
+
+            await expectRevert(
+                this.merkleContract.checkMembership(leafData, leafIndex, rootHash, wrongSizeProof),
+                "Length of merkle proof must be some multiplication of 32."
```
pgebal: be some multiplication -> be a multiple

plasma_framework/test/src/utils/UtxoPosLib.test.js
```diff
@@ -0,0 +1,47 @@
+const UtxoPosLib = artifacts.require('UtxoPosLibWrapper');
+
+const { BN } = require('openzeppelin-test-helpers');
+const { expect } = require('chai');
+
+contract('UtxoPosLib', () => {
+    beforeEach('setup contract and utxo pos values', async () => {
```
pgebal: Same as for MerkleTree.

plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -0,0 +1,49 @@
+const Merkle = artifacts.require('MerkleWrapper');
+
+const { expectRevert } = require('openzeppelin-test-helpers');
+const { expect } = require('chai');
+
+const { MerkleTree } = require('../../helpers/merkle.js');
+
+contract('Merkle', () => {
+    beforeEach(async () => {
```
pgebal: beforeEach -> before. We are testing pure functions here, so no need to deploy each time.

boolafish: haha, you're really picky at this XDD

boolafish:
updated with the CR. merged.

PR #140, opened 2019-06-28
124 implement erc20vault

### LINKED ISSUE #95: Contract should be able to support ERC20 without return value
## Issue Type 
As OMG contract is implemented in a way that does not return "true" when transfer, as we want to be able to support OMG token, our contract need to support ERC20 `transferFrom` function without a return value.

For details, see this: https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca

```
[ ] bug report
[x] feature request
```

## Current Behavior
In ` function depositFrom(bytes _depositTx)`

```
// Warning, check your ERC20 implementation. TransferFrom should return bool
require(ERC20(decodedTx.outputs[0].token).transferFrom(msg.sender, address(this), decodedTx.outputs[0].amount));
```

## Expected Behavior
Does not need the `transferFrom` to return bool.

## Suggested Fix
Perhaps use `address.call` directly, and in solidity 0.5, it returns success or failed + return value. We can rely on the result of "success" or "failed' directly instead.

Or the article above (in Issue Type) do provide some wrapper example code as well.

## Motivation for Change
To support OMG token

### LINKED ISSUE #124: [abstract] Implement Erc20Vault
parent story: #105

kevsul:
### Note
- Implements Erc20Vault
- Uses open-zepplin's SafeERC20 lib to handle old implementations of ERC20 standard
- I've refactored common code into Vault.sol under the assumption that all deposit transactions will have the same format and some common validations. If that assumption is wrong I can revert...

### Test
  Contract: Erc20Vault
    deposit
      ✓ should store erc20 deposit (242ms)
      ✓ should spend erc20 tokens from depositing user (227ms)
      ✓ should not store a deposit when the tokens have not been approved (159ms)
      ✓ should not store a deposit from user who does not match output address (122ms)
      ✓ should not store an ethereum deposit that sends funds
      ✓ should not store an ethereum deposit that does not send funds (138ms)
      ✓ should not accept transaction that does not match expected transaction type (117ms)
      ✓ should not accept transaction that does not conform to deposit input format (116ms)
      ✓ should not accept transaction with more than one output (147ms)
    deposit from BadERC20
      ✓ should store erc20 deposit (204ms)

[edit: close #124 , close #95]


boolafish:
Not a real CR, just replying the assumption of re-use.

> _validateDepositFormat(DepositTx.Transaction memory _deposit)

This is not reusable, as different vault would have different depositTx actually. eg, current TX structure would not be able to support ERC721.

In fact, I am thinking about making the "validate format" part done by some injectable code/predicate. So we can be able to swap deposit tx following our tx version upgrade. 

Here is my POC code: https://github.com/boolafish/plasma-contracts/blob/vault_poc/plasma_framework/contracts/src/vaults/EthVault.sol#L30

kevsul:
Yeah, I know that _some_ parts of the transaction structure will be different depending on the type of token, but my point is that other parts will always be the same e.g.
- there can only be one input
- the input must be bytes32 of 0
- there can only be one output
- the address of the sender must match the address of the output

so these common checks can go in the base class to avoid duplication of code.

> In fact, I am thinking about making the "validate format" part done by some injectable code/predicate. So we can swap deposit tx following our tx version upgrade.

You're right - if we're going to upgrade the tx format, then this will have to be upgradeable as well.

boolafish:
> but my point is that other parts will always be the same

I agree with you those logic would quite likely be always the same, but I don't know how to code that part to be not aware of deposit tx type though. Like you basically need to define which `DepositTx.Transaction` to use in your code, and you cannot really do things like "interface" for struct.

plasma_framework/contracts/src/vaults/Erc20Vault.sol
```diff
@@ -0,0 +1,47 @@
+pragma solidity ^0.5.0;
+
+import "./Vault.sol";
+import {PaymentTransactionModel as DepositTx} from "../transactions/PaymentTransactionModel.sol";
+import {IERC20 as IERC20} from "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
```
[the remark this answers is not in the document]
kevsul: I was following these guidelines:
```
This form is not recommended for use, because it unpredictably pollutes the namespace. 
```

from here https://solidity.readthedocs.io/en/v0.5.9/layout-of-source-files.html#importing-other-source-files

but I actually agree with you and prefer the simpler form of
```
import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
```

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -0,0 +1,138 @@
...
+const ERC20 = artifacts.require('ERC20Mintable');
+const BadERC20 = artifacts.require('BadERC20');
+
+const { BN, expectRevert } = require('openzeppelin-test-helpers');
+const { expect } = require('chai');
+
+const Testlang = require('../../helpers/testlang.js');
+const { PaymentTransaction, PaymentTransactionOutput } = require('../../helpers/transaction.js');
+
+contract('Erc20Vault', (accounts) => {
```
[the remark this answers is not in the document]
kevsul: I do like this, but then the linter complains about `_` being an unused variable 😞 
It would be nice if it followed Elixir's standard of ignoring vars that start with `_`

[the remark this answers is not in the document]
boolafish: ah...okay : ( didn't aware about that.

plasma_framework/contracts/src/vaults/Erc20Vault.sol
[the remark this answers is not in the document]
boolafish: ah....did not know about that.

How about this one then? We can start changing our convention on solidity import further on.
```
import "filename" as symbolName;
```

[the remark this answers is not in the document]
kevsul: But then you have to use  `symbolName.symbolName` in the code, which is kind of ugly. 
E.g. if you do:
```
import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol" as IERC20;
```
then `IERC20` refers to the file scope and you need to use `IERC20.IERC20` to refer to the exported contract.

To be honest, I would just stick with using 
```
import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
```

I understand _why_ it's not recommended, but I think the vast majority of contracts are just going to export one global object - the contract itself.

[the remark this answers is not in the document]
boolafish: oh... :( again wasn't aware we need to `symbolName.symbolName`....
Okay let's stick with our existing version then XDD : )

pgebal:
@kevsul we've been following that: https://omisego.atlassian.net/wiki/spaces/ENG/pages/63111218/Commit+Message+Conventions+OMG+Network for naming commits. It's pinned on dev-plasma channel. I think it would be nice if we could stick to that convention.

boolafish:
just a friendly reminder, I think the discussion about "reuse code" is still on?

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -0,0 +1,139 @@
...
+        // as described here https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca
+        // Erc20Vault should support both versions.
+        before('setup', async () => {
+            this.badErc20 = await BadERC20.new(INITIAL_SUPPLY);
+            await this.badErc20.transfer(alice, DEPOSIT_VALUE, { from: accounts[0] });
+        });
+
+        it('should store erc20 deposit', async () => {
+            await this.badErc20.approve(this.erc20Vault.address, DEPOSIT_VALUE, { from: alice });
+            let nextDepositBlock = parseInt(await this.blockController.nextDepositBlock(), 10);
```
pgebal: Let's compare expected and actual deposit blocks as in `should store erc20 deposit` test.

pgebal:
@boolafish I like "composition over inheritance" approach. I could have used that in my PR (that PoC piece of code you posted above).

kevsul:
I agree with moving the verification code out into a `depositVerifier` as in the POC, as it will give us 'future proofing' for if we need to upgrade the transaction format.
But I don't see how it helps with code reuse? Unless each verifier shares common code?

boolafish:
I think those are actually two things, for `depositVerifier` stuff I .... don't know how we can reuse any of the code due to tx structure limit actually.

For general reuse, I just want to start discussion probably we would prefer composite over inheritance and not just use `Vault` as a place to reuse code.

kevsul:
Ah, ok.
I think that there may be _some_ cases when inheritance makes sense, but we can have a general guideline of preferring composition.
Is the idea to allow upgradeability whenever possible?

boolafish:
> Is the idea to allow upgradeability whenever possible?

I guess case by case, for the verifier that is for upgradability. For other cases I guess just which coding style we like more. Current reusable logic is pretty thin (constructor and submit block), which might be fine. I am just concern about opening this door would make us start throwing more and more logics there in the future without thinking twice.

plasma_framework/contracts/src/vaults/predicates/IErc20DepositVerifier.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import {PaymentTransactionModel as DepositTx} from "../../transactions/PaymentTransactionModel.sol";
+
+contract IErc20DepositVerifier {
```
kevsul: This is an abstract contract instead of an interface because:
- If it was an interface, then the `verify` method would need to be declared `external`
- and that would mean that the `_depositTx` param would need to be declared `calldata`
- ... but this causes an `UnimplementedFeatureError` from the solidity compiler.

There's no more info, but I imagine it has to do with `pragma experimental ABIEncoderV2`

boolafish: yes, it seems like struct + public would crash if the caller does not state "pragma experimental ABIEncoderV2".
And sadly there would not be warning error that point you where to fix :(
https://github.com/ethereum/solidity/issues/5659

plasma_framework/contracts/src/vaults/predicates/IErc20DepositVerifier.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import {PaymentTransactionModel as DepositTx} from "../../transactions/PaymentTransactionModel.sol";
```
boolafish: we cannot hardcode our tx model dependency in interface, otherwise we are in the same situation that we cannot upgrade the tx type for deposit. Then there would be really less benefit to put an interface on the verifier.

That is why my previous interface for ETH was taking raw bytes as input, and it is the implementation decide which tx model to use to decode.

plasma_framework/contracts/src/vaults/Erc20Vault.sol
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @notice Deposits approved amount of ERC20 token. Approve must have been called first.
+     * @param _depositTx RLP encoded transaction to act as the deposit.
+     */
+    function deposit(bytes calldata _depositTx) external {
+        DepositTx.Transaction memory decodedTx = DepositTx.decode(_depositTx);
+
+        _depositVerifier.verify(decodedTx, msg.sender);
+
+        IERC20 erc20 = IERC20(decodedTx.outputs[0].token);
```
boolafish: As mentioned in another comment, we cannot hardcode our tx model dependency here if we want to have the benefit to upgrade to new deposit tx type. In the case here, I guess to be able to do so we have to return some value from the "verifier" interface.

eg. (pseudo code)
```
(owner, token, amount) = depositVerifier.verify(_depositTx, msg.sender); 
erc20 = IERC20(token);
erc20.safeTransferFrom(msg.sender, address(this), amount);
```

btw, I feel 
```
require(erc20.allowance(msg.sender, address(this)) == decodedTx.outputs[0].amount, "Tokens have not been approved");
```
can be done inside verifier?

plasma_framework/contracts/src/vaults/predicates/IErc20DepositVerifier.sol
kevsul: Understood, thanks 👍

plasma_framework/contracts/src/vaults/predicates/IErc20DepositVerifier.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+interface IErc20DepositVerifier {
+    /**
+     * @notice Verifies a deposit transaction.
+     * @param _depositTx The deposit transaction.
+     * @param _sender The owner of the deposit transaction.
```
boolafish: `_vault` is missing

plasma_framework/contracts/src/vaults/EthVault.sol
```diff
@@ -24,15 +17,10 @@ contract EthVault is Vault {
...
-        for (uint i = 0; i < 16; i++) {
-            root = keccak256(abi.encodePacked(root, zeroHashes[i]));
-        }
-
-        blockController.submitDepositBlock(root);
+        super._submitDepositBlock(_depositTx);
     }
 
-    function _validateDepositFormat(DepositTx.Transaction memory _deposit) private {
+    function _validateDepositFormat(DepositTx.Transaction memory _deposit) internal view {
```
boolafish: minor: why do we move to `internal`? is this something accidentally changed?

kevsul: Yeah, that was accidental, I'll change it back.
This method will disappear anyway...

PR #153, opened 2019-07-05
refactor: optimize gas cost when starting standard payment exit

pgebal:
Applying gas cost optimizations suggested by @kevsul:
 - change order of verifying that start of payment exit is valid, call most costly `require` as the last one
 - do not use a temporary memory variable and put a structure directly into storage

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
+++ b/plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
@@ -115,25 +115,24 @@ contract PaymentStandardExitable is OnlyWithValue {
         private
         view
     {
+        require(data.output.amount > 0, "Should not exit with amount 0");
+        require(data.output.owner() == msg.sender, "Only output owner can start an exit");
+        require(exits[data.exitId].exitable == false, "Exit already started");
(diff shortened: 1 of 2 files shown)


plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
```diff
@@ -116,24 +116,23 @@ contract PaymentStandardExitable is OnlyWithValue {
         view
     {
         bytes32 leafData = keccak256(data.rlpTx);
```
kevsul: This line could be moved down as well, to avoid hashing until necessary

pgebal: done, amended the commit

boolafish:
@pgebal 
not really related to this PR, but do you mind to do a flavor to remove this line as well if you two are working to merge this soon: https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/src/exits/utils/ExitableTimestamp.test.js#L21

I just see the test longer than 2 secs and realized I forgot to remove my test code on the block timestamp issue ><"

pgebal:
@boolafish sure, done, I should have catch that when doing CR

PR #155, opened 2019-07-08
draft: output guard + output predicate in challenge

boolafish:
### Note
**This is now a draft for discussion purpose! (would keep on working to make it review ready, however, want early feedbacks)** 

This draft shows what I think how we can do `outputGuard` and `output predicate` (I named it `SpendingCondition` in the code here).

#### Output type and registries
There are two registries here: Output guard parser + Spending Condition (Output Predicate)

Instead of having those in PlasmaFramework, I think it is fair to just register these two per tx type. First, it saves some gas when getting data, also it mitigate the area how these two registries can impact. This makes `output type` only unique under same `tx type`. I think it is totally fine and good.

#### Output guard
The basic logic does the following:
1. if `outputType` == 0 --> `output.outputGuard` hold owner address directly.
1. else --> parse the output data from `outputGuardParser`

We need to be able to get the `outputGuardParser`, thus a registry for `outputGuardParser`.
So far I only put one function (`parseExitTarget`) in and I think this `IOutputGuardParser` interface should be general enough for all tx types to use (?).

Also, since `output guard` value is verified in the step `startStandardExit`. To save some gas, we save only the hash of all data needed in `challengeStandardExit` step and pass those data as input (check the variable `challengeDataHash`). So we don't need to verify those data again in other steps as well as not need to save multiple data in storage.

#### Spending Condition (Output Predicate)
I made Output Predicate (`SpendingCondition`) bind to tx type.  (see: `IPaymentSpendingCondition`).
I am just not sure whether that is general enough or not as current interface does not pass in whole output data (eg. token and amount) to the `SpendingCondition` contract.

```
function verifyViaOutputGuard(
        bytes calldata _outputGuardData,
        uint256 _utxoPos,
        bytes32 _outputId,
        bytes calldata _consumeTx,
        uint8 _inputIndex,
        bytes calldata _witness
    ) external returns (bool);
```

- In this one, `utxoPos` and `outputId` serves same purpose, where checks with `consumeTx.inputs[inputIndex]`. Only one of them would be used actually. This is how we can support both `utxoPos` and `outputId` to be as input format.
- `outputGuardData`  provide us data of authentication owner, eg. venue address or trader address depends on the need. The SpendingCondition contract would knows how to parse the outputGuardData. (Quite likely reuse the `outputGuardParser`)
- `witness` holds **all** proof data to be used. eg. output owner signature, confirm signature, zk-snark proof data, zk-snark proof public inputs. (is this a good idea?)

The another function `verifyViaOwner` is just a simplified version of the previous one when output guard is not needed.

For DEX tx consuming Payment, it would be using output guard.

For transition to ERC721 tx, I think they can just use output type 0, where owner info is directly in output guard field. Would be using different Spending Condition because the consume tx type of ERC721 tx is different from Payment tx.

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
+++ b/plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
@@ -2,9 +2,16 @@ pragma solidity ^0.5.0;
 pragma experimental ABIEncoderV2;
 
 import "./PaymentExitDataModel.sol";
+import "./spendingConditions/IPaymentSpendingCondition.sol";
+import "./spendingConditions/PaymentSpendingConditionRegistry.sol";
+import "../IOutputGuardParser.sol";
+import "../OutputGuardParserRegistry.sol";
 import "../utils/ExitId.sol";
 import "../utils/ExitableTimestamp.sol";
+import "../utils/OutputId.sol";
+import "../utils/OutputGuard.sol";
 import "../../framework/interfaces/IPlasmaFramework.sol";
+import "../../framework/interfaces/IExitProcessor.sol";
 import "../../framework/models/ExitModel.sol";
 import "../../utils/IsDeposit.sol";
 import "../../utils/OnlyWithValue.sol";
@@ -14,7 +21,12 @@ import "../../utils/Merkle.sol";
 import "../../transactions/PaymentTransactionModel.sol";
 import "../../transactions/outputs/PaymentOutputModel.sol";
 
-contract PaymentStandardExitable is OnlyWithValue {
+contract PaymentStandardExitable is
+    IExitProcessor,
+    OnlyWithValue,
+    OutputGuardParserRegistry,
+    PaymentSpendingConditionRegistry
+{
     using PaymentOutputModel for PaymentOutputModel.Output;
     using UtxoPosLib for UtxoPosLib.UtxoPos;
     using TxPosLib for TxPosLib.TxPos;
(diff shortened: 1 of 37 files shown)


plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
```diff
@@ -14,7 +18,11 @@ import "../../utils/Merkle.sol";
 import "../../transactions/PaymentTransactionModel.sol";
 import "../../transactions/outputs/PaymentOutputModel.sol";
 
-contract PaymentStandardExitable is OnlyWithValue {
+contract PaymentStandardExitable is
```
kevsul: Should this contract also be a `IExitProcessor`?
In `enqueueStandardExit()` it assigns itself to be the exitProcessor:
```
 ExitModel.Exit memory exitDataForQueue = ExitModel.Exit({
            exitProcessor: address(this),
```

plasma_framework/contracts/src/exits/OutputGuardParserRegistrable.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.0;
+
+import "./IOutputGuardParser.sol";
+import "../framework/utils/Operated.sol";
+
+contract OutputGuardParserRegistrable is Operated {
```
kevsul: Minor naming question: why `Registrable`? Isn't it a `Registry`?

plasma_framework/contracts/src/exits/payment/IPaymentSpendingCondition.sol
```diff
@@ -0,0 +1,46 @@
...
+     * @notice The function that checks output spending condition via authenticate owner.
+     * @param _authenticateOwner The authenticate user address.
+     * @param _utxoPos (optional) serves as the identifier of output. One of utxoPos and outputId must be with value.
+     * @param _outputId (optional) serves as the identifier of output. One of utxoPos and outputId must be with value.
+     * @param _consumeTx The transaction that consumes the output.
+     * @param _inputIndex The input index of the consume transaction that points to the output.
+     * @param _witness Witness data that would be able to prove that output is able to be consumed.
+     */
+    function verifyViaOwner(
+        address payable _authenticateOwner,
```
kevsul: Can we have just one `verify()` method and if necessary pass the address as bytes?
Then the SpendingCondition (or its GuardParser) will know to convert it to an address.

plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
boolafish: ah...good point. Would add that.

plasma_framework/contracts/src/exits/OutputGuardParserRegistrable.sol
boolafish: At one point I thought people might prefer the pattern of `XXX is (adj)` instead of `XXX is (noun)` for mixing that why I changed my preference. But it seems like it is leading to confusion so I will change it back.

boolafish:
After taking @kevsul 's advice, we can do the following interface:
```
    function verify(
        bytes calldata _outputOwnershipData,
        uint256 _utxoPos,
        bytes32 _outputId,
        bytes calldata _consumeTx,
        uint8 _inputIndex,
        bytes calldata _witness
    ) external view returns (bool);
```

Here are some alternatives I can think of
1.
```
    function verify(
        bytes32 _outputGuard, // this would be the same data as outputGuard field in tx
        uint256 _utxoPos,
        bytes32 _outputId,
        bytes calldata _consumeTx,
        uint8 _inputIndex,
        bytes calldata _witness // output guard pre image would be here
    ) external view returns (bool);
```
We can do this since `output guard pre image` can and quite likely would be part of witness of the consuming transaction. I actually prefer this one then the first one as this would be more consistent with the transaction.

2. 
```
    function verify(
        bytes _output, // encoded output data
        uint256 _utxoPos,
        bytes32 _outputId,
        bytes calldata _consumeTx,
        uint8 _inputIndex,
        bytes calldata _witness
    ) external view returns (bool);
```
If we need data more then just `outputGuard` (eg. token, amount) then this is probably what we need.

One critical question though:
Does DEX need the information of `token` and `amount` ? @paulperegud

boolafish:
would reopen using branch inside org instead of fork. closing this.

PR #156, opened 2019-07-09
PlasmaFramework should not be ExitGameRegistry

kevsul:
PlasmaFramework should not be ExitGameRegistry because it's an ExitGameController, which is an ExitGameRegistry.

### DIFF
--- a/plasma_framework/contracts/src/framework/PlasmaFramework.sol
+++ b/plasma_framework/contracts/src/framework/PlasmaFramework.sol
@@ -4,11 +4,10 @@ pragma experimental ABIEncoderV2;
 import "./BlockController.sol";
 import "./ExitGameController.sol";
 import "./registries/VaultRegistry.sol";
-import "./registries/ExitGameRegistry.sol";
 import "./utils/Operated.sol";
 import "./interfaces/IPlasmaFramework.sol";
 
-contract PlasmaFramework is IPlasmaFramework, Operated, VaultRegistry, ExitGameRegistry, ExitGameController, BlockController {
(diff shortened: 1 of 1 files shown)


boolafish:
This is an interesting one. In fact we only need the following If we want to minimize things here.
```
PlasmaFramework is IPlasmaFramework, ExitGameController, BlockController
```

I chose to duplicate the inheritance while writing this to explicit stating what PlasmaFramework is using. However, I don't really know what should be a general best practice on this.

kevsul:
Ah, that is interesting. I guess it's my C++ background, but seeing that duplicated inheritance gives me the creeps 😱

I know Solidity lets you do it but it just seems weird to me, especially as it's not necessary because `ExitGameRegistry` is already there in the hierarchy under `ExitGameController`

I do get what you mean about it being more explicit though.
But then where do we stop - should we explicitly list all levels of sub-contracts? Could get messy 😃

boolafish:
> But then where do we stop - should we explicitly list all levels of sub-contracts? 

Personally I am trying to explicit whatever dependency that it is going to use its public function. Well....so far it seems to be everything. (In POC there are some pure modifier contract then those would probably be something I would not explicit about)

But back to this PR, as I really don't know what is better haha I don't mind either way. But probably we'd like to remove other inheritance declaration here as I mentioned above.

kevsul:
Originally I was afraid that the storage would be duplicated i.e once for `PlasmaFramework. ExitGameRegistry` and again for `PlasmaFramework. ExitGameController.ExitGameRegistry`

But it seems that Solidity is smart enough not to do that 😅 

So it still seems weird to me, but I can get used to it...

Let's leave it as it is for now, and if PlasmaFramework ends up having 20+ base contracts maybe we can come back to it.

Issue #157, opened 2019-07-12
[PaymentExitGame] challenge standard exit


boolafish:
as title

PR #158, opened 2019-07-12
157 challenge standard exit

boolafish:
### Note
- previous draft PR (some discussion was there): https://github.com/omisego/plasma-contracts/pull/155
- Implement challenge standard exit with spending condition abstraction
- Add `OutputGuardParserRegistry` and `PaymentSpendingConditionRegistry`. These two register directly on Exit Game. Different from previous registry, they should to be "frozen" before the Exit Game got registered on Plasma. Otherwise fraud operator can register new path to consume afterward and user cannot audit that.
- Port EIP712 struct, make some small changes (add `txType` and make `owner` bytes32 instead of address)
- Add `MAX_INPUT_NUM` and `MAX_OUTPUT_NUM` for transaction. (Due to EIP712 struct needs to pad empty value, we need to know the max value)
- Utils contracts: OutputId, OutputGuard, Freezable

closes: #157 

### Test
truffle test
```
 Contract: Freezable
    onlyNonFrozen
      ✓ should reject the call when frozen (125ms)
      ✓ should pass when the call is not frozen (54ms)

Contract: PaymentEip712Lib
    hashTx
      ✓ should hash normal transaction correctly (75ms)
      ✓ should hash deposit transaction correctly (71ms)
      ✓ should hash transaction correctly given transaction has empty metaData (73ms)

Contract: OutputGuardParserRegistry
    outputGuardParsers
      ✓ should receive empty outputGuardParser contract address when output type not registered
    registerOutputGuardParser
      ✓ should be able to register successfully (45ms)
      ✓ should reject when not registered by operator
      ✓ should reject when frozen (64ms)
      ✓ should reject when trying to register with output type 0
      ✓ should reject when trying to register with an empty address
      ✓ should reject when the output type is already registered (91ms)

  Contract: OutputGuard
    build
      ✓ should get the correct output guard given output type 0 and empty output guard data
      ✓ should get the correct output guard given output type non 0 and empty output guard data
      ✓ should get the correct output guard given output type non 0 and non empty output guard data

  Contract: OutputId
    compute
      ✓ should get the correct output id for deposit tx output
      ✓ should return distinct output ids for deposits that differs in utxo pos (40ms)
      ✓ should get the correct output id for non deposit tx output when output index is 0
      ✓ should get the correct output id for non deposit tx output when output index is not 0
      ✓ should get the same output id for non deposit tx output no matter what utxo pos is passed in (41ms)

Contract: PaymentExitGame - End to End Tests
    Given contracts deployed, exit game and ETH vault registered
      Given alice deposited and transfered to bob
        When alice tries to start the standard exit on the deposit tx
          Then bob can challenge the standard exit spent
            ✓ should challenge it successfully

Contract: PaymentStandardExitable
    challengeStandardExit
      ✓ should fail when exit for such exit id does not exists (45ms)
      ✓ should fail when utxo pos mismatch exit data (81ms)
      ✓ should fail when output id mismatch exit data (83ms)
      ✓ should fail when output type mismatch exit data (79ms)
      ✓ should fail when output guard mismatch exit data (86ms)
      ✓ should fail when not able to find the spending condition contract (84ms)
      ✓ should fail when spending condition contract returns false (158ms)
      ✓ should fail when spending condition contract reverts (174ms)
      ✓ should calls the Spending Condition contract with expected params given output type 0 (231ms)
      ✓ should calls the Spending Condition contract with expected params given output type non 0 (266ms)
      ✓ should deletes the exit data when successfully challenged (203ms)
      ✓ should transfer the standard exit bond to challenger when successfully challenged (188ms)
      ✓ should emit ExitChallenged event when successfully challenged (170ms)

Contract: PaymentOutputToPaymentTxCondition
    verify
      ✓ should fail when consume tx is not type of payment tx (56ms)
      ✓ should fail when consume tx does not point to the utxo pos in input (53ms)
      ✓ should fail when consume tx not correctly signed by the input owner (70ms)
      ✓ should return true when all verification passes (68ms)

  Contract: PaymentSpendingConditionRegistry
    spendingConditions
      ✓ should receive empty contract address when pair (outputType, consumeTx) not registered
    registerSpendingCondition
      ✓ should be able to register successfully (44ms)
      ✓ should reject when not registered by operator
      ✓ should reject when frozen (60ms)
      ✓ should reject when trying to register with consume tx type 0
      ✓ should reject when trying to register with an empty address
      ✓ should reject when the pair of (output type, consume tx type) is already registered (93ms)
```


plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol
```diff
@@ -0,0 +1,120 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "../PaymentTransactionModel.sol";
+import "../outputs/PaymentOutputModel.sol";
+import "../../utils/UtxoPosLib.sol";
+
+library PaymentEip712Lib {
```
boolafish: @pnowosie do you mind to help me quickly do a sanity check on the implementation here?
I am porting the `Eip712StructHash`. Some changes:
1. add `txType` to Transaction
1. In Output, owner was `address` --> `bytes32` to be compatible with `outputGuard` field.

plasma_framework/test/helpers/paymentEip712.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const inputs = tx.inputs.map(hashInput);
+    const outputs = tx.outputs.map(hashOutput);
+
+    const txHash = web3.utils.sha3(web3.eth.abi.encodeParameters([
+        'bytes32', 'uint256', 'bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes32', 'bytes32',
+    ], [
+        TX_TYPE_HASH,
+        tx.transactionType,
+        inputs.length > 0 ? inputs[0] : hashInput(0),
```
pnowosie: You can save on computation doing this `hashInput(0)` & `hashOutput(NULL_OUTPUT)` once before this. 
Also these can be precomputed in constants on the begining, so you wouldn't need `NULL_OUTPUT`

plasma_framework/test/helpers/paymentEip712.js
```diff
@@ -0,0 +1,75 @@
...
+        inputs.length > 2 ? inputs[2] : hashInput(0),
+        inputs.length > 3 ? inputs[3] : hashInput(0),
+        outputs.length > 0 ? outputs[0] : hashOutput(NULL_OUTPUT),
+        outputs.length > 1 ? outputs[1] : hashOutput(NULL_OUTPUT),
+        outputs.length > 2 ? outputs[2] : hashOutput(NULL_OUTPUT),
+        outputs.length > 3 ? outputs[3] : hashOutput(NULL_OUTPUT),
+        tx.metaData || EMPTY_BYTES32,
+    ]));
+
+    return web3.utils.soliditySha3(
```
pnowosie: Not sure why you're using `soliditySha3` here, but `sha3` & `encodeParameters` elsewhere.
Consider to keep it consistent

plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol
pnowosie: :heavy_check_mark:  LGTM. 

Just one optimization hint: we can save on computing empty in/outputs hashes precomputing them in `Constants`. Also we already guarantee that in/outputs don't intersperse with empty one - so loops in _txHash could be optimized as well.

It is You to decide going for :point_up: or as further improvement.

plasma_framework/test/src/exits/OutputGuardParserRegistry.test.js
```diff
@@ -0,0 +1,66 @@
...
+        });
+
+        it('should reject when not registered by operator', async () => {
+            await expectRevert(
+                this.registry.registerOutputGuardParser(1, this.dummyOutputGuardParser.address, { from: other }),
+                'Not being called by operator',
+            );
+        });
+
+        it('should reject when frozen', async () => {
```
pgebal: I think currently anyone can freeze. Don't we want to limit that?

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -124,21 +162,47 @@ contract('PaymentExitGame - End to End Tests', ([_, alice, bob]) => {
...
+                        const { logs } = await this.exitGame.challengeStandardExit(input);
+                        this.challengeTxLogs = logs;
+                    });
+
+                    it('should challenge it successfully', async () => {
+                        expectEvent.inLogs(
+                            this.challengeTxLogs,
+                            'ExitChallenged',
+                            { utxoPos: new BN(this.depositUtxoPos) },
+                        );
```
pgebal: Let's also check that Bob gets the bond.

plasma_framework/test/src/exits/payment/PaymentStandardExitable.test.js
```diff
@@ -80,53 +92,94 @@ contract('PaymentStandardExitable', ([_, alice, bob]) => {
...
-                    data.utxoPos, data.tx, data.merkleProof,
+                    data.utxoPos, data.tx, OUTPUT_TYPE_ZERO, EMPTY_BYTES, data.merkleProof,
                     { from: nonOutputOwner, value: STANDARD_EXIT_BOND },
                 ),
-                'Only output owner can start an exit',
+                'Only exit target can start an exit',
+            );
+        });
+
+        it('should fail when output guard mismatch pre-image data given output type non 0', async () => {
```
pgebal: mismatch -> mismatches

plasma_framework/test/src/exits/payment/PaymentStandardExitable.test.js
```diff
@@ -204,5 +266,312 @@ contract('PaymentStandardExitable', ([_, alice, bob]) => {
...
+            await this.exitGame.setExit(input.exitId, getTestExitData(outputRelatedDataHash, alice));
+
+            const revertMessage = await conditionRevert.revertMessage();
+            await expectRevert(
+                this.exitGame.challengeStandardExit(input),
+                revertMessage,
+            );
+        });
+
+        it('should calls the Spending Condition contract with expected params given output type 0', async () => {
```
pgebal: calls -> call

plasma_framework/test/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.test.js
```diff
@@ -0,0 +1,117 @@
...
+        });
+    });
+
+    beforeEach('setup contracts', async () => {
+        this.dummyFramework = constants.ZERO_ADDRESS;
+        this.condition = await PaymentOutputToPaymentTxCondition.new(this.dummyFramework);
+    });
+
+    describe('verify', () => {
+        it('should fail when consume tx is not type of payment tx', async () => {
```
pgebal: consume tx type does not match payment tx

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
```diff
@@ -0,0 +1,49 @@
...
+        uint8 _inputIndex,
+        bytes calldata _signature
+    )
+        external
+        view
+        returns (bool)
+    {
+        PaymentTransactionModel.Transaction memory consumeTx = PaymentTransactionModel.decode(_consumeTx);
+        require(consumeTx.txType == PAYMENT_TX_TYPE, "The consume tx is not of payment tx type");
+        require(consumeTx.inputs[_inputIndex] == bytes32(_utxoPos), "The consume tx does not consume the output with such utxo pos");
```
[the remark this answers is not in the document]
boolafish: This is more or less not fixed yet. see slack discussion with @pdobacz : [here](https://omisego.slack.com/archives/CKF1ZSUMB/p1562918077190000)

In Piotr's first round investigation he preferred `utxo_pos` instead of output_id for payment tx. (However, he is on the fence now haha.) Because of that I move forward to use `utxo_pos` for Payment and supporting `output_id` for DEX settlement. However, I do hope we can end up just use one for simplicity though.

plasma_framework/test/helpers/paymentEip712.js
```diff
@@ -0,0 +1,75 @@
...
+const NULL_OUTPUT = { outputGuard: EMPTY_BYTES32, token: constants.ZERO_ADDRESS, amount: 0 };
+
+const EIP191_PREFIX = '0x1901';
+const EIP712_DOMAIN_HASH = web3.utils.sha3('EIP712Domain(string name,string version,address verifyingContract,bytes32 salt)');
+const TX_TYPE_HASH = web3.utils.sha3('Transaction(uint256 txType,Input input0,Input input1,Input input2,Input input3,Output output0,Output output1,Output output2,Output output3,bytes32 metadata)Input(uint256 blknum,uint256 txindex,uint256 oindex)Output(bytes32 owner,address currency,uint256 amount)');
+const INPUT_TYPE_HASH = web3.utils.sha3('Input(uint256 blknum,uint256 txindex,uint256 oindex)');
+const OUTPUT_TYPE_HASH = web3.utils.sha3('Output(bytes32 owner,address currency,uint256 amount)');
+const SALT = '0xfad5c7f626d80f9256ef01929f3beb96e058b8b4b0e3fe52d84f054c0e2a7a83';
+
+const hashInput = (input) => {
```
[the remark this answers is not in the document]
boolafish: https://github.com/airbnb/javascript#arrows--one-arg-parens

Though it is confusing (sigh) but this is the default linter expected style. Uses braces when with `{}` and don't use when it's single param and without `{}`

plasma_framework/test/helpers/paymentEip712.js
boolafish: I use `soliditySha3` for `keccak256(abi.encodePacked(...))`
And use `sha3` + `encodedParameters` for `keccak256(abi.encode(...))`

However, as you mentioned, what is the reason we use both `abi.encodePacked(...)` and `abi.encode(...)` differently in solidity in the beginning?

plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol
```diff
@@ -0,0 +1,120 @@
...
+        bytes32 OUTPUT_TYPE_HASH;
+        bytes32 DOMAIN_SEPARATOR;
+    }
+
+    function initConstants(address _verifyingContract) internal pure returns (Constants memory) {
+        bytes2 EIP191_PREFIX = "\x19\x01";
+        bytes32 EIP712_DOMAIN_HASH = keccak256(abi.encodePacked(
+            "EIP712Domain(string name,string version,address verifyingContract,bytes32 salt)"
+        ));
+        bytes32 TX_TYPE_HASH = keccak256(abi.encodePacked(
```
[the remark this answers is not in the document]
boolafish: yeah, I was pretty confused on what should we do for deposit when I was writing this.
So you're proposing 4 inputs with EMPTY value and 3 more outputs with empty value?

Personally I find our current deposit input confusing as it explicits one empty value instead of pure empty input. I think we can go either make it no input or make it 4 empty. As for output probably it does not matters much as there would be payment transaction without using all 4 outputs anyway. Unless we want to enforce all transaction need to fill in input/output with empty value?

plasma_framework/test/helpers/utils.js
```diff
@@ -3,4 +3,48 @@ async function spentOnGas(receipt) {
...
 
-module.exports.spentOnGas = spentOnGas;
+function buildOutputGuard(outputType, outputGuardData) {
+    const hashValue = web3.utils.soliditySha3(
+        { t: 'uint256', v: outputType },
+        { t: 'bytes', v: outputGuardData },
+    );
+    const rightMostBytes20 = hashValue.substring(hashValue.length - 40, hashValue.length);
+    const paddingZeros = Array(24).fill(0).join('');
+    return `0x${paddingZeros}${rightMostBytes20}`;
```
kevsul: JS has a `String.padStart()` method for easier padding:
```
    return `0x${rightMostBytes20.padStart(64, '0')}`;
```

plasma_framework/test/helpers/utils.js
```diff
@@ -3,4 +3,48 @@ async function spentOnGas(receipt) {
...
+}
+
+function addressToOutputGuard(address) {
+    const paddingZeros = Array(24).fill(0).join('');
+    return `0x${paddingZeros}${address.substring(2)}`;
```
kevsul: JS has a `String.padStart()` method for easier padding:
```
    return `0x${address.substring(2).padStart(64, '0')}`;
```

plasma_framework/test/helpers/utils.js
boolafish: today I learned! :)

plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
```diff
@@ -28,26 +40,71 @@ contract PaymentStandardExitable is OnlyWithValue {
...
+        bytes outputGuardData;
+        bytes outputTxInclusionProof;
+    }
+
     /**
-    @dev data to be passed around startStandardExit helper functions
+     * @dev data to be passed around startStandardExit helper functions
      */
     struct StartStandardExitData {
+        StartStandardExitInputData input;
```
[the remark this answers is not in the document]
boolafish: haha thanks, was using `input` as function input :p but it is definitely confusing. Would change to `args`.

plasma_framework/contracts/src/exits/payment/spendingConditions/IPaymentSpendingCondition.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.5.0;
+
+interface IPaymentSpendingCondition {
+    /**
+     * @notice The function that checks output spending condition via authenticate owner.
+     * @param _outputGuard OutputGuard of the output.
+     * @param _utxoPos (optional) serves as the identifier of output. One of utxoPos and outputId must be with value.
+     * @param _outputId (optional) serves as the identifier of output. One of utxoPos and outputId must be with value.
+     * @param _consumeTx The transaction that consumes the output.
```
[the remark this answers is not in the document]
boolafish: not really haha agree spending is more consistent.

plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
```diff
@@ -58,54 +115,100 @@ contract PaymentStandardExitable is OnlyWithValue {
...
+     */
+    function challengeStandardExit(ChallengeStandardExitInputData memory _input)
+        public
+        payable
+    {
+        ChallengeStandardExitData memory data = ChallengeStandardExitData({
+            input: _input,
+            exitData: exits[_input.exitId]
+        });
+        verifyChallengeExitExists(data);
```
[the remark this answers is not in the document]
boolafish: Not really sure about this one. The exit is not challenged yet though.

plasma_framework/contracts/src/exits/payment/spendingConditions/IPaymentSpendingCondition.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.5.0;
+
+interface IPaymentSpendingCondition {
+    /**
+     * @notice The function that checks output spending condition via authenticate owner.
```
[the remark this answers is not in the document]
boolafish: I am assuming you mean to cut down the remaining part of `via authenticate owner`.
Please let me know if I misunderstood your intention.

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentSpendingConditionRegistry.sol
```diff
@@ -0,0 +1,40 @@
...
+     * @param _outputType output type that the parser is registered with.
+     * @param _consumeTxType output type that the parser is registered with.
+     * @param _address Address of the spending condition contract.
+     */
+    function registerSpendingCondition(uint256 _outputType, uint256 _consumeTxType, address _address)
+        public
+        onlyOperator
+        onlyNonFrozen
+    {
+        require(_consumeTxType != 0, "Transaction Type would never be 0");
```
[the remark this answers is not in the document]
boolafish: I would go for pgebal 's adive of "must not" here. I would like to emphasize a bit it is by design of tx type itself would never be 0.

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentSpendingConditionRegistry.sol
```diff
@@ -0,0 +1,40 @@
+pragma solidity ^0.5.0;
+
+import "./IPaymentSpendingCondition.sol";
+import "../../../utils/Freezable.sol";
+import "../../../framework/utils/Operated.sol";
+
+contract PaymentSpendingConditionRegistry is Operated, Freezable {
+    mapping(bytes32 => IPaymentSpendingCondition) private _spendingConditions;
+
+    function spendingConditions(uint256 _outputType, uint256 _consumeTxType)
```
kevsul: There are still a few `consumeTx` instead of `spendingTx` left in this file

plasma_framework/test/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.test.js
```diff
@@ -0,0 +1,117 @@
...
+        });
+
+        it('should fail when spending tx does not point to the utxo pos in input', async () => {
+            const outputGuard = addressToOutputGuard(alice);
+            const output = new PaymentTransactionOutput(
+                1000, outputGuard, ETH,
+            );
+            const utxoPos = buildUtxoPos(100, 0, 1);
+            const wrongUtxoPosInTx = utxoPos + 1000;
+            const nonUsingOutputId = '0x';
```
pgebal: It stays the same for all tests. I would move that one scope up.

plasma_framework/test/src/exits/payment/spendingConditions/PaymentSpendingConditionRegistry.test.js
```diff
@@ -0,0 +1,84 @@
...
+const { expect } = require('chai');
+
+contract('PaymentSpendingConditionRegistry', ([_, other]) => {
+    beforeEach(async () => {
+        this.registry = await PaymentSpendingConditionRegistry.new();
+        this.dummyCondition = await DummyPaymentSpendingCondition.new();
+    });
+
+    describe('spendingConditions', () => {
+        it('should receive empty contract address when pair (outputType, spendingTx) not registered', async () => {
```
pgebal: What's the benefit of returning zero address over reverting?

plasma_framework/test/src/exits/utils/OutputId.test.js
```diff
@@ -0,0 +1,86 @@
...
+
+            const isDeposit = true;
+            const outputIndex = 0;
+            const dummyUtxoPos = 1000000000;
+
+            expect(await this.contract.compute(isDeposit, dummyTxBytes, outputIndex, dummyUtxoPos))
+                .to.equal(computeDepositOutputId(dummyTxBytes, outputIndex, dummyUtxoPos));
+        });
+
+        it('should return distinct output ids for deposits that differs in utxo pos', async () => {
```
pgebal: differs -> differ

plasma_framework/contracts/src/exits/payment/spendingConditions/IPaymentSpendingCondition.sol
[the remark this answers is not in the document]
pgebal: I would remove 'The function that' part as well. We know that's a function.

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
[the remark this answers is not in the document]
pgebal: Ok.

plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol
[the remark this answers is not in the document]
pgebal: In current design we enforce transaction to conform to the EIP format presented here, that is each transaction has 4 inputs and 4 outputs (some of them may be fillers - all 0s). 

I think the good way for deposits it to make it no input, but if we make it no input then they do not conform to EIP format. That might be confusing for users. Solution would be to have a different EIP format for deposits or use dynamic arrays in EIP (I don't know how to do it, I was trying to come up with something for my IFE, but gave up and saved it for later).

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentSpendingConditionRegistry.sol
```diff
@@ -0,0 +1,40 @@
...
+        public
+        view
+        returns (IPaymentSpendingCondition)
+    {
+        bytes32 key = keccak256(abi.encodePacked(_outputType, _consumeTxType));
```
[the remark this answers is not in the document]
boolafish: hmmm, we could but there is almost no benefit doing so. I tested the gas cost in remix with the following:
```
   function testKey(uint128 _a, uint128 _b) public pure returns (uint) {
        return (uint256(_a) << 128) | _b;
    }
    
    function testKey2(uint256 _a, uint256 _b) public pure returns (bytes32) {
        return keccak256(abi.encodePacked(_a, _b));
    }
```

it ends up `testKey` takes 504 execution gas and `testKey2` takes 493, which is surprisingly slightly lesser. (well....if we pass in _a, _b as uint256 and remove the step to cast those then it is like 10 gas lesser then hashing. but in general it is neglectable.)

plasma_framework/test/src/exits/OutputGuardParserRegistry.test.js
boolafish: oh yeah, I missed that. However, I just realized openzepplin `Ownable` contract has the function of [renounceOwnership(...)](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/ownership/Ownable.sol#L54) which can do the same thing as we want.

plasma_framework/test/src/exits/payment/spendingConditions/PaymentSpendingConditionRegistry.test.js
boolafish: hmmm....it is default behavior and we don't need to write extra `require(...)` for that. Also this is consistent with any other registry with single value as key

plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol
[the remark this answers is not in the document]
boolafish: AFAIK, metamask would be a blocker to sign dynamic array (though not really EIP712 protocol issue itself): https://github.com/omisego/elixir-omg/issues/551#issuecomment-479461060

The simplest solution now is we never put empty/dummy value in transaction and keep it 4 inputs 4 outputs, and only do that right before singing (transform to EIP712 transaction format). For me, as long as it is 1:1 mapping between two tx model I think it should be fine.

I do agree it might looks weird when user needs to sign a bunch of 0s (lol). Curious have we got any feedback on this @Pongch from user using our current EIP712 solution?

plasma_framework/test/helpers/paymentEip712.js
pnowosie: Interesting question. I did some look at elixir code & EIP. 
* abi.encode(...) - encodes value in regards to ABI for value’s type
* abi.encodePacket - despite smart name just join all values together 😂

Anyway it’s good we decipher what hides behing above. Thanks ✅

PR #170, opened 2019-07-15
Elevate private fields to public

pik694:
This PR changes `private` access modifiers into `public`, where the fields did have publicly available getters. We now use getters implicitly provided by Solidity.


boolafish:
It would be nice to put the reasoning in PR description or in commit message so if people want to trace would know why.

pik694:
The motivation for these changes is that if we add `public` modifier to a state variable, the compiler automatically generates the getters which are `external` and are marked as `view`s. There is no need to explicitly write this code.

Issue #173, opened 2019-07-16
New vault must wait certain period to take effect


boolafish:
### Why
In this doc: [Upgradeable component and its security analysis](
https://omisego.atlassian.net/wiki/spaces/RES/pages/66125827/Upgradeable+components+and+its+security+analysis)

If I analyzed it correctly, new vault should wait some time to be able to take effect.
The earliest time for vault to be able to start should be one min-finalization-period

### Successful criteria
- When vault registered, record the earliest taking effect time.
- `submitDepositBlock` of `BlockController` in `PlasmaFramework` should reject if not after such timing.

boolafish:
done by #176

Issue #174, opened 2019-07-16
New deposit verifier of vault should wait certain period to take effect


boolafish:
### Why
In this doc: [Upgradeable component and its security analysis](
https://omisego.atlassian.net/wiki/spaces/RES/pages/66125827/Upgradeable+components+and+its+security+analysis)

If I analyzed it correctly, new deposit verifier should wait some time to be able to take effect. (same issue as new vault)

The earliest time for new verifier to be able to take effect should be one min-finalization-period

### Current design
It is using `setDepositVerifier` to directly and immediately swap the verifier to upgrade to new tx type.

### Possible designs
1. Register verifier with version and then when time allows, upgrade to such version.
1. Register verifier with tx type. deposit function take `tx type` as input so it would knows which to use. Before calling verifier, the deposit function would check the min taking effect time.

### Successful critieria
update both `ERC20Vault` and `ETHVault`

pnowosie:
Implementer note about mentioned possible designs.
I took the 1st option, which is 
* when next (second, ...) deposit verifier is set on the Vault, it's registered along with its effectiveness time (now + MIN_EXIT_PERIOD)
* when deposit is made after the effectiveness time new deposit verifier is swapped with currently in-force one.

Issue #172, opened 2019-07-16
New exit game must wait certain period to be able to take effect


boolafish:
### Why
In this doc: [Upgradeable component and its security analysis](
https://omisego.atlassian.net/wiki/spaces/RES/pages/66125827/Upgradeable+components+and+its+security+analysis)

If I analyzed it correctly, quite a lot of functions should check when a new exit game contract can take effect. As a result, when registering, we should calculate the earliest time such new exit game contract can work.

The earliest "able-to-exit" timestamp should be 2 * min-finalization-period.

### Successful criteria
- Record min able-to-exit timestamp of a new registered exit game
- When enqueue, it checks the minimum earliest able-to-exit timestamp
- ~When vault is called to withdraw, it would check whether it fits such earliest able-to-exit timestamp~
will be tracked in separate :issue:  #186 
- When exit game writes to the finalized output mapping, it checks earliest able-to-exit timestamp (depends on whether #171 goes first or not)

PR #176, opened 2019-07-17
feat: allow new vaults to be quarantined

kevsul:
WIP - requesting feedback as to whether this is the right approach.

Added the ability for new vaults to be quarantined for a period of time, #173 

We will want to deploy the system with some pre verified vaults (e.g. ETH and ERC20), so I've added a way to do that.

### DIFF
--- a/plasma_framework/contracts/mocks/utils/QuarantineMock.sol
+++ b/plasma_framework/contracts/mocks/utils/QuarantineMock.sol
@@ -2,32 +2,22 @@ pragma solidity ^0.5.0;
 
 import "../../src/framework/utils/Quarantine.sol";
 
-contract QuarantineRegistryMock is Quarantine {
-    mapping(uint256 => address) private _contracts;
+contract QuarantineMock {
+    using Quarantine for Quarantine.Data;
+    Quarantine.Data internal _quarantine;
 
     constructor(uint256 _period, uint256 _initialImmuneCount)
-    public
-    Quarantine(_period, _initialImmuneCount)
-    {}
-
-    function registerContract(uint256 _contractId, address _contractAddress) public {
-        _contracts[_contractId] = _contractAddress;
-        Quarantine.quarantine(_contractAddress);
+        public
+    {
+        _quarantine.quarantinePeriod = _period;
+        _quarantine.immunitiesRemaining = _initialImmuneCount;
     }
 
-    function test() public view notQuarantined(msg.sender) returns (bool) {
-        return true;
-    }
-}
-
-contract QuarantinedContractMock {
-    QuarantineRegistryMock private _registry;
-
-    constructor(address _reg) public {
-        _registry = QuarantineRegistryMock(_reg);
+    function quarantineContract(address _contractAddress) public {
+        _quarantine.quarantine(_contractAddress);
     }
 
-    function test() public view returns (bool) {
-        return _registry.test();
+    function isQuarantined(address _contractAddress) public view returns (bool) {
+        return _quarantine.isQuarantined(_contractAddress);
     }
 }
--- a/plasma_framework/contracts/mocks/utils/QuarantineMock.sol
+++ b/plasma_framework/contracts/mocks/utils/QuarantineMock.sol
@@ -1,0 +1,33 @@
+pragma solidity ^0.5.0;
+
+import "../../src/framework/utils/Quarantine.sol";
+
+contract QuarantineRegistryMock is Quarantine {
+    mapping(uint256 => address) private _contracts;
+
+    constructor(uint256 _period, uint256 _initialImmuneCount)
+    public
+    Quarantine(_period, _initialImmuneCount)
+    {}
+
+    function registerContract(uint256 _contractId, address _contractAddress) public {
+        _contracts[_contractId] = _contractAddress;
+        Quarantine.quarantine(_contractAddress);
+    }
+
+    function test() public view notQuarantined(msg.sender) returns (bool) {
+        return true;
+    }
+}
+
+contract QuarantinedContractMock {
+    QuarantineRegistryMock private _registry;
+
+    constructor(address _reg) public {
+        _registry = QuarantineRegistryMock(_reg);
+    }
+
+    function test() public view returns (bool) {
+        return _registry.test();
+    }
+}
--- a/plasma_framework/contracts/src/framework/BlockController.sol
+++ b/plasma_framework/contracts/src/framework/BlockController.sol
@@ -15,7 +15,10 @@ contract BlockController is Operated, VaultRegistry {
         uint256 blockNumber
     );
 
-    constructor(uint256 _interval, uint256 _minExitPeriod) public VaultRegistry(_minExitPeriod) {
+    constructor(uint256 _interval, uint256 _minExitPeriod, uint256 _initialImmuneVaults)
+        public
+        VaultRegistry(_minExitPeriod, _initialImmuneVaults)
+    {
         childBlockInterval = _interval;
         nextChildBlock = childBlockInterval;
         nextDepositBlock = 1;
--- a/plasma_framework/contracts/src/framework/BlockController.sol
+++ b/plasma_framework/contracts/src/framework/BlockController.sol
@@ -15,7 +15,7 @@ contract BlockController is Operated, VaultRegistry {
         uint256 blockNumber
     );
 
-    constructor(uint256 _interval) public {
+    constructor(uint256 _interval, uint256 _minExitPeriod) public VaultRegistry(_minExitPeriod) {
         childBlockInterval = _interval;
         nextChildBlock = childBlockInterval;
         nextDepositBlock = 1;
@@ -45,7 +45,7 @@ contract BlockController is Operated, VaultRegistry {
      * @dev Block number adds 1 per submission, could have at most 'childBlockInterval' deposit blocks between two child chain blocks.
      * @param _blockRoot Merkle root of the block.
      */
-    function submitDepositBlock(bytes32 _blockRoot) public onlyFromVault {
+    function submitDepositBlock(bytes32 _blockRoot) public onlyFromVault notQuarantined(msg.sender) {
         require(nextDepositBlock < childBlockInterval, "Exceeded limit of deposits per child block interval");
 
         uint256 blknum = nextChildBlock - childBlockInterval + nextDepositBlock;
--- a/plasma_framework/contracts/src/framework/BlockController.sol
+++ b/plasma_framework/contracts/src/framework/BlockController.sol
@@ -48,8 +48,9 @@ contract BlockController is Operated, VaultRegistry {
      * @dev Block number adds 1 per submission, could have at most 'childBlockInterval' deposit blocks between two child chain blocks.
      * @param _blockRoot Merkle root of the block.
      */
-    function submitDepositBlock(bytes32 _blockRoot) public onlyFromVault notQuarantined(msg.sender) {
+    function submitDepositBlock(bytes32 _blockRoot) public onlyFromVault {
         require(nextDepositBlock < childBlockInterval, "Exceeded limit of deposits per child block interval");
+        require(!_quarantine.isQuarantined(msg.sender), "Vault is quarantined.");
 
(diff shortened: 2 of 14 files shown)


plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
```diff
@@ -24,6 +30,7 @@ contract VaultRegistry is Operated {
 
         _vaults[_vaultId] = _vaultAddress;
         _vaultToId[_vaultAddress] = _vaultId;
+        quarantine(_vaultAddress);
```
boolafish: I would write `Quarantine.quarantine(...)` to explicit where this function from to make code tracing easier.

plasma_framework/contracts/src/framework/utils/Quarantine.sol
```diff
@@ -0,0 +1,44 @@
...
+     * @param _initialImmuneCount the first _initialImmuneCount contracts don't get quarantined.
+     */
+    constructor (uint256 _period, uint8 _initialImmuneCount) public {
+        _quarantinePeriod = _period;
+        _immunitiesRemaining = _initialImmuneCount;
+    }
+
+    modifier notQuarantined(address _contractAddress) {
+        require(block.timestamp > _quarantine[_contractAddress], "Contract is quarantined.");
+        delete _quarantine[_contractAddress];
```
boolafish: A general best practice is not to have things aside from `assertion` in modifier.
https://consensys.github.io/smart-contract-best-practices/recommendations/#use-modifiers-only-for-assertions

plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
```diff
@@ -1,11 +1,17 @@
 pragma solidity ^0.5.0;
 
 import "../utils/Operated.sol";
+import "../utils/Quarantine.sol";
+
+contract VaultRegistry is Operated, Quarantine {
+    uint8 public constant INITIAL_IMMUNE_VAULTS = 2;
```
boolafish: (I am not sure if my suggestion is better)
Does it worth to set this in `PlasmaFramework` instead? Just like `minExitPeriod` as different env/deployment might want to start with different num of vaults?

boolafish: btw, just see this today: https://ethereum.stackexchange.com/questions/3067/why-does-uint8-cost-more-gas-than-uint256

Surprisingly this might actually cost more (?)

plasma_framework/contracts/src/framework/utils/Quarantine.sol
kevsul: Agreed. 
Also, after the contract isn't in quarantine any more the `delete` would get repeated every time it calls any function with that modifier, which would end up costing more gas than I was trying to save...

plasma_framework/contracts/src/framework/registries/VaultRegistry.sol
kevsul: > (I am not sure if my suggestion is better)
> Does it worth to set this in `PlasmaFramework` instead? Just like `minExitPeriod` as different env/deployment might want to start with different num of vaults?

Yeah, I agree that it would be nice to have that as an option. But I suspect that PlasmaFramework will end up having a large amount of constructor options...

I'll make the change now anyway and we can re-think it later if it becomes too unwieldy

boolafish: yeah....also I am not really sure whether we need this or not. But yeah, let's just revisit if anyone complains XDD

plasma_framework/contracts/mocks/utils/QuarantineMock.sol
```diff
@@ -0,0 +1,33 @@
...
+
+contract QuarantineRegistryMock is Quarantine {
+    mapping(uint256 => address) private _contracts;
+
+    constructor(uint256 _period, uint256 _initialImmuneCount)
+    public
+    Quarantine(_period, _initialImmuneCount)
+    {}
+
+    function registerContract(uint256 _contractId, address _contractAddress) public {
```
boolafish: It took me quite a while to understand why `register` in here. And then I realize it is probably mimicking where `Quarantine` is used (VaultRegistry). I feel `Quarantine` does not seems to be designed to be used only in registry though. Personally I would just call `Quarantine.quarantine` directly in test instead of having a registry to call that.

With above mind set, I would just have `QuarantinedContractMock` and probably only need to extend the `test` function.

plasma_framework/contracts/mocks/utils/QuarantineMock.sol
```diff
@@ -0,0 +1,33 @@
...
+        _contracts[_contractId] = _contractAddress;
+        Quarantine.quarantine(_contractAddress);
+    }
+
+    function test() public view notQuarantined(msg.sender) returns (bool) {
```
boolafish: oh damn, this is so simple and smart! Why were I using event or writing states to test modifier previously.

plasma_framework/test/helpers/utils.js
```diff
@@ -31,10 +31,17 @@ function computeNormalOutputId(txBytes, outputIndex) {
     );
 }
 
+function sleep(ms) {
```
boolafish: I would probably use `advanceTime` instead of `sleep` for testing.
https://gist.github.com/ejwessel/76767933280be5cbc59816dae6742661#file-utils-js-L1

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -117,6 +119,21 @@ contract('Erc20Vault', (accounts) => {
                 'Must have only one output',
             );
         });
+
+        it('should not store a deposit in a newly registered vault', async () => {
```
boolafish: I think it is better to test the feature in `BlockController` instead of `Vault`.

 I think what we want is: `PlasmaFramework can protect user from operator registering bad vaults` 
and not: `Our new vault has the feature of not being able to depositBlock before certain time`

Also, it reduces the tests needed.

plasma_framework/test/src/vaults/EthVault.test.js
```diff
@@ -106,5 +108,18 @@ contract('EthVault', ([_, alice]) => {
                 'Must have only one output.',
             );
         });
+
+        it('should not store a deposit in a newly registered vault', async () => {
```
boolafish: same as above, better test in `BlockController` instead.

plasma_framework/contracts/src/framework/utils/Quarantine.sol
```diff
@@ -0,0 +1,43 @@
...
+        require(block.timestamp > _quarantine[_contractAddress], "Contract is quarantined.");
+        _;
+    }
+
+    /**
+     * @notice Put a contract into quarantine.
+     * @param _contractAddress the address of the contract.
+     */
+    function quarantine(address _contractAddress) public onlyOperator {
+        require(_contractAddress != address(0), "Can not quarantine an empty address");
```
boolafish: I think this two `require` is not tested

plasma_framework/contracts/src/framework/utils/Quarantine.sol
```diff
@@ -0,0 +1,43 @@
...
+contract Quarantine is Operated {
+    mapping(address => uint256) private _quarantine;
+    uint256 private _quarantinePeriod;
+    uint256 private _immunitiesRemaining;
+
+    /**
+     * @dev The _initialImmuneCount parameter allows us to deploy the platform with some
+     * pre-verified contracts that don't get quarantined.
+     * @param _period the period of time in seconds to quarantine contracts.
+     * @param _initialImmuneCount the first _initialImmuneCount contracts don't get quarantined.
```
boolafish: how about sth like `@param _initialImmuneCount num of contracts that don't get quarantined at first.` ?

kevsul:
> Just found a high level issue of this design. I would like to reuse this for #174 and #172 too. However, at least in #172, as they (VaultRegistry + ExitGameRegistry) would all be inherited to one single `PlasmaFramework`, they would end up being only one `Quarantine ` storage, which means single `_quarantinePeriod` and single `_immunitiesRemaining`. (well I think it will fail during compilation as the constructor inputs are different)
> 
> With above issue, I start to doubt is `modifier` the best approach here though. Probably it makes more sense to just make it library with struct that holds `quarantinePeriod ` and `immunitiesRemaining`?

You're right, it won't work. I had thought it would, but the way inheritance works in Solidity means that you can't have multiple instances of a base class in the hierarchy...

A library is probably the way to go instead, I'll change it

plasma_framework/test/helpers/utils.js
kevsul: Even better, OpenZepplin test-helpers `time.increase()` 😄 
https://github.com/OpenZeppelin/openzeppelin-test-helpers

plasma_framework/contracts/src/framework/utils/Quarantine.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity ^0.5.0;
+
+/**
+ * @title Provides a way to quarantine (disable) contracts for a period of time
+ * @dev The immunitiesRemaining member allows us to deploy the platform with some
+ * pre-verified contracts that don't get quarantined.
+ */
+library Quarantine {
```
boolafish: I think we should use `internal` instead of `public` for functions here. This reduce some gas to external contract and we also don't need to link anymore. (though it increases contract deployment cost)

plasma_framework/test/src/framework/BlockController.test.js
```diff
@@ -115,4 +126,17 @@ contract('BlockController', ([_, other]) => {
             );
         });
     });
+
+    describe('new vault', () => {
```
boolafish: minor: the tests in this file is structured by function name and I think it would be nice if we put this under `registerVault` so the result of test can easily serve as a document of the function here. I think structuring test in this style might help audit process. But I will leave this to you as that is more or less just my personal preference.

plasma_framework/test/src/utils/Quarantine.test.js
```diff
@@ -0,0 +1,67 @@
...
+        it('should return true from isQuarantined when the contract is quarantined', async () => {
+            expect(await this.quarantineMock.isQuarantined(dummyAddress)).to.be.true;
+        });
+
+        it('should return false from isQuarantined when the quarantine period has passed', async () => {
+            await time.increase((QUARANTINE_PERIOD + 1) * 1000);
+            expect(await this.quarantineMock.isQuarantined(dummyAddress)).to.be.false;
+        });
+
+        it('should revert when attempting to quarantine an empty address', async () => {
```
boolafish: minor: just being nitpicking, does this and the following two tests really follows the description here:
`contract is quarantined` ?

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -47,6 +50,7 @@ contract BlockController is Operated, VaultRegistry {
      */
     function submitDepositBlock(bytes32 _blockRoot) public onlyFromVault {
         require(nextDepositBlock < childBlockInterval, "Exceeded limit of deposits per child block interval");
+        require(!_quarantine.isQuarantined(msg.sender), "Vault is quarantined.");
```
boolafish: I would be really conservative on accessing vars from parent contract: it is not easy to tell where this var is from, also when multi inheritance hit in and if ExitGameRegistry has a same named var.....what would happen on the layer of PlasmaFramework (? I guess it should be fine but....who knows XD) 

I'd like to propose just change the modifier `onlyFromVault` to `onlyFromNonQuarantinedVault` and add this line in the modifier.

plasma_framework/test/src/framework/BlockController.test.js
```diff
@@ -123,7 +123,7 @@ contract('BlockController', ([_, other]) => {
         });
     });
 
-    describe('new vault', () => {
+    describe('registerVault', () => {
```
boolafish: ouch, I left my previous comment thinking this is under `VaultRegister` instead of `BlockController`. Sorry then I would just put this under `submitDepositBlock` actually.

boolafish:
I helped rebase the PR as @kevsul is in vocation.

PR #177, opened 2019-07-18
fix: use 'exitableAt' + 'txPos' as enqueue priority

### LINKED ISSUE #175: Use "exitable" timestamp instead of "utxo pos" as priority when enqueue
### Why
I somehow was in the wrong impression that priority queue is using utxo pos as the priority. Not until recently I realized (#169) it should be "exitable" timestamp.

### Successful Criteria
- `enqueue` function of `ExitGameController` probably does not need to take `priority` anymore, it can get `exitable` data from `ExitModel.Exit` and use that directly.
- SE for payment already implemented enqueue with utxoPos, should alter that.

boolafish:
### Note
- Fix the bug of using `utxo_pos` as priority. It should be the `exitableAt` timestamp.
- Use `uint64` for `exitableAt` timestamp to be able to pack it into the priority of priority queue.
- Remove `exitableAt` from `ExitModel.Exit`. Parse from priority instead to save gas. Also, move `ExitMode.Exit` to `ExitGameController` since it is not reused anymore.
- Change the function signature of `enqueue`. Remove the struct from api and pass in all data instead to simplify.
- refactor some tests in ExitGameController, see the changed test result below. Group some tests together like: `when successully enqueued`, `given the queue already has an exitable exit`, ` given multiple exits with different priorities in queue`

### Test

```
Contract: ExitPriority
    computePriority
      ✓ should be positive correlative with "exitableAt" input given same nonce
      ✓ should be positive correlative with "exitableAt" input given nonce negatively correlated
    parseExitableAt
      ✓ should be able to parse the "exitableAt" data out from priority given nonce is 0
      ✓ should be able to parse the "exitableAt" data out from priority given max nonce value
      ✓ should be able to parse the "exitableAt" data out from priority given exitable timestamp is 0
      ✓ should be able to parse the "exitableAt" data out from priority given max exitable timestamp of uint64

Contract: ExitGameController
    enqueue
      ✓ rejects when not called from exit game contract (65ms)
      ✓ rejects when such token has not been added yet (101ms)
      ✓ rejects when called from a newly registered (still quarantined) exit game (154ms)
      ✓ can enqueue with the same exitable timestamp (priority) multiple times (224ms)
      when successully enqueued
        ✓ increases `exitQueueNonce` once (53ms)
        ✓ inserts the new unique priority to the queue (58ms)
        ✓ saves the exit data to map
    processExits
      ✓ rejects when such token has not been added yet (99ms)
      ✓ rejects when the exit queue is empty (98ms)
      ✓ rejects when the "top unique priority" mismatches with the specified one (161ms)
      ✓ does not process exit that is not able to exit yet (138ms)
      given the queue already has an exitable exit
        ✓ should be able to process when the "top unique priority" is set to 0 (74ms)
        ✓ should be able to process when the "top unique priority" is set to the exact top of the queue (72ms)
        ✓ should call the "processExit" function of the exit processor when processes (104ms)
        ✓ should delete the exit data after processed (91ms)
        ✓ should stop to process when queue becomes empty (52ms)
      given multiple exits with different priorities in queue
        ✓ should process with the order of priority and delete the processed exit from queue (117ms)
        ✓ should process no more than the "maxExitsToProcess" limit (52ms)
```

closes #175


plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
```diff
@@ -257,18 +257,17 @@ contract PaymentStandardExitable is
     }
 
     function enqueueStandardExit(StartStandardExitData memory data) private {
-        uint256 exitableAt = exitableTimestampCalculator.calculate(
+        uint32 exitableAt = exitableTimestampCalculator.calculate(
```
pik694: Why do you use only 32 bits? Don't  we have space for additional 32?

plasma_framework/contracts/mocks/framework/DummyPlasmaFramework.sol
```diff
@@ -20,9 +25,13 @@ contract DummyPlasmaFramework is IPlasmaFramework {
         return 60 * 60 * 24 * 7; // 7 days
     }
 
-    function enqueue(uint192 _priority, address _token, ExitModel.Exit memory _exit) public returns (uint256) {
-        bytes32 key = keccak256(abi.encodePacked(uint256(_priority), _token));
-        testExitQueue[key] = _exit;
+    function enqueue(address _token, ExitModel.Exit memory _exit) public returns (uint256) {
```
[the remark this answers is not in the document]
boolafish: ouch, given you pointed out....I just realize it would automatically return 0 if nothing provide (sigh.....)
I will add a `return 0` anyway

plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol
boolafish: Originally was from this comment: https://github.com/omisego/plasma-contracts/blob/master/contracts/RootChain.sol#L1104

And it seems like a uint32 would be able to serve us until Sunday, February 7, 2106, which I think is long enough for now.

However, as you mentioned and I checked, the standard timestamp is using signed integer instead of unsigned then would lead to [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem). Given that, I think it's better we bump it to 64 instead of 32 though if we keep it "unsigned" we are probably quite fine.

plasma_framework/test/helpers/positions.js
```diff
@@ -1,10 +1,19 @@
 const BLOCK_OFFSET = 1000000000;
 const TX_OFFSET = 10000;
+const BLOCK_OFFSET_FOR_TX_POS = 1000000000 / 10000;
```
pik694: I do not understand this constant and if I did understand I would rather write it as 
`... = BLOCK_OFFSET / TX_OFFSET`

plasma_framework/test/helpers/positions.js
```diff
@@ -1,10 +1,19 @@
...
 
 function buildUtxoPos(blockNum, txIndex, outputIndex) {
     return blockNum * BLOCK_OFFSET + txIndex * TX_OFFSET + outputIndex;
 }
 
+function buildTxPos(blockNum, txIndex) {
+    return blockNum * BLOCK_OFFSET_FOR_TX_POS + txIndex;
```
pik694: Consider moving the `BLOCK_OFFSET_FOR_TX_POS` constant into this code ... in here it makes sense, otherwise it takes time to understand why this constant exists

plasma_framework/test/helpers/positions.js
boolafish: aw, you're totally right. This was style from copy pasting contract code as TxPos and UTXOPOS is different lib.

PR #178, opened 2019-07-18
chore: use uint16 for output index

### LINKED ISSUE #168: Increase the limit of outputIndex
In RootChain.sol, output index is using `uint8` so I uses the same size in PlasmaFramework. However, if we are re-using the code, this would limit our DEX batch size to 256 (2^8). I think this is a bit small.

```
    function getOindex(uint256 _utxoPos)
        internal
        pure
        returns (uint8)
    {
        return uint8(_utxoPos % TX_OFFSET);
    }
```

IDK, bump it to uint16 (there is no `uintxx` in between)? However, we have a `TX_OFFSET = 10000` limit here which is smaller than 2^16. Nevertheless, we can check some `MAX_OUTPUT_INDEX_VALUE` anyway. 

Some options in my mind:
1. Bump it to uint16 with some cap/max value limit lower than TX_OFFSET
1. Bump it to uint16 and bump the TX_OFFSET

Any ideas? @paulperegud @pik694 @pgebal @kevsul

boolafish:
We were using uint8 for output index which limits the tx output to
around 256. For DEX batch settlement this seems to be small. This commit
extends the field to uint16 and cap the output index to limit of
TX_OFFSET (10000).

closes #168

### DIFF
--- a/plasma_framework/contracts/mocks/utils/UtxoPosLibWrapper.sol
+++ b/plasma_framework/contracts/mocks/utils/UtxoPosLibWrapper.sol
@@ -15,7 +15,7 @@ contract UtxoPosLibWrapper {
         return UtxoPosLib.UtxoPos(_utxoPos).txIndex();
     }
 
-    function outputIndex(uint256 _utxoPos) public pure returns (uint8) {
+    function outputIndex(uint256 _utxoPos) public pure returns (uint16) {
         return UtxoPosLib.UtxoPos(_utxoPos).outputIndex();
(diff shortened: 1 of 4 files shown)


boolafish:
Nope, in `RootChain` it can pack `exitableTimestamp` and `exitId` into one `uint256`. However, now we need one more info on the call back contract (`ExitProcessor`) as a result it would not be possible to pack the info together anymore. Thus, I use a separate struct to save the data.

PR #180, opened 2019-07-18
Fix InFlightExit id miscalculation

pik694:
We stated that the least 151 significant bits in exit id are reserved to
the tx's hash, but due to a bug it was no more that 105 bits of hash
used.

Shifting 256-bit left by 151 means that the 151 most significant bits
become zeroes and only 101 bits from the initial value are used (256 - 151 = 105).

### DIFF
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -841,7 +841,7 @@ contract RootChain {
         pure
         returns (uint192)
     {
-        return uint192((uint256(keccak256(_tx)) >> 151).setBit(152));
+        return uint192((uint256(keccak256(_tx)) >> 105).setBit(152));
     }
 
     /**
--- a/contracts/RootChain.sol
+++ b/contracts/RootChain.sol
@@ -841,7 +841,7 @@ contract RootChain {
         pure
         returns (uint192)
     {
-        return uint192((uint256(keccak256(_tx)) >> 105).setBit(152));
+        return uint192((uint256(keccak256(_tx)) >> 105).setBit(151));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pik694:
@boolafish The failing test is unfortunately a symptom of some other error, though.
I will go through the code and fix it today.

pik694:
The bit layout of exit id should be as following:
```
255:160 - padding
159:152 - oindex
151     - IFE flag
150:0   - tx hash
```

pik694:
A false start again 😢

boolafish:
This is interesting. it seems like if we ever had a standard exit with output index = 1 it should be misread as in-flight exit. I thought we would have bumped into this and learned this already :disappointed:

pik694:
>  I thought we would have bumped into this and learned this already 

yeah, a nice catch. I will test the behaviour of old version.

boolafish:
I am really curious how all Plasma dog, ODP and network upgrade could possibly not hit this issue. This seems really easy to hit (output index = 1). And let me concern if we have missed anything there :(

paulperegud:
Also, please update comments about bit layout.

pik694:
> Also, please update comments about bit layout.

Actually the layout in comments did not change.

boolafish:
move discussion from slack to GH for visibility

---
boolafish  [3 days ago]
@Pong [Not Justin Sun], can you check whether the process exit of SE got the fund back. I think the transaction would not fail, but the user should not have fund back.

boolafish  [3 days ago]
@pzelazko I think I know why the shift 151 to 105 breaks the test
- https://circleci.com/gh/omisego/plasma-contracts/413
- failed test: https://github.com/omisego/plasma-contracts/blob/master/tests/contracts/root_chain/test_piggyback_in_flight_exit.py#L148

So during processExit, it treated SE as IFE, amazingly with empty value of IFE (since it tries to get from map and would result a empty IFE struct) it would just be consider as nothing piggybacked and skip everything. However, at the end, such IFE mapping slot with the standard exit id would be flagged as finalized.

Now, after shift 151 to 105 without fixing in-flight-exit flag, now the SE exit id and IFE exit id start to crash. So the next start_in_flight_exit would get the previously set finalized IFE and fail now. Previously since the id does not crash, it would just slip through.

Issue #141, opened 2019-07-01
Provide events when any new contract registered


boolafish:
### Why
Watcher would need to monitor any newly registered contracts:
1. To know there is new endpoint
1. To warn user as a bad contract can break the system

### Successful criteria
Emit event for the following
1. ExitGameRegistry
1. VaultRegistry

### Side notes
The following registry is fine without event because they should be "frozen" before ExitGame contract is registered. Watcher won't be able to know where to get the event before ExitGame registered anyway.

1. OutputGuardParserRegistry
1. PaymentSpendingConditionRegistry

PR #181, opened 2019-07-19
feat: emit event on ExitGame and Vault registering

boolafish:
### Note
Watcher would need to monitor new registered contracts as it can potentially be a byzantine event. This commit adds event to ExitGameRegistry and VaultRegistry.

closes: https://github.com/omisego/plasma-contracts/issues/141

### DIFF
--- a/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
+++ b/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
@@ -6,6 +6,11 @@ contract ExitGameRegistry is Operated {
     mapping(uint256 => address) private _exitGames;
     mapping(address => uint256) private _exitGameToTxType;
 
+    event ExitGameRegistered(
+        uint256 txType,
+        address exitGameAddress
+    );
+
     modifier onlyFromExitGame() {
         require(_exitGameToTxType[msg.sender] != 0, "Not being called by registered exit game contract");
         _;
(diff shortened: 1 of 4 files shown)


plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
```diff
@@ -6,6 +6,11 @@ contract ExitGameRegistry is Operated {
     mapping(uint256 => address) private _exitGames;
     mapping(address => uint256) private _exitGameToTxType;
 
+    event ExitGameRegistered(
+        uint256 txType,
```
pik694: I'am afraid that hardcoding one tx type per exit game would prevent us from the future upgrade as talked with @paulperegud. 

We would rather make the interface of such registration future proof -> at some point, we would like to prepare such an exit game that could serve for all tx types.

boolafish: well....I am not really sure how the "future proofed" exit game should exactly look like. And I guess what I have in my mind might be different from yours.

What I have in my mind is that even we have an interface (well...I think it would be an abstract contract actually, as we probably need some logic), each implementation of exit game would still be 1:1 mapping. It is just the core logic for plasma protocol need would be pre-written in a template/abstract contract.

However, I am very curious what is the construction in you and pepesza's mind?

boolafish:
I will merge this. 
However, let's keep the discussion of exit-game <> tx type on. If necessary, we can start a new thread as new GH issue in research too.

@pik694 @paulperegud

PR #182, opened 2019-07-22
add eth-gas-reporter to our test

boolafish:
### Note
- adds eth-gas-reporter to help us profile our gas usage.
- https://www.npmjs.com/package/eth-gas-reporter
- I do find a down side of setting this, the unit test speed become slower : (
  - without reporter: truffle test  42.05s user 3.95s system 52% cpu 1:26.89 total
  - with reporter: truffle test  72.68s user 43.38s system 60% cpu 3:11.32 total
  - thus 2~3 times slower for current tests.

[update @2019.08.06]
- apply @kevsul's comment, use env variably to turn on reporter.
- add README for general dev setup + `MOCHA_REPORTER` env var setup.
- run `npm audit fix` as it shows warning.
- add integration with CodeChecks to show "Gas Usage" in PR. (See the CI part below in this PR)
- add `CC_SECRET` env inside circle ci for plasma-contract (required by CodeChecks).

### Test
Please see the CI part of this PR, click the details of "Gas Usage".

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,8 @@ jobs:
           name: Run tests
           command: truffle test 
           working_directory: ~/repo/plasma_framework
+          environment:
+            MOCHA_REPORTER: eth-gas-reporter
 
   Truffle linter:
     executor: truffle_executor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,8 +64,13 @@ jobs:
           command: truffle test 
           working_directory: ~/repo/plasma_framework
           environment:
+            CI: true
             MOCHA_REPORTER: eth-gas-reporter
-
+      - run:
+          name: Codechecks for gas usage
+          command: npx codechecks
+          working_directory: ~/repo/plasma_framework
+          
   Truffle linter:
     executor: truffle_executor
     working_directory: ~/repo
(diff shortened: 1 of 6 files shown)


boolafish:
Question: how do people feel with this? I don't like the fact it slows down the test and honestly we don't need to profile every time. However, I do like the report and profiling feature (though I recognize some functions should be hit but the gas report is not shown :( )

We could:
1. Add on our test setup. So we would have more understanding on our gas data per change. However, it makes our local development process slightly slower as the tests run slower.
1. Don't do anything. Only use periodically when we want to profile our contract.

Ideally I'd like to only add this to our CI, however, I find it not trivial to do so. It rely on `mocha` config inside the file `truffle-config.js`, which I don't see anyway to pass in flag to disable the reporter from truffle. eg. `truffle test --reporter xxxx`. The only (really ugly) way I can think of is in CI we "write the config file on the fly". So the CI would write the `mocha` part into our `truffle-config.js`. Given it's such an ugly approach I don't like that too : (

cc. @paulperegud @kevsul @pgebal @pik694

pik694:
One other approach is to have two configs and let the CI choose the appropriate one.
It would mean maintaining two configs, though.

kevsul:
We can optionally set the reporter in CI using an environment variable. In truffle-config.js
```
    mocha: {
        reporter: process.env.MOCHA_REPORTER || '',
```

By default there's no reporter. 
But we can enable it in CI by setting the env `MOCHA_REPORTER=eth-gas-reporter`.

boolafish:
@kevsul oh that's smart and clean!

PR #185, opened 2019-07-25
Allow new exit games to be quarantined

### LINKED ISSUE #172: New exit game must wait certain period to be able to take effect

pnowosie:
Fixes #172 

## Overview

New exit game must wait certain period to be able to take effect

## Changes

Reuses the `Quarantine` library for exit games.

## Testing
* ExitGameRegistry - checks that exit game is rejected when under quarantine
* ExitGameRegistry - checks that exit game passed when quarantine is over
* ExitGameController - rejects enqueue exit on quarantined ExitGame

### DIFF
--- a/plasma_framework/contracts/mocks/framework/registries/ExitGameRegistryMock.sol
+++ b/plasma_framework/contracts/mocks/framework/registries/ExitGameRegistryMock.sol
@@ -7,7 +7,7 @@ contract ExitGameRegistryMock is ExitGameRegistry {
         ExitGameRegistry(_minExitPeriod, _initialImmuneExitGames) public {
     }
 
-    function checkOnlyFromExitGame() public onlyFromNonQuarantinedExitGame view returns (bool) {
+    function checkOnlyFromNonQuarantinedExitGame() public onlyFromNonQuarantinedExitGame view returns (bool) {
         return true;
     }
 }
--- a/plasma_framework/contracts/mocks/framework/registries/ExitGameRegistryMock.sol
+++ b/plasma_framework/contracts/mocks/framework/registries/ExitGameRegistryMock.sol
@@ -3,7 +3,11 @@ pragma solidity ^0.5.0;
 import "../../../src/framework/registries/ExitGameRegistry.sol";
 
 contract ExitGameRegistryMock is ExitGameRegistry {
-    function checkOnlyFromExitGame() public onlyFromExitGame view returns (bool) {
+    constructor (uint256 _minExitPeriod, uint256 _initialImmuneExitGames)
+        ExitGameRegistry(_minExitPeriod, _initialImmuneExitGames) public {
+    }
+
+    function checkOnlyFromExitGame() public onlyFromNonQuarantinedExitGame view returns (bool) {
         return true;
     }
 }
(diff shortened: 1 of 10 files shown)


plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
```diff
@@ -1,18 +1,29 @@
 pragma solidity ^0.5.0;
 
 import "../utils/Operated.sol";
+import "../utils/Quarantine.sol";
 
 contract ExitGameRegistry is Operated {
     mapping(uint256 => address) private _exitGames;
     mapping(address => uint256) private _exitGameToTxType;
+    using Quarantine for Quarantine.Data;
+    Quarantine.Data internal _quarantine;
```
boolafish: this can be private

plasma_framework/contracts/mocks/framework/registries/ExitGameRegistryMock.sol
```diff
@@ -3,9 +3,15 @@ pragma solidity ^0.5.0;
...
+    uint256 public constant QUARANTINE_PERIOD = 0;
+    uint256 public constant INITIAL_IMMUNE_EXIT_GAMES = 0;
+
     event OnlyFromExitGameChecked();
 
-    function checkOnlyFromExitGame() public onlyFromExitGame {
+    constructor () ExitGameRegistry(QUARANTINE_PERIOD, INITIAL_IMMUNE_EXIT_GAMES) public {
+    }
+
+    function checkOnlyFromExitGame() public onlyFromNonQuarantinedExitGame {
```
boolafish: function name should probably be changed as well: `checkOnlyFromNonQuarantinedExitGame`.

Also, the test description in that uses this should as well be updated: [code pointer](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/src/framework/registries/ExitGameRegistry.test.js#L15)
we should probably add a test in the above place too.

boolafish:
This doesn't covers one part from the issue:
`When vault is called to withdraw, it would check whether it fits such earliest able-to-exit timestamp`

It doesn't need to be in this PR, but please open another GH issue for that so we won't miss that as this PR would close the original issue.

plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
```diff
@@ -1,18 +1,29 @@
...
     event ExitGameRegistered(
         uint256 txType,
         address exitGameAddress
     );
 
-    modifier onlyFromExitGame() {
+    constructor (uint256 _minExitPeriod, uint256 _initialImmuneExitGames)
+        public
+    {
+        _quarantine.quarantinePeriod = _minExitPeriod;
```
boolafish: different from Vault, the `quarantinePeriod` should be `2 * _minExitPeriod` I believe.

PR #195, opened 2019-07-30
Deposit verifier waits certain period to take effect

### LINKED ISSUE #174: New deposit verifier of vault should wait certain period to take effect

pnowosie:
Fixes #174 

__Solution description__
Proposed by @boolafish during review. :pray: 

Vault stores both deposit verifier addresses, lower index is earlier set. In-force verifier address is returned by getEffectiveDepositVerifier function, which checks whether _maturity timestamp_ is in future.

When new verifier contract is set, 0-index is assigned to current in-force, 1-index is this new one, maturity timestamp is set accordingly.

### DIFF
--- a/plasma_framework/contracts/src/vaults/EthVault.sol
+++ b/plasma_framework/contracts/src/vaults/EthVault.sol
@@ -17,8 +17,7 @@ contract EthVault is Vault {
      * @param _depositTx RLP encoded transaction to act as the deposit.
      */
     function deposit(bytes calldata _depositTx) external payable {
-        swapDepositVerifiersIfNewerGetsEffective();
-        IEthDepositVerifier(getDepositVerifier()).verify(_depositTx, msg.value, msg.sender);
+        IEthDepositVerifier(getEffectiveDepositVerifier()).verify(_depositTx, msg.value, msg.sender);
 
         super._submitDepositBlock(_depositTx);
     }
--- a/plasma_framework/contracts/src/vaults/EthVault.sol
+++ b/plasma_framework/contracts/src/vaults/EthVault.sol
@@ -5,7 +5,7 @@ import "./verifiers/IEthDepositVerifier.sol";
 import "../framework/PlasmaFramework.sol";
 
 contract EthVault is Vault {
-    IEthDepositVerifier private _depositVerifier;
+    //IEthDepositVerifier private _depositVerifier;
 
     event EthWithdrawn(
         address payable indexed target,
@@ -15,19 +15,11 @@ contract EthVault is Vault {
     constructor(PlasmaFramework _framework) Vault(_framework) public {}
 
     /**
-     * @notice Set the deposit verifier contract. This can be only called by the operator.
-     * @param _contract address of the verifier contract.
-     */
-    function setDepositVerifier(address _contract) public onlyOperator {
-        _depositVerifier = IEthDepositVerifier(_contract);
-    }
-
-    /**
      * @notice Allows a user to submit a deposit.
      * @param _depositTx RLP encoded transaction to act as the deposit.
      */
     function deposit(bytes calldata _depositTx) external payable {
-        _depositVerifier.verify(_depositTx, msg.value, msg.sender);
+        IEthDepositVerifier(getDepositVerifier()).verify(_depositTx, msg.value, msg.sender);
 
         super._submitDepositBlock(_depositTx);
     }
--- a/plasma_framework/contracts/src/vaults/EthVault.sol
+++ b/plasma_framework/contracts/src/vaults/EthVault.sol
@@ -5,8 +5,6 @@ import "./verifiers/IEthDepositVerifier.sol";
 import "../framework/PlasmaFramework.sol";
 
 contract EthVault is Vault {
-    //IEthDepositVerifier private _depositVerifier;
-
     event EthWithdrawn(
         address payable indexed target,
         uint256 amount
@@ -19,6 +17,7 @@ contract EthVault is Vault {
      * @param _depositTx RLP encoded transaction to act as the deposit.
      */
     function deposit(bytes calldata _depositTx) external payable {
+        swapDepositVerifiersIfNewerGetsEffective();
         IEthDepositVerifier(getDepositVerifier()).verify(_depositTx, msg.value, msg.sender);
 
         super._submitDepositBlock(_depositTx);
--- a/plasma_framework/contracts/src/vaults/Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Vault.sol
@@ -35,15 +35,34 @@ contract Vault is Operated {
     }
 
     /**
-     * @notice Set the deposit verifier contract. This can be only called by the operator.
+     * @notice Sets the deposit verifier contract. This can be only called by the operator.
+     * @notice When one contract is already set next will be effective after MIN_EXIT_PERIOD.
      * @param _contract address of the verifier contract.
      */
     function setDepositVerifier(address _contract) public onlyOperator {
-        _currentDepositVerifier = _contract;
-        //_depositVerifier = IErc20DepositVerifier(_contract);
+        require(_contract != address(0), "Cannot set an empty address as deposit verifier");
+
+        if (_currentDepositVerifier != address(0)) {
+            _newDepositVerifier = _contract;
+            _newDepositVerifierEffectivePeriod = block.timestamp + framework.minExitPeriod();
+        } else {
+            _currentDepositVerifier = _contract;
+        }
     }
 
+    /**
+     * @notice Gets currently effective deposit verifier contract address.
+     * @return contract address of deposit verifier.
+     */
     function getDepositVerifier() public view returns (address) {
+        require(_currentDepositVerifier != address(0), "Deposit verifier was not set yet.");
         return _currentDepositVerifier;
+    }
+
+    function swapDepositVerifiersIfNewerGetsEffective() internal {
+        if (_newDepositVerifier != address(0) && _newDepositVerifierEffectivePeriod <= block.timestamp) {
+            _currentDepositVerifier = _newDepositVerifier;
+            _newDepositVerifier = address(0);
+        }
     }
 }
--- a/plasma_framework/contracts/src/vaults/Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Vault.sol
@@ -14,7 +14,7 @@ contract Vault is Operated {
      *  `newDepositVerifierMaturityTimestamp` point of time and second become effective after
     */
     address[2] public depositVerifiers;
-    uint256 public newDepositVerifierMaturityTimestamp = 2 ** 256 - 1; // point far in the future
+    uint256 public newDepositVerifierMaturityTimestamp = 2 ** 255; // point far in the future
 
     constructor(PlasmaFramework _framework) public {
         framework = _framework;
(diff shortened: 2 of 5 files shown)


plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -8,6 +8,10 @@ contract Vault is Operated {
     PlasmaFramework framework;
     bytes32[16] zeroHashes;
 
+    address private _currentDepositVerifier;
```
pik694: why keeping all of this as private fields?
If we keep those fields public, then the compiler will generate getters for us - less code to maintain, deploy.

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -29,4 +33,36 @@ contract Vault is Operated {
...
+            _currentDepositVerifier = _contract;
+        }
+    }
+
+    /**
+     * @notice Gets currently effective deposit verifier contract address.
+     * @return contract address of deposit verifier.
+     */
+    function getDepositVerifier() public view returns (address) {
+        require(_currentDepositVerifier != address(0), "Deposit verifier was not set yet.");
```
pik694: I think that the tx will fail nevertheless and such a situation can only occur when the contract is being set up.
Please add a test that ensures that the deposit fails when there is no verifier, but not necessarily explicitly failed by the code.

And if we omit the check it is nothing less than a simple getter without any logic inside.

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -127,6 +130,26 @@ contract('Erc20Vault', (accounts) => {
                 'Must have only one output',
             );
         });
+
+        it('deposit verifier waits a period of time before takes effect', async () => {
+            const depositValue = DEPOSIT_VALUE / 2;
+            const deposit = Testlang.deposit(depositValue, alice, this.erc20.address);
+            await this.erc20.approve(this.erc20Vault.address, depositValue, { from: alice });
+            expect(this.currentDepositVerifier).to.not.equal(null);
```
pik694: are you checking the value from the contract or from the js?

pik694: I think that you have just set it in the before code, so you do not have to explicitly check it 😉

plasma_framework/test/src/vaults/EthVault.test.js
```diff
@@ -115,6 +117,24 @@ contract('EthVault', ([_, alice]) => {
                 'Must have only one output.',
             );
         });
+
+        it('deposit verifier waits a period of time before takes effect', async () => {
```
pik694: I hardly like the idea of maintaining two sets of tests for both ERC20 tokens and Eth .... maybe we can extract a high level abstraction so there is only one test set

boolafish:
I think we don’t need a swap. We can just put a 

if(now > period) useNew else useOld.

In register, we put the new to old and update the new one with new period. 

(Sorry for not well formatted comment, done by phone)

pik694:
> I think we don’t need a swap. We can just put a
> 
> if(now > period) useNew else useOld.
> 
> In the getter.
> 
> And then in register, we put the new to old and update the new one with new period.
> 
> Then the cost of deposit would be constant.

Technically yes, but is it clean? I mean the code becomes dirtier, because you newer know which verifier will be used by just the variable name.

I think I prefer the swapping solution.

plasma_framework/contracts/src/vaults/Vault.sol
pnowosie: Does `public` means any other contract can change this immediately on the Vault contract?
I want to keep code that is allowed to modify these values close together.

boolafish:
We can wrap functions for what client would be interested instead of relying directly on the variable in. Eg. currentEffectingContract() nextIncomingContract() nextIncomingContactEffectingTime()


Well...it’s more or less do we want to be customer obsessive and add some complexity to us or not.

plasma_framework/contracts/src/vaults/EthVault.sol
```diff
@@ -5,29 +5,20 @@ import "./verifiers/IEthDepositVerifier.sol";
...
-        _depositVerifier = IEthDepositVerifier(_contract);
-    }
-
     /**
      * @notice Allows a user to submit a deposit.
      * @param _depositTx RLP encoded transaction to act as the deposit.
      */
     function deposit(bytes calldata _depositTx) external payable {
-        _depositVerifier.verify(_depositTx, msg.value, msg.sender);
+        swapDepositVerifiersIfNewerGetsEffective();
```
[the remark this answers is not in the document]
pnowosie: As a drawback of chosen approach was mentioned 2-step process, so
1. set new verifier address (annonce it will be changed)
2. swap verifiers if when the quarantine time passed
Then if one wants to deposit (with new tx type) has first check if in-force verifier version match tx.

With current proposition it puts constant cost on every deposit to check condition and first deposit after new verifier maturity time will be penalized with a cost of swap.
Is the cost of these operations significant enough to move to the 2-step process :question: 
WDYT @boolafish @paulperegud

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -29,4 +33,36 @@ contract Vault is Operated {
...
+    function setDepositVerifier(address _contract) public onlyOperator {
+        require(_contract != address(0), "Cannot set an empty address as deposit verifier");
+
+        if (_currentDepositVerifier != address(0)) {
+            _newDepositVerifier = _contract;
+            _newDepositVerifierEffectivePeriod = block.timestamp + framework.minExitPeriod();
+        } else {
+            _currentDepositVerifier = _contract;
+        }
+    }
```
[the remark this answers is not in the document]
pnowosie: Funds already deposited are safe whatever deposit verifier is about to mature.
If I registered malicious verifier and you haven't verified its code but deposited anyway after its maturity - yes, it's your fault (other upgrade-ability mechanisms works that way as well).

plasma_framework/contracts/src/vaults/EthVault.sol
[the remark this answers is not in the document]
boolafish: Yeah, I think the main issue is the cost of swap. I have a proposal in previous comment, however I don’t know which is better actually haha. (Like two step vs more dirty code) would like to hear your opinion.

plasma_framework/test/src/vaults/Erc20Vault.test.js
pnowosie: Oh this `expect` is stupid - :+1:

plasma_framework/contracts/src/vaults/Vault.sol
pik694: https://solidity.readthedocs.io/en/v0.5.10/contracts.html#getter-functions

plasma_framework/test/src/vaults/EthVault.test.js
pnowosie: I agree that both tests are very similar the thing is that `swap....` is called in the `deposit` which specific vault implement by itself. To be really sure I haven't forgotten any of them - here we're both  :open_hands:

plasma_framework/contracts/src/vaults/Vault.sol
pik694: This check only prevents the malicious operator from setting the address to zero - obviously invalid one. But it does not prevent him from setting it to any other obviously malicious ex. `address(1)`

plasma_framework/contracts/src/vaults/EthVault.sol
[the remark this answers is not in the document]
pik694: one thing is to add a small constant cost, but another thing is to force users to pay for the swap

plasma_framework/contracts/src/vaults/Vault.sol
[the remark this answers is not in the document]
pik694: I know it, but you somehow must know that the verifier is to be changed in the future.

plasma_framework/contracts/src/vaults/Vault.sol
pik694: It will not allow anyone to change the value.

pik694:
But still you do not have a single variable ex `currentVerifier`

plasma_framework/contracts/src/vaults/Vault.sol
boolafish: I think it will though? The child contract can access and rewrite the value right?

boolafish:
If we’ve provide enough api for client the currentVerifier is of no interest to anyone then we just keep it private?

boolafish:
I will leave it to you two to choose one and move forward. Personally, I prefer two steps or the abstracted apis to be customer obsessive. 

I think both two steps or abstracted apis are kind of complexity to us in a different way though. (Or there might be better option? :)

pnowosie:
I've updated solution description in the [comment](https://github.com/omisego/plasma-contracts/issues/174#issuecomment-517171089)

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -42,27 +46,26 @@ contract Vault is Operated {
     function setDepositVerifier(address _contract) public onlyOperator {
```
pik694: input parameter as `_verifier` rather than just generic `_contract`

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -5,12 +5,16 @@ import "../framework/PlasmaFramework.sol";
...
 
-    address private _currentDepositVerifier;
-    address private _newDepositVerifier;
-    uint256 private _newDepositVerifierEffectivePeriod;
+    /**
+     * @notice Stores deposit verifier contracts addresses where first was effective upto
+     *  `newDepositVerifierMaturityTimestamp` point of time and second become effective after
+    */
+    address[2] public depositVerifiers;
+    uint256 public newDepositVerifierMaturityTimestamp = 2 ** 256 - 1; // point far in the future
```
pik694: does this even work and not overflow?

pik694: I tested it and behaves as intended, but it may be causing confusions in the future.
`2 ** 255` is safe as well, but does not require thinking and testing the overflow.

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -42,27 +46,26 @@ contract Vault is Operated {
...
-    function getDepositVerifier() public view returns (address) {
-        require(_currentDepositVerifier != address(0), "Deposit verifier was not set yet.");
-        return _currentDepositVerifier;
-    }
-
-    function swapDepositVerifiersIfNewerGetsEffective() internal {
-        if (_newDepositVerifier != address(0) && _newDepositVerifierEffectivePeriod <= block.timestamp) {
-            _currentDepositVerifier = _newDepositVerifier;
-            _newDepositVerifier = address(0);
+    function getEffectiveDepositVerifier() public view returns (address) {
```
pik694: 👍

plasma_framework/contracts/src/vaults/Vault.sol
pnowosie: Yeah, I was confused by it as well :+1:

PR #200, opened 2019-08-05
Upgrade Python RLP encoding library to v1

pik694:
Upgrade the RLP encoding library to the latest version, so that we can continue with other upgrades and switching into `plasma-framework`. 

The code must have changed a little, as the API of the lib has changed.

Refers to  #28 
I took a little different approach and applied some of the comments that appeared there in a CR.

### DIFF
--- a/plasma_core/block.py
+++ b/plasma_core/block.py
@@ -1,6 +1,6 @@
 import rlp
-from rlp.sedes import binary, CountableList, big_endian_int
-from ethereum import utils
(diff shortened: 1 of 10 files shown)


boolafish:
curious, this becomes mandatory because we are using new python stack?

pik694:
The new stack deprecates `py-ethereum` and several others libs and as a result we have to upgrade RLP

Issue #197, opened 2019-07-31
Increase Quarantined period of Exit Game


boolafish:
### Why
Current implementation uses `2 * minExitPeriod`. At first I thought we can block any new priority lower than such period to be added to priority queue. Then we only need to use the quarantined  period for user to "start exit" and does not need to "process exit". If so, two weeks should be enough (same assumption of current plasma)

However, as we also introduces `flagOutputSpent` function, an invalid ExitGame can still block valid exit by flagging output. (Same applies to vault `withdraw` function) This means we actually need to finish processing exits from queue before the invalid ExitGame contract to start take effect. Then 2 weeks would not be enough because it would left no time for user to process the exit. (Because the `exitableAt` time would be mined block timestamp + 2 week, so there is almost no time for user to run `processExits`)

I would like to propose to increase to **at least** `3 * minExitPeriod`, so we have a week of time to process. Or we can even make it `4 * minExitPeriod` if we want to get extra security promise and don't make `processExits` a scale blocker (max num of utxo) of our plasma.

### Successful Criteria
- increase the quarantined period from 2 -> 3 or 4

boolafish:
For a security first approach we can probably do 4 weeks so there would be 2 weeks for users to process the exits in queue. UX wise, if the user trust our operator, it does not really matters. They can still deposit and do plasma tx and just not able to exit.

A counter opinion would be gas wise so far it seems like process exit is cheaper than start standard exit and we only promises 1 week of time on start exit so it might not worth an extra week of waiting.

gas report: https://github.com/omisego/plasma-contracts/pull/182/checks

boolafish:
Moving discussion from slack for availability

---

Gas-wise processExits seems cheaper than startSE, so 1 week does not decrease our scalability (size of #uxto) in this sense as we would be limited to startSE anyway. (edited)

Pong [want green lights]  23 hours ago
ProcessExit is already an assumed cost for operations for child chain operator. So I think this is okay.
As for the exit period, I dont think I have enough info to give inputs.. I assume for mass exit scenario- the longer the safer ? What are the tradeoffs here ?

boolafish  22 hours ago
For new feature, if users want to onboard us safely, they would need to wait extra time for the new feature (exit game) to land fully.
However, if they trust us not fraud, they can deposit and make plasma tx earlier in risk of not able to exit before the time passed. (edited)

Pong [want green lights]  22 hours ago
would need to wait extra time
what does this wait time entail exactly ?

boolafish  22 hours ago
[...]

PR #201, opened 2019-08-06
Withdraw can be called only by known, non-quarantined exit games

### LINKED ISSUE #186: Under-quarantine ExitGame cannot call Vault's withdraw, exit finalization time takes no effect on quarantine.
Moved one work-item from #172 as a separate issue.

### What
It should check the `quarantined` period of the `ExitGame`. That is the protection period for the `ExitGame` to take any effect, and ExitGame would call `withdraw` function of vaults so we need to protect the vault from new ExitGame contracts..

pnowosie:
Fixes #186 

## Changes 

Vaults uses similar to ExitGameRegisty's modifier `onlyFromNonQuarantinedExitGame`. Info about exit games quarantines are retrieved from `framework` (which extends `ExitGameRegistry`)

### DIFF
--- a/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
+++ b/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
@@ -17,7 +17,7 @@ contract ExitGameRegistry is Operated {
     constructor (uint256 _minExitPeriod, uint256 _initialImmuneExitGames)
         public
     {
-        _quarantine.quarantinePeriod = 2 * _minExitPeriod;
+        _quarantine.quarantinePeriod = 3 * _minExitPeriod;
         _quarantine.immunitiesRemaining = _initialImmuneExitGames;
     }
 
--- a/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
+++ b/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
@@ -28,11 +28,11 @@ contract ExitGameRegistry is Operated {
     }
 
     /**
-     * @notice Exposes information about exit games quarantine
+     * @dev Exposes information about exit games quarantine
      * @param _contract address of exit game contract
      * @return A boolean value denoting whether contract is safe to use, is not under quarantine
      */
-     function isSafeToUse(address _contract) public view returns (bool) {
+     function isExitGameSafeToUse(address _contract) public view returns (bool) {
          return _exitGameToTxType[_contract] != 0 && !_quarantine.isQuarantined(_contract);
      }
 
--- a/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
+++ b/plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
@@ -28,6 +28,15 @@ contract ExitGameRegistry is Operated {
     }
 
     /**
+     * @notice Exposes information about exit games quarantine
+     * @param _contract address of exit game contract
+     * @return A boolean value denoting whether contract is safe to use, is not under quarantine
+     */
+     function isSafeToUse(address _contract) public view returns (bool) {
+         return _exitGameToTxType[_contract] != 0 && !_quarantine.isQuarantined(_contract);
+     }
+
+    /**
      * @notice Register the exit game to Plasma framework. This can be only called by contract admin.
      * @param _txType tx type that the exit game want to register to.
      * @param _contract Address of the exit game contract.
--- a/plasma_framework/contracts/src/vaults/Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Vault.sol
@@ -24,7 +24,7 @@ contract Vault is Operated {
     modifier onlyFromNonQuarantinedExitGame() {
         require(
             ExitGameRegistry(framework).isSafeToUse(msg.sender),
-            "Not called from a registered Exit Game contract"
+            "Called from a nonregistered or quarantined Exit Game contract"
         );
         _;
     }
--- a/plasma_framework/contracts/src/vaults/Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Vault.sol
@@ -23,7 +23,7 @@ contract Vault is Operated {
 
     modifier onlyFromNonQuarantinedExitGame() {
         require(
-            ExitGameRegistry(framework).isSafeToUse(msg.sender),
+            ExitGameRegistry(framework).isExitGameSafeToUse(msg.sender),
             "Called from a nonregistered or quarantined Exit Game contract"
         );
         _;
--- a/plasma_framework/contracts/src/vaults/Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Vault.sol
@@ -21,9 +21,9 @@ contract Vault is Operated {
         zeroHashes = ZeroHashesProvider.getZeroHashes();
     }
 
-    modifier onlyFromExitGame() {
+    modifier onlyFromNonQuarantinedExitGame() {
         require(
-            framework.exitGameToTxType(msg.sender) != 0,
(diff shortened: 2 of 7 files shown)


plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
```diff
@@ -27,6 +27,15 @@ contract ExitGameRegistry is Operated {
         _;
     }
 
+    /**
+     * @notice Exposes information about exit games quarantine
+     * @param _contract address of exit game contract
+     * @return A boolean value denoting whether contract is safe to use, is not under quarantine
+     */
+     function isSafeToUse(address _contract) public view returns (bool) {
```
boolafish: I would name this `isExitGameSafeToUse` to avoid possible name collision with any other mixins we combined to `PlasmaFramework`.

Another minor comment is that this function might be able to use `external` instead of `public`. But I just realized all functions in this file defaults to public instead of external (sigh, my bad) so I will leave this to you. I don't think it matters that much here.

plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
```diff
@@ -27,6 +27,15 @@ contract ExitGameRegistry is Operated {
         _;
     }
 
+    /**
+     * @notice Exposes information about exit games quarantine
```
boolafish: `@notice` is used to explain to end user. If we want to use notice here we should probably explain how the whole function works, so the exit game exists part should be expressed as well IMO.

Otherwise we can use `@dev` to make this line like a comment of implementation detail for developer.

https://solidity.readthedocs.io/en/v0.5.10/natspec-format.html

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -21,10 +21,10 @@ contract Vault is Operated {
         zeroHashes = ZeroHashesProvider.getZeroHashes();
     }
 
-    modifier onlyFromExitGame() {
+    modifier onlyFromNonQuarantinedExitGame() {
         require(
-            framework.exitGameToTxType(msg.sender) != 0,
-            "Not called from a registered Exit Game contract"
+            ExitGameRegistry(framework).isSafeToUse(msg.sender),
```
boolafish: This is interesting. I believe it is not necessary to cast it to `ExitGameRegistry`. But I will leave this to you as your intention might be making this code explicit on the dependency and make code tracing easier.

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -176,6 +176,33 @@ contract('Erc20Vault', (accounts) => {
             expect(postBalance).to.be.bignumber.equal(expectedPostBalance);
         });
 
+        it('should fail when called from under quarantine vault', async () => {
```
boolafish: It should be under quarantine **Exit Game** instead of vault here.

I would break this into two tests to test one thing at once. One for this and another testing withdraw can success when quarantined period passes.

If you really want to group things together, I think we can leverage `describe` and do sth like below. 
```
describe('Given non quarantined exit game', () => {
   it ('should fail when called by the exit game contract', ...);

  describe('And then the quarantined period passes', () => {
        beforeEach(() => { pass the period.... })

       it('should run the withdraw successfully', ...);
  })
})
```

plasma_framework/test/src/vaults/EthVault.test.js
```diff
@@ -170,5 +170,29 @@ contract('EthVault', ([_, alice]) => {
                 },
             );
         });
+
+        it('should fail when called from under quarantine vault', async () => {
```
boolafish: Same as previous comment.

pnowosie:
> Do you want to as well handle this task together? #197

Absolutely!

plasma_framework/contracts/src/framework/registries/ExitGameRegistry.sol
pnowosie: I'm leaving as is until I gain my own opinion on this :) Thanks for pointing me this concern!

plasma_framework/contracts/src/vaults/Vault.sol
pnowosie: Yeah I saw possible name collision with Vaults as you suggested... but I think I'll leave it to satisfy the principle `explicit is better than implicit`

plasma_framework/test/src/vaults/Erc20Vault.test.js
pnowosie: Really nice idea - thanks!

plasma_framework/contracts/src/vaults/Vault.sol
boolafish: ah...cool, didn't realized we can use this to avoid name crash. Now I like this casting practice actually.

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -194,6 +194,42 @@ contract('Erc20Vault', (accounts) => {
         });
     });
 
+    describe('given quarantined exit game', () => {
```
boolafish: I would still put this under `withdraw`. Current test result/spec looks like this:
```
  withdraw
      ✓ should fail when not called by a registered exit game contract (40ms, 25079 gas)
      ✓ should transfer ERC token to the target (64ms, 33864 gas)
      ✓ should emit Erc20Withdrawn event correctly (45ms, 33864 gas)
  given quarantined exit game
      ✓ should fail when called under quarantine (41ms, 31294 gas)
      ✓ and then quarantine period passes (67ms, 33864 gas)
```
Which is a bit confusing on what's the target of the test without reading the implementation on the `given quarantined exit game` part

I think it would be easier to understand like this:
```
  withdraw
      ✓ should fail when not called by a registered exit game contract (40ms, 25079 gas)
      ✓ should transfer ERC token to the target (64ms, 33864 gas)
      ✓ should emit Erc20Withdrawn event correctly (45ms, 33864 gas)
      given quarantined exit game
          ✓ should fail when called under quarantine (41ms, 31294 gas)
          ✓ and then quarantine period passes (67ms, 33864 gas)
```

(PS. we can totally nest `describe`)

plasma_framework/test/src/vaults/Erc20Vault.test.js
```diff
@@ -194,6 +194,42 @@ contract('Erc20Vault', (accounts) => {
...
+        });
+
+        it('should fail when called under quarantine', async () => {
+            await expectRevert(
+                this.newExitGame.proxyErc20Withdraw(alice, this.erc20.address, this.testFundAmount),
+                'Called from a nonregistered or quarantined Exit Game contract',
+            );
+        });
+
+        it('and then quarantine period passes', async () => {
```
boolafish:
```suggestion
        it('should succeed after quarantine period passes', async () => {
```

plasma_framework/test/src/vaults/EthVault.test.js
```diff
@@ -171,4 +171,37 @@ contract('EthVault', ([_, alice]) => {
...
+        });
+
+        it('should fail when called under quarantine', async () => {
+            await expectRevert(
+                this.newExitGame.proxyEthWithdraw(alice, DEPOSIT_VALUE),
+                'Called from a nonregistered or quarantined Exit Game contract',
+            );
+        });
+
+        it('and then quarantine period passes', async () => {
```
boolafish:
```suggestion
        it('should succeed after quarantine period passes', async () => {
```

plasma_framework/test/src/vaults/Erc20Vault.test.js
pnowosie: Thx, it further simplifies test setup

Issue #198, opened 2019-07-31
Make sure we can trace exit in queue given a tx.


boolafish:
### Why
We want to make exit traceable: https://github.com/omisego/omg-block-explorer/issues/14
We would need a way for client to trace the exit given a tx.

Two things:
1. Make the heap under the priority queue public readable for client. 
2. Now a tx can derive `exitId` but not `uniquePriority` due to the fact that we are using a nonce for that. As a result, there is no mechanism for client to know which item in the priority queue is the one that the client is of interest.

### Proposed solution
emit event with `(uniquePriority, exitId)` pair in ExitGame whenever enqueued. Client can search event by `exitId` and get the priority, and then search the heap with the priority.

### Alternative solution
We go back to old design, which packs `exitId` into priority. 
```
     *   42 bits - timestamp (exitable_at); unix timestamp fits into 32 bits
     *   54 bits - blknum * 10^9 + txindex; to represent all utxo for 10 years we need only 54 bits
     *   8 bits - oindex; set to zero for in-flight tx
     *   1 bit - in-flight flag
     *   151 bit - tx hash
```
Downside of this:
   - We add one more protocol assumption which `exitId` must be unique for *all* tx types instead of *a* tx type. 
   - Since we bumped output index to uint16, we need to redesign the bits here anyway.
   - Existing `txPos` only support 10 years? Is this good enough?

PR #202, opened 2019-08-06
feat: emit ExitQueued event on enqueue

kevsul:
# Note
Make exits traceable in the priority queue by emitting an `ExitQueued` event that contains the exit's `uniquePriority`

# Tests
```
 Contract: ExitGameController
    enqueue
      ✓ emits an ExitEnqueued event
    find exit in the priority queue
      ✓ should be able to find the exit in the priority queue 
```
 
Closes: #198

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -27,6 +27,11 @@ contract ExitGameController is ExitGameRegistry {
         address token
     );
 
+    event ExitQueued(
+        uint192 indexed exitId,
+        uint256 uniquePriority
+    );
+
     constructor(uint256 _minExitPeriod, uint256 _initialImmuneExitGames)
         public
         ExitGameRegistry(_minExitPeriod, _initialImmuneExitGames)
@@ -85,6 +90,8 @@ contract ExitGameController is ExitGameRegistry {
             exitProcessor: _exitProcessor,
             exitId: _exitId
         });
+
+        emit ExitQueued(_exitId, uniquePriority);
 
         return uniquePriority;
     }
--- a/plasma_framework/test/src/framework/ExitGameController.test.js
+++ b/plasma_framework/test/src/framework/ExitGameController.test.js
@@ -144,32 +144,11 @@ contract('ExitGameController', () => {
             expect(await priorityQueue.currentSize()).to.be.bignumber.equal(new BN(2));
         });
 
-        it('emits an ExitEnqueued event', async () => {
-            const { receipt } = await this.dummyExitGame.enqueue(
-                this.dummyExit.token,
-                this.dummyExit.exitableAt,
-                this.dummyExit.txPos,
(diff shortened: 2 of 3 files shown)


plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -27,6 +27,11 @@ contract ExitGameController is ExitGameRegistry {
         address token
     );
 
+    event ExitQueued(
+        uint256 uniquePriority,
+        uint192 exitId
```
boolafish: I think this should be `indexed` to be searchable from client?

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -458,4 +479,40 @@ contract('ExitGameController', () => {
...
+            };
+            const { tx } = await this.dummyExitGame.enqueue(
+                dummyExit.token,
+                dummyExit.exitableAt,
+                dummyExit.txPos,
+                dummyExit.exitId,
+                dummyExit.exitProcessor,
+            );
+
+            // Get the exit's uniquePriority from the ExitQueued event
```
boolafish: If we want to mimic what a client would be doing, I guess we have to search with `exitId` here, this assumes given tx, we can know exitId accordingly.

Or I guess we can just get receipt and log from the previous line if we are not aiming to mimic client behavior?
```
const {receipt, log} = await this.dummyExitGame.enqueue(...);
```

I am not really sure what's your intention while designing the test so I will leave this to your decision.

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -144,6 +144,27 @@ contract('ExitGameController', () => {
             expect(await priorityQueue.currentSize()).to.be.bignumber.equal(new BN(2));
         });
 
+        it('emits an ExitEnqueued event', async () => {
```
boolafish: This can be put under the describe below: `'when successully enqueued'` so we can reuse the setup for enqueue.

ah...and just realized there's a typo there on `successully` >"< Do you mind to fix that?

plasma_framework/test/src/framework/ExitGameController.test.js
kevsul: Yes, the intention was to mimic client behaviour - I was just taking a shortcut to get the logs.
Your way is closer to what a real client would do though...

kevsul: > Otherwise I guess we can just get receipt and log from the previous line if we are not aiming to mimic client behavior?
`const {receipt, log} = await this.dummyExitGame.enqueue(...);`

BTW This doesn't work because the `log` returned here is empty.

I _think_ that this is because it only contains any logs written by the 'called' transaction (i.e. `dummyExitGame.enqueue()`) and not by any internal transactions (i.e.`ExitGameController.enqueue()`).

Not sure if this is a bug or intentional behaviour...

PR #203, opened 2019-08-07
refactor: rename BadERC20, remove GoodERC20

### LINKED ISSUE #188: Rename Good and Bad ERC20
### Why
PR comment from @pik694 
https://github.com/omisego/plasma-contracts/pull/183/files#r308000861

We can probably name those better as `Good` and `Bad` sounds naive and a bit misdirecting.

### Successful criteria
- Come up with a good naming for the two.
- Refactor all codes including tests.

kevsul:
- Rename BadERC20 to NonCompliantERC20, which is a better description of its purpose.
- Remove unnecessary GoodERC20 contract, tests can use ERC20Mintable directly.

closes #188


plasma_framework/contracts/mocks/utils/Imports.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity ^0.5.0;
```
boolafish: haha I like this `Imports.sol` file. Probably we can just put it directly under `mocks/` as it would be used as a global importing file not only for contract/tests in `utils/`? how do you think?

kevsul: Yep, makes sense 👍

Issue #191, opened 2019-07-29
Provide our team's solidity convention doc


boolafish:
For most of our code I think we follow the solidity style guide quite well: https://solidity.readthedocs.io/en/v0.5.10/style-guide.html#function-names

But we do have some style not in the guide. For instance: variable starts with underscore prefix and function with underscore prefix. We should probably either decide a style and stick to it. And it we want to keep the underscore style, we should provide when should we do this and short reasoning.

PR #209, opened 2019-08-09
refactor: controller lib to solve deploy gas limit

### LINKED ISSUE #207: PaymentExitGame hitting deployment gas limit
### What
While adding feature to `PaymentInFlightExitable`, it hit the gas limit and failed to deploy on test. Given the fact that current `PaymentExitGame` that only contains `PaymentStandardExitable` has already used `5702176` of gas, it is almost sure plus `PaymentInFlightExitable`, it would hit the limit.

### Potential Solution
- move stuff to `init()` function...failed. (I tried on current `PaymentInFlightExitable` with my piggyback implementation, still failed to deploy :( )
- make it several contracts deployment. Break the `PaymentExitGame` into multiple contracts and calls each other. The hard part would be the storage, as it need to be shared across the contracts.

boolafish:
### Note
- Use `Router - Controller` pattern to move each functionality to a single library. To do this, we pass in the `exitMap` storage to each controller library.
- Since currently IFE only has code on startIFE, so all of them moved to the `PaymentStartInFlightExit.sol`
- `PaymentStandardExitable` breaks into `PaymentStartStandardExit.sol`, `PaymentChallengeStandardExit.sol` and `PaymentProcessStandardExit.sol`. Also, the tests are break into three files as well.

closes #207


boolafish:
Nice, proven deployment decreased!

see gas report:
PaymentExitGame -2756151 gas

@pgebal this is ready to CR now.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeStandardExit.sol
```diff
@@ -0,0 +1,87 @@
...
+     * @dev data to be passed around helper functions
+     */
+    struct ChallengeStandardExitData {
+        Controller controller;
+        PaymentStandardExitRouterArgs.ChallengeStandardExitArgs args;
+        PaymentExitDataModel.StandardExit exitData;
+    }
+
+    function run(
+        Controller memory self,
```
pgebal: Sometimes we append a dash '_' to argument name and sometimes we don't. Is there a rule? I prefer not to use dashes at all.

boolafish: nah...I don't think there's a rule on this. That's why this issue: https://github.com/omisego/plasma-contracts/issues/191.

I was just following previous convention that uses `_` but in general I agree with you in that I don't really see value of those underscore.

The only time I think it has value is when we have private variable but want some getter and not using the default getter logic. If we don't underscore those var, we cannot provide getter with the style `varname()` as it is occupied by compiler (even you make it private :(

PR #210, opened 2019-08-13
Challenge in-flight exit not canonical

### LINKED ISSUE #161: [PaymentExitGame] challengeInFlightExitNotCanonical

pnowosie:
This feature is opened again in #229 as it implements adopted router-controller approach.

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
+++ b/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
@@ -368,7 +368,7 @@ contract PaymentInFlightExitable is
     function verifyFirstPhaseNotOver(PaymentExitDataModel.InFlightExit storage ife) private view {
         uint256 phasePeriod = framework.minExitPeriod() / 2;
         bool firstPhasePassed = ((block.timestamp - ife.exitStartTimestamp) / phasePeriod) >= 1;
-        require(firstPhasePassed, "Canonicity challege phase for this exit has ended");
+        require(!firstPhasePassed, "Canonicity challege phase for this exit has ended");
     }
 
     function getTokenAmountOut(PaymentTransactionModel.Transaction memory inFlightTx, address token) private pure returns (uint256) {
--- a/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
+++ b/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
@@ -28,6 +28,7 @@ contract PaymentInFlightExitable is
     using RLP for bytes;
     using RLP for RLP.RLPItem;
     using UtxoPosLib for UtxoPosLib.UtxoPos;
+    using Bits for uint256;
 
     uint8 constant public MAX_INPUT_NUM = 4;
     uint256 public constant IN_FLIGHT_EXIT_BOND = 31415926535 wei;
@@ -40,6 +41,12 @@ contract PaymentInFlightExitable is
     event InFlightExitStarted(
         address indexed initiator,
         bytes32 txHash
+    );
+
+    event InFlightExitChallenged(
+        address indexed challenger,
+        bytes32 txHash,
+        uint256 challengeTxPosition
     );
 
     /**
@@ -88,6 +95,19 @@ contract PaymentInFlightExitable is
         bytes32[] outputIds;
     }
 
+    struct ChallengeCanonicityArgs {
+        bytes inFlightTx;
+        uint8 inFlightTxInputIndex;
+        bytes competingTx;
+        uint8 competingTxInputIndex;
+        uint256 competingTxInputPos;
+        bytes32 competingTxInputOutputId;
+        uint256 competingTxInputOutputType;
+        uint256 competingTxPos;
+        bytes competingTxInclusionProof;
+        bytes competingTxWitness;
+    }
+
     constructor(PlasmaFramework _framework) public {
         framework = _framework;
         isDeposit = IsDeposit.Predicate(framework.CHILD_BLOCK_INTERVAL());
@@ -106,6 +126,73 @@ contract PaymentInFlightExitable is
         verifyStart(startExitData);
         startExit(startExitData);
         emit InFlightExitStarted(msg.sender, startExitData.inFlightTxHash);
+    }
+
+    function challengeInFlightExitNotCanonical(ChallengeCanonicityArgs memory args) public payable {
+        // Check if there is an active in-flight exit from this transaction?
+        uint192 exitId = ExitId.getInFlightExitId(args.inFlightTx);
+        PaymentExitDataModel.InFlightExit storage ife = inFlightExits[exitId];
+        require(ife.exitStartTimestamp != 0, "In-fligh exit doesn't exists");
+
+        // Check that the exit is active and in period 1
+        verifyFirstPhaseNotOver(ife);
+        // Check if exit's input was spent via MVP exit
+        verifyInputNotSpent(ife);
+
+        // Check that two transactions are not the same
+        require(
+            keccak256(args.inFlightTx) != keccak256(args.competingTx),
+            "The competitor transaction is the same as transaction in-flight"
+        );
+
+        // Check that both transactions share the same input
+        PaymentTransactionModel.Transaction memory inFlight = PaymentTransactionModel.decode(args.inFlightTx);
+        PaymentTransactionModel.Transaction memory competitor = PaymentTransactionModel.decode(args.competingTx);
+        require(
+            inFlight.inputs[args.inFlightTxInputIndex] == competitor.inputs[args.competingTxInputIndex],
+            "The competitor and transcation in-flight have to share input at given positions"
+        );
+
+        // Check that shared input owner signes competing transaction
+        require(
+            isSpendingConditionMet(
+                ife.inputs[args.inFlightTxInputIndex].outputGuard,
+                args.competingTxInputPos,
+                args.competingTxInputOutputId,
+                args.competingTxInputOutputType,
+                args.competingTx,
+                competitor.txType,
+                args.competingTxInputIndex,
+                args.competingTxWitness
+            ),
+            "Competing input spending condition is not met"
+        );
+
+        // Determine the position of the competing transaction.
+        uint256 competitorPosition = ~uint256(0);
+        if (args.competingTxPos != 0) {
+            UtxoPosLib.UtxoPos memory competingUtxoPos = UtxoPosLib.UtxoPos(args.competingTxPos);
+            (bytes32 root, ) = framework.blocks(competingUtxoPos.blockNum());
+            require(
+                Merkle.checkMembership(keccak256(args.competingTx), competingUtxoPos.txIndex(), root, args.competingTxInclusionProof),
+                "Competing transaction is not included in plasma chain"
+            );
+            competitorPosition = args.competingTxPos;
+        }
+
+        // Competitor must be first or must be older than the current oldest competitor.
+        require(
+            ife.oldestCompetitorPosition == 0 || ife.oldestCompetitorPosition > competitorPosition,
+            "Competing transaction is not older than already known competitor"
+        );
+
+        ife.oldestCompetitorPosition = competitorPosition;
+        ife.bondOwner = msg.sender;
+
+        // Set a flag so that only the inputs are exitable, unless a response is received.
+        setNonCanonicalChallenge(ife);
+
+        emit InFlightExitChallenged(msg.sender, keccak256(args.inFlightTx), competitorPosition);
     }
 
     function createStartExitData(StartExitArgs memory args) private view returns (StartExitData memory) {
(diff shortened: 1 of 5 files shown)


pnowosie:
Hey @boolafish I have a problem checking signatures (here it is [IPaymentSpendingCondition.verify](https://github.com/omisego/plasma-contracts/pull/210/files#diff-38df255e98b27b78f4cd7427ab67e66eR13))
For my `challengeNonCanonical` I need more params especialy for `verify`: [these 3 guys](https://github.com/omisego/plasma-contracts/pull/210/files#diff-9d8c77a694065c81f84f59924306da40R103-R105)

Probably I cannot escape from `outputType`, so it needs to be provided any way.
However regarding the other two input's `utxoPosition` and `outputId`. For `verify` I'd need just one of them not both, and its value has to be recoverable from `competingTx` transaction. Am I correct :question:

I'd like to require just enough parameters to `challengeNonCanonical` so maybe we can refactor `verify` to accept just single `bytes32 outputPointer`:question: Probably we could easily encode `uint256 utxoPos` to `bytes32`. What do you think?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -108,6 +128,73 @@ contract PaymentInFlightExitable is
...
+        require(
+            inFlight.inputs[args.inFlightTxInputIndex] == competitor.inputs[args.competingTxInputIndex],
+            "The competitor and transcation in-flight have to share input at given positions"
+        );
+
+        // Check that shared input owner signes competing transaction
+        require(
+            isSpendingConditionMet(
+                ife.inputs[args.inFlightTxInputIndex].outputGuard,
+                args.competingTxInputPos,
```
boolafish: This and the `outputId` should be some already checked `utxoPos` or `outputId`. In SE, both of them are checked while exit starts and then saved. Then challenge simply pass in the checked data from storage. Not sure what's best to do here though.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -310,4 +448,10 @@ contract PaymentInFlightExitable is
             ife.inputs[i] = inputTxs[i].outputs[outputIndex];
         }
     }
+
+    function setNonCanonicalChallenge(PaymentExitDataModel.InFlightExit storage ife)
+        private
+    {
+        ife.exitStartTimestamp = ife.exitStartTimestamp.setBit(255);
```
boolafish: I know this is on draft status, but an early comment here. I know this is from old design, but I found it pretty confusing to flag on timestamp var for this. 

I think we can:
1. Make `exitStartTimestamp` a `uint64` instead. And then we put a `bool` for `isCanonical` flag in the struct. They would automatically share the same slot in EVM as long as they can be packed into 32 bytes.
1. Flag it in `exitMap`. It is designed for flagging stuff already and it seems to have more space to flag. However, I personally much prefer the first solution as it is more readable.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -85,9 +92,21 @@ contract PaymentInFlightExitable is
         uint256[] inputUtxosTypes;
         bytes[] inputTxsInclusionProofs;
         bytes[] inFlightTxWitnesses;
+        // TODO: Output Ids are only computed and not used in exit
```
pnowosie: @pdobacz @boolafish: After discussed changes in `IPaymentSpendingCondition` output ids seems to be not used...
also this exit game is for Payment Tx - so for now it's using `utxoPos`-based input identifiers.

Should I live it as is - or remove?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -108,6 +127,70 @@ contract PaymentInFlightExitable is
...
+        PaymentExitDataModel.InFlightExit storage ife = inFlightExits[exitId];
+        require(ife.exitStartTimestamp != 0, "In-fligh exit doesn't exists");
+
+        // Check that the exit is active and in period 1
+        verifyFirstPhaseNotOver(ife);
+
+        // Check if exit's input was spent via MVP exit
+        // If I'm correct we no longer can check any of inputs was previously exited by Std exit
+        // RootChain sets 254-bit in the timestamp to check this
+        /// verifyInputNotSpent(ife);
```
pnowosie: @pdobacz , @boolafish - here is my another concern - right now we're unable to check condition we checked in `RootChain` contract that no output from ife was exited via MVP (standard) exit.

Are we still safe?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -108,6 +127,70 @@ contract PaymentInFlightExitable is
...
+        PaymentTransactionModel.Transaction memory inFlightTx = PaymentTransactionModel.decode(args.inFlightTx);
+
+        // Check that shared input owner signes competing transaction
+        require(
+            isSpendingConditionMet(
+                ife.inputs[args.inFlightTxInputIndex].outputGuard,
+                uint256(0), // should not be used
+                inFlightTx.inputs[args.inFlightTxInputIndex],
+                // can we trust the caller or we should extend storage for input outputTypes?
+                args.competingTxInputOutputType,
```
pnowosie: @pdobacz @boolafish - do you think it open us to a thread when we allow caller to "remind" us the input type here.
The option is to extend storage when IFE is started and we can verify the types correctness (ehh, or I believe we can)

IMHO this could be a problem when malicious operator will register "always true" predicate here and challenge everything...
Are spending condition contracts quarantined?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
```diff
@@ -108,6 +127,70 @@ contract PaymentInFlightExitable is
...
+                args.competingTx,
+                // also, shouldn't txType be hardcoded into this contract, as it servers as exit game for this particular tx type?
+                inFlightTx.txType,
```
pnowosie: @pdobacz @boolafish: also this exit game seems to be just for Payment Tx - should we hardcode `TxType` in it then?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
boolafish: oh ouch. Now I remembered why I put both `outputId` and `utxoPos` in the interface and passing in both valid value from the ExitGame contract to Spending Condition. I did this to make sure in future we don't need to break into ExitGame and change it from passing `utxoPos` to passing `outputId` so hopefully we don't need to re-audit the ExitGame contract when doing upgrade. ah, hope #214 would solve this.


Back to your question. With current interface, if you're not using it then it should be fine to not introduce that. Meanwhile, it should be separated as outputIds for inputs and outputs. We need to do a final check of all inputs' outputIds but flag both in/output outputIds as spent on `processExit`.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
boolafish: yeah, I don't think there's reason to check this if we are checking the exiting output already spent or not during `processExit`. That is whatever flagged as spent cannot exit again.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
boolafish: oh ahh....I don't think we want to "hardcode" because we'd like to reuse the exit game for PaymentTx v2, v3....
But I just noticed that if we're relying on the tx type of decoded tx, Payment v2 tx can run through our checks in a Payment v1 Exit Game. Probably we should inject the tx type from constructor then and use that?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
boolafish: I think there are potentially two kinds of attack:

1. There is new possible path on spending condition after exit game registered (and quarantined period passed). That is, after exit game registered, operator adds some new spending condition that impacts it. Though potentially we can safe guard by quarantined period, I was thinking how to just make this not possible. I was thinking like freezing the spending condition registry before the ExitGame is registered, and we only need to rely on a single ExitGame quarantined period to safe guard whole exit game. My latest thinking on this is [in here](https://github.com/omisego/plasma-contracts/issues/214#issuecomment-523722519). It seems to me this solution would be simpler and easier and safer.

2. If two spending condition somehow passes with same data input. Then you can pass-in a tx with some output type 1 and try to cheat using output type 2 spending condition predicate. I think we can solve this by each spending condition hardcoding and check the type. And a wrong spending condition would basically just break the exit game, as a result would be safe guarded by quarantined period of exit game.

plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
```diff
@@ -457,5 +457,220 @@ contract('PaymentInFlightExitable', ([_, alice, bob, carol]) => {
...
+
+                await this.exitGame.challengeInFlightExitNotCanonical(this.challengeArgs, { from: alice });
+
+                await expectRevert(
+                    this.exitGame.challengeInFlightExitNotCanonical(this.challengeArgs, { from: alice }),
+                    'Competing transaction is not older than already known competitor',
+                );
+            });
+
+            it('succeeds when competitor is not included with appropriate large position', async () => {
```
boolafish: I found it hard to understand this description. Is this testing the case the the competing transaction is inflight (as thus no inclusion proof and no position?

plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
```diff
@@ -457,5 +457,220 @@ contract('PaymentInFlightExitable', ([_, alice, bob, carol]) => {
...
+                const {
+                    args: cArgs, block, decodedCompetingTx,
+                } = createValidNoncanonicalChallengeArgs(this.argsDecoded.inFlightTx);
+
+                this.challengeArgs = cArgs;
+                this.competingTx = decodedCompetingTx;
+                this.competingTxBlock = block;
+            });
+
+            it('should successfully challenge ife', async () => {
```
boolafish: We don't have test covering the side effect of storage changes.
As a minimum we should add this here. Personally I would nest it with a `describe` and use several `it` for all side effects when challenge successfully to make our test result more explicit.

eg.
```
when successfully challenge inFlight exit
    should emit InFlightExitChallenged event
    should set the oldest competitorPosition
    should set the bond owner to challenger
    should flag the exit non canonical
```

plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
```diff
@@ -457,5 +457,220 @@ contract('PaymentInFlightExitable', ([_, alice, bob, carol]) => {
...
+            it('fails when first phase is over', async () => {
+                await time.increase((MIN_EXIT_PERIOD / 2) + 1);
+
+                await expectRevert(
+                    this.exitGame.challengeInFlightExitNotCanonical(this.challengeArgs, { from: alice }),
+                    'Canonicity challege phase for this exit has ended',
+                );
+            });
+
+            it('fails when competing tx is not included', async () => {
```
boolafish: I might be a bit more explicit since it is possible to challenge with inflight competing transaction.
```
fails when competing tx is not included in the given position
```

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
pdobacz: I think I'm unable to really reason about this without the conflated specs :(.

My other generic comment is that we should aim to separate thinking between:
1/ interfaces and code design
2/ contract security

The (2/) should just rely on the `ExitGames` - if a broken one gets introduced, everyone exits. Then those `ExitGames` should be self-contained and include all the necessary code, shrink-wrapped; no registration of SpendingConditions nor other sub-components. We still may have those abstractions, but that's (1/), so their implementations/instances are fixed from PoV of a given `ExitGame`.

The registration could be a feature of maybe some `ExitGame, v13` that will come later and will allow that. Now we're in desperate need for simplicity.

An ExitGame is bound to a tx type, and drives all "exiting actions" that that tx type can perform (currently SE & IFE, as we know them) and sets the rules for them:

The tx type (`tx type 1`) introduces various output types (`output types 1`) that are protected by spending conditions. A new tx type (`tx type 2`) introduced with a new ExitGame2 will introduce new "exiting actions" for `tx type 2`. `tx type 2` will be able to spend `output types 1` according to **their rules**, which should not restrict the spending tx type (we only would do such restriction if that output type was "special", e.g. RC deposit).

But I'm getting ahead of myself, let's discuss this when fleshing out specs. For now it's just a random plea for simplicity

boolafish: > The (2/) should just rely on the ExitGames 

I am full hand support on this! 

For `1/ interfaces and code design`, the main goal here is to avoid re-audit, and thus bringing in the registry though what we need is a predicate factory/provider (which registry by default have such feature) that we can inject/update to new one for new exit game. 

> tx type 2 will be able to spend output types 1 according to their rules, which should not restrict the spending tx type

I am supportive on throwing away spending condition registry if we know how to do update without re-audit the main code. I think such rule on output is fairly easy to be just decided by output type, but how we are going to do with input rules? We need those when IFE.

pdobacz: > avoid re-audit

What if we label that as a nice-to-have feature? The more I think about all this, the more I realize allowing for audits of just parts of the contract suite is slippery ground. The same as with the "if ExitGame not nice, then exit" part - the "not nice" is seen holistically anyway, I doubt one could decide only by looking at the new stuff.

> how we are going to do with input rules?

maybe this is the bit that I'm missing to see the problem, I think those rules should be enough to IFE securely. I'm putting this down as point to clarify when doing specs.

boolafish: > What if we label that as a nice-to-have feature? 

cc. @kasima. I totally see what you're seeing lol and I totally agree and is skeptical on "allowing for audits of just parts of the contract suite is slippery ground". But if that works it's quite a lot of saving on money on the other hands.....

> tx type 2 will be able to spend output types 1 according to their rules, which should not restrict the spending tx type

another thing I find it hard to be without considering spending tx type is that how are you going to decode the spending transaction if you don't know the type? I think you mentioned you played around on elixir side, how did you handle that?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
pnowosie: Good point :+1:

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
pdobacz: for future reference - the thread kinda continued here https://github.com/omisego/plasma-contracts/issues/214#issuecomment-524234173, a more suitable spot.

Not sure how to resolve the original question, considering the discussion at hand, yet...

boolafish: I think this PR does not need to address the interface change for Spending Condition of #214.
Back to the original problem, "do you think it open us to a thread when we allow caller to "remind" us the input type here." As for current interface, it does. We basically cannot check the correctness of output type inside spending condition predicate.

pnowosie:
Is continued in #229

plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
pnowosie: yes - I will rephrase in #229

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol
pnowosie: Created :issue: #230 to address this .

PR #211, opened 2019-08-13
[WIP] BondSize library

kevsul:
## Overview
 Implement BondSize lib, as discussed [here](https://github.com/omisego/research/issues/107)

The idea is:

1. Maintain an average gas price (exponential weighted moving average is cheap) using e.g. startExit calls.
2. Calculate bond size as: Bond size = average gas price * gas cost for challenge * safety multiplier (e.g. 1.5)

### DIFF
--- a/plasma_framework/contracts/src/exits/utils/BondSize.sol
+++ b/plasma_framework/contracts/src/exits/utils/BondSize.sol
@@ -2,28 +2,59 @@ pragma solidity ^0.5.0;
 
 import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
 
+/**
+ * @notice Calculates bond size based on current average gas prices.
+ * @dev An Exponentially Weighted Moving Average is used as it's cheap in terms of storage and resistant to sudden changes.
+ * @dev `updateGasPrice()` must be called regularly to maintain the average gas price.
+ */
 library BondSize {
     using SafeMath for uint256;
 
+    uint256 constant GWEI = 1000000000;
+
+    /**
+    * @param currentGasPrice The current average gas price.
+    * @param targetGasCost The gas cost of the target method that can claim the bond (e.g. challenge).
+    * @param ewmaFactor The 'smoothing' factor for the EWMA algorithm. Higher value means the the average changes more slowly.
+    * @param safetyFactor Adjustment factor to multiply the bond size by. Value is a percentage i.e. 100 means unchanged.
+    */
     struct Params {
         uint256 currentGasPrice;
-        uint256 challengeGasCost;
-        uint256 ewmaFactor;
-        uint256 safetyFactor;
+        uint128 targetGasCost;
+        uint64 ewmaFactor;
+        uint64 safetyFactor;
     }
 
-    function bondSize(Params storage _self) internal view returns (uint256) {
-        return _self.challengeGasCost
-            .mul(_self.currentGasPrice)
-            .mul(_self.safetyFactor)
+    /**
+    * @notice Returns the bond size based on current gas prices.
+    * @dev Rounds down to nearest gwei to avoid small changes in value.
+    * @param _self Described in `Params` above.
+    */
+    function bondSize(Params memory _self) internal pure returns (uint256) {
+        uint256 bond = _self.currentGasPrice
+            .mul(uint256(_self.targetGasCost))
+            .mul(uint256(_self.safetyFactor))
             .div(100);
+
+        return bond.div(GWEI).mul(GWEI);
     }
 
-    function updateGasPrice(Params storage _self) internal {
-        _self.currentGasPrice = ewma(_self.currentGasPrice, tx.gasprice, _self.ewmaFactor);
+    /**
+    * @notice Updates the current gas price average.
+    * @param _self Described in `Params` above.
+    * @param _newGasPrice The new gas price to be included in the average.
+    */
+    function updateGasPrice(Params storage _self, uint256 _newGasPrice) internal {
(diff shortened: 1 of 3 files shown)


plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.5.0;
+
+import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+
+library BondSize {
+    using SafeMath for uint256;
+
+    struct Params {
+        uint256 currentGasPrice;
+        uint256 challengeGasCost;
```
boolafish: The naming here can be more generalized and not take in the concept of `challenge` I think (?)

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
...
+            .mul(_self.currentGasPrice)
+            .mul(_self.safetyFactor)
+            .div(100);
+    }
+
+    function updateGasPrice(Params storage _self) internal {
+        _self.currentGasPrice = ewma(_self.currentGasPrice, tx.gasprice, _self.ewmaFactor);
+    }
+
+    function ewma(uint avg, uint value, uint factor) internal pure returns (uint) {
```
boolafish: Would this be better in `private`?

plasma_framework/test/src/utils/BondSize.test.js
```diff
@@ -0,0 +1,43 @@
...
+            .muln(this.safetyFactor)
+            .divn(100);
+    });
+
+    it('should return the initial bond size', async () => {
+        const bondSize = await this.contract.bondSize();
+        expect(bondSize).to.be.bignumber.equal(this.initialBondSize);
+    });
+
+    it('should lower the bond size when average gas price decreases', async () => {
```
boolafish: Would we not like to test with the `ewma` formula or we'd only care about increase/decrease?

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.5.0;
+
+import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+
+library BondSize {
```
boolafish: let's put document on our formula used in this library.

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    function bondSize(Params storage _self) internal view returns (uint256) {
+        return _self.challengeGasCost
+            .mul(_self.currentGasPrice)
+            .mul(_self.safetyFactor)
+            .div(100);
+    }
+
+    function updateGasPrice(Params storage _self) internal {
```
boolafish: 1. I think the following interface
```
function updateGasPrice(Params storage _self, uint256 _newGasPrice)
```
might be more easy to understand to user compare to a magic `tx.price` inside here, and would make this a more generic interface.


2. Meanwhile, though might only be personal preference, how do you feel about using the following instead of current design that save the updated state directly:
```
bond.updateGasPrice(tx.gasprice).save();
```
vs
```
bond.updateGasPrice(tx.gasprice);
```

We can do this by the following I guess:
```
function updateGasPrice(Params memory _self) internal pure returns (Params memory)
```
This makes the update function taking in `memory` + make the function pure. So if anyone ever needs this bond to do some logic and does not need to save it in storage, it would be more generic useful.

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
...
+    using SafeMath for uint256;
+
+    struct Params {
+        uint256 currentGasPrice;
+        uint256 challengeGasCost;
+        uint256 ewmaFactor;
+        uint256 safetyFactor;
+    }
+
+    function bondSize(Params storage _self) internal view returns (uint256) {
```
boolafish: I think this can be `function bondSize(Params memory _self)`. With `memory`, the function can take both `storage` and `memory` struct as input and would be more flexible.

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -0,0 +1,29 @@
...
+
+import 'openzeppelin-solidity/contracts/math/SafeMath.sol';
+
+library BondSize {
+    using SafeMath for uint256;
+
+    struct Params {
+        uint256 currentGasPrice;
+        uint256 challengeGasCost;
+        uint256 ewmaFactor;
```
boolafish: If I understand it correctly, `ewmaFactor` and `safetyFactor` would be some constant number and small. I guess we can use `uint128` for the two so they would only need one slot of storage space. Or, probably we can pack the `challengeGasCost` too, and use sth like `uint96`(?).

plasma_framework/test/src/utils/BondSize.test.js
kevsul: Hmm, I think it's better that the tests don't know the exact formula, it kind of makes them brittle

boolafish: Okay, so what we want to get from this bond library is more of an "average" concept, and we don't really care about what average algo/formula we are using. Just want to make sure I have your intention right.

boolafish:
@kevsul , I think we can close this due to our last discussion right?
https://github.com/omisego/research/issues/107#issuecomment-525267486

PR #212, opened 2019-08-14
Figure out better use spending condition

pnowosie:
Signature of `IPaymentSpendingCondition.verify` was controversial because it requires 2 arguments for parameters `utxoPos` and `outputId`.

@pdobacz help me to figure out the meaning of these two :point_up: and that they are both needed there.
So as I'm understanding the matter at the moment (assuming spending condition for Payment transaction):
* Spending condition responsibility is to check: 
   * that provided transaction spends provided input,
   * and provided guard unlocks utxo for spending (in old words: owner of utxo signs tx)
* Spending condition is registered for concrete Tx type so it knows how to decode passed transaction,
* meaning of `outputId` is an input pointer to check that provided tx indeed spends that input 
  `tx.inputs[inputIndex] == expected_input_pointer`
* meaning of `utxoPos` - not used at the moment, but could serve as point in time (or age) of the input.

:notebook: Please note that as currently implemented our **`PaymentTransaction` inputs are pointed also by utxo position - however this will change toward `outputId`**, so it makes sense to treat the parameters to `verify` aligned with theirs meaning

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
+++ b/plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
@@ -20,15 +20,16 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
     /**
      * @notice Checks if given output has been spent by owner in given spending transaction.
      * @param _outputGuard bytes that hold the address of owner directly.
-     * @param _utxoPos serves as the identifier of output.
+     * @dev _utxoPos not used, serves as the position identifier of output.
+     * @param _outputId serves as the identifier of output (input pointer).
      * @param _spendingTx The rlp encoded transaction that spends the output.
      * @param _inputIndex The input index of the spending transaction that points to the output.
      * @param _signature The signature of the output owner.
      */
     function verify(
         bytes32 _outputGuard,
-        uint256 _utxoPos,
-        bytes32 /*_outputId*/,
+        uint256, /*_utxoPos  NOTE: It's unclear how & if this will be used at all, see: https://github.com/omisego/plasma-contracts/pull/212*/
+        bytes32 _outputId,
         bytes calldata _spendingTx,
         uint8 _inputIndex,
         bytes calldata _signature
@@ -39,7 +40,7 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
     {
         PaymentTransactionModel.Transaction memory spendingTx = PaymentTransactionModel.decode(_spendingTx);
         require(spendingTx.txType == PAYMENT_TX_TYPE, "The spending tx is not of payment tx type");
-        require(spendingTx.inputs[_inputIndex] == bytes32(_utxoPos), "The spending tx does not spend the output at this utxo pos");
+        require(spendingTx.inputs[_inputIndex] == _outputId, "The spending tx does not spend the output at this utxo pos");
 
         address payable owner = AddressPayable.convert(address(uint256(_outputGuard)));
         require(owner == ECDSA.recover(eip712.hashTx(spendingTx), _signature), "Tx not correctly signed");
--- a/plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
+++ b/plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
@@ -19,9 +19,9 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
 
     /**
      * @notice Checks if given output has been spent by owner in given spending transaction.
-     * @param _outputGuard bytes that hold the address of owner directly.
      * @dev _utxoPos not used, serves as the position identifier of output.
-     * @param _outputId serves as the identifier of output (input pointer).
+     * @param _outputGuard bytes that hold the address of owner directly.
+     * @param _outputIdentifier serves as the identifier of output (spendingTx is supposed to contain it as input).
      * @param _spendingTx The rlp encoded transaction that spends the output.
      * @param _inputIndex The input index of the spending transaction that points to the output.
      * @param _signature The signature of the output owner.
@@ -29,7 +29,7 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
     function verify(
         bytes32 _outputGuard,
         uint256, /*_utxoPos  NOTE: It's unclear how & if this will be used at all, see: https://github.com/omisego/plasma-contracts/pull/212*/
-        bytes32 _outputId,
+        bytes32 _outputIdentifier,
         bytes calldata _spendingTx,
         uint8 _inputIndex,
         bytes calldata _signature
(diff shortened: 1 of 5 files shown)


pnowosie:
@boolafish and myself had an interesting discussion which is worth to keep as show our resoning

----  
@pnowosie: we eventually will move towards outputIds in Tx inputs so its more logical that spending condition is checking the right parameter

@boolafish: okay one problem I can sense now is that Payment -> DEX would be using outputId while Payment -> Payment would be using utxoPos
so in SE challenge we need both, the way you did on IFE would no apply to SE challenge

@pnowosie:  
> one problem I can sense now is that Payment -> DEX would be using outputId...

well I didn't change it really because our PaymentOutputToPaymentTxCondition really uses utxoPos but in the wrong parameter
what it really checks was outputId because it looks at Tx inputs
so if DEX will use real outputId in Tx inputs and pass approprate outputId to the spending condition it will work
I proposed to Piotr to remove utxoPos from `SpendingCondition.verify` but he pointed out it could serve as additional check in future the one which is based on utxo position in the chain
e.g. `utxoPos.blknum > 120000` or smth in `guard`
but smth related to the **position in chain** not the **generic pointer** to the utxo

@boolafish: but should that be the position of the input? not the tx that would be useful?anyway, the problem I mentioned would be here: https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/PaymentStandardExitable.sol#L326-L343

currently ExitGame does not know how the spending tx would point to PaymentTx.
but it need to pass in the correct value, either utxoPos or outputId
currently I just always pass in the correct value of both :disappointed:

if we are passing in the outputId slot, ExitGame need to know how to pass in and do a **if-else** there?

@pnowosie: ah OK what you're saying is that ExitGame will call SpendingCondition right?

aren't we will eventually replace `utxoPos` with `outputId` also in Payment txs - right?

@boolafish: ah, oh……I think I get what you mean now. hmmm, lol then I am on your side of remove `utxoPos` totally

@pnowosie: I'll keep it ("for future use") ... maybe we will remove it later at the point all things get more clear :wink:

pdobacz:
@boolafish hm, from these notes, one thing that is confusing to me:

> Payment -> DEX would be using outputId while Payment -> Payment would be using utxoPos

this doesn't sound right. Isn't it:
 - Payment -> anything uses utxoPos, because anything spends a utxoPos-pointed output
 - Dex -> anything uses output_id, because anything spends an output_id-pointed output

?

Then the SE challenge needs to check the spending condition of "anything" versus the output in question. So you need to know the input pointer that points to the exited output, check that "anything" includes it as its input and check that whatever the conditions to spend the output were met by "anything"

pnowosie:
@pdobacz (regarding your exaple) isn;t it that `Payment -> anything` transaction will have different Tx type than `DEX -> anything` so we will have 2 different `IPaymentSpendingCondition` implementations :question:

boolafish:
hmm @pdobacz I was thinking the opposite. I was thinking input of `Payment` would always be `utxoPos` while input of `DEX` would always be `outputId`. As a result, to use `DEX deposit output` from `Payment` in a `DEX` tx, it would be referred using `outputId` as `DEX` always uses `outputId`. It is not necessary to do so but I was thinking that this might be simpler (?) then `DEX` having two kinds of input types. Hmmm at least data-wise it is more consistent. But probably your proposal is more simple actually (?) 

@pnowosie yeah, I guess there would be sth like `PaymentToDexSettlementCondition`

pdobacz:
@boolafish 

OK, I'm a bit out of brain already (but it's awesome we're having this conversation!).

I think you're right that the way you mentioned is also generally possible. What it conflicts with, is the 1:1 binding. I've just got used to it too much already, so I took it for settled, possibly producing confusion on my part.

Now to explain how 1:1 binding kicks in here:

Assuming 1:1 binding is there, so the output type is what "determines" how it will be pointed to (input pointer type), when spent.

Now, assume that I'm sending a payment tx intending to place an exchange deposit. So clearly, I want this deposit to be spendable in a DEX tx. So I'd need to somehow "tell" the world that, so that the world then knows whether to reference it via `output_id` or `utxo_pos`. If I did that, I'd disclose that I'm making an exchange deposit.

Conversely, if all payment txs create outputs that are to be referenced via `utxo_pos`, both kinds of such outputs (ones that are deposits and ones that aren't) go to the same "bag".

It hard for me to give concrete evidence it is so, but somehow intuitively I think this actually is a "pro" of the 1:1 binding. When some new `tx_type2` & `ExitGame2` are introduced, we'd probably prefer that they introduce a new `input_pointer_type` that can be **created** out of `tx_type2`, rather than introduce a new way existing outputs can be pointed to. Seems like better separation of concerns between "old" and "new" tx types.

boolafish:
> What it conflicts with, is the 1:1 binding

Technically speaking, it should not conflict as `DEX deposit` is of different output type of `Payment output`. It is still an output type per way to be pointed to. And we've solved the privacy issue of `output type` disclosure using the idea of `outputGuard`. Whoever wants to spend the output would know the pre-image and would know that such output type should be pointed via `outputId` instead of `utxoPos`.

However, practically speaking, when I read your description, I guess ch-ch would like to know how it can be pointed to in advance actually, so it can update the `utxo` mapping with just a single key format (either using `utxoPos` or `outputId`) instead of updating two mappings.

If that's the case, I think the 1:1 mapping is not on output type but on tx type actually? So output of one tx type can only be pointed by one mechanism?

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
```diff
@@ -20,15 +20,16 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
     /**
      * @notice Checks if given output has been spent by owner in given spending transaction.
      * @param _outputGuard bytes that hold the address of owner directly.
-     * @param _utxoPos serves as the identifier of output.
+     * @dev _utxoPos not used, serves as the position identifier of output.
```
boolafish: Minor: I might move this to be below @notice as this is using @dev not @param.

plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol
```diff
@@ -20,15 +20,16 @@ contract PaymentOutputToPaymentTxCondition is IPaymentSpendingCondition {
...
      * @param _spendingTx The rlp encoded transaction that spends the output.
      * @param _inputIndex The input index of the spending transaction that points to the output.
      * @param _signature The signature of the output owner.
      */
     function verify(
         bytes32 _outputGuard,
-        uint256 _utxoPos,
-        bytes32 /*_outputId*/,
+        uint256, /*_utxoPos  NOTE: It's unclear how & if this will be used at all, see: https://github.com/omisego/plasma-contracts/pull/212*/
+        bytes32 _outputId,
```
boolafish: This naming would conflict with our normal definition of `outputId` as currently `outputId` means a specific way of computing the id. In here it means a more generic identifier (both utxoPos and outputId) and would lead to confusion.

IDK, how about sth like `outputIdentifier` ? (I know this...is not really better) or hope you can come up with some better naming :p.

pnowosie: This is pointing to input of the tx in-question. Maybe inputPointer or inputIdentifier ?

PR #215, opened 2019-08-20
test: use deposit tx as input to IFE in tests

kevsul:


plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
```diff
@@ -94,6 +93,11 @@ contract('PaymentInFlightExitable', ([_, alice, bob, carol]) => {
             return new PaymentTransaction(IFE_TX_TYPE, [input], [output]);
         }
 
+        function createDepositTransaction(owner, amount, token = ETH) {
+            const output = new PaymentTransactionOutput(amount, addressToOutputGuard(owner), token);
+            return new PlasmaDepositTransaction(output);
```
boolafish: this is self is not enough to make it a `deposit tx`, in fact, what we really need is to make sure it is mined in a deposit block number (that's how contract checks it). Basically I think we need two blocks in test, one child chain block an one deposit block to make sure we hit both branch.

plasma_framework/test/src/exits/payment/PaymentInFlightExitable.test.js
```diff
@@ -32,26 +32,36 @@ contract('PaymentInFlightExitable', ([_, alice, bob, carol]) => {
...
     const AMOUNT = 10;
     const TOLERANCE_SECONDS = new BN(1);
 
     describe('startInFlightExit', () => {
-        function buildValidIfeStartArgs(amount, [ifeOwner, inputOwner1, inputOwner2], blockNum) {
-            // TODO: make one of those a deposit transaction
+        function buildValidIfeStartArgs(amount, [ifeOwner, inputOwner1, inputOwner2], blockNum1, blockNum2) {
             const inputTx1 = createInputTransaction(DUMMY_INPUT_1, inputOwner1, amount);
-            const inputTx2 = createInputTransaction(DUMMY_INPUT_2, inputOwner2, amount);
+            const inputTx2 = createDepositTransaction(inputOwner2, amount);
```
boolafish: seems like the `blockNum1/2` does not need to be normal tx or deposit tx. Perhaps do a if-else here for both inputTx?
```
const inputTx = isDeposit(blockNum)? createDepositTransaction(...) : createInputTransaction(...);
```

kevsul:
The original RLP issue doesn't show up for me or on CI, so I'd merge it.
@pgebal do you want to check if it breaks for you now, or after merge? :trollface:

PR #220, opened 2019-08-21
Some fixes for deploying with truffle

kevsul:
# Fixes
- Add `await` when getting the authorityNonce
- Clean up the Infura URL before creating the HDWalletProvider


truffle-config.js
```diff
@@ -41,9 +41,11 @@ module.exports = {
     infura: {
       skipDryRun: true,
       provider: function () {
+        const infuraUrl = `${process.env.INFURA_URL}/${process.env.INFURA_API_KEY}`;
+        const cleanInfuraUrl = infuraUrl.replace(/([^:])(\/\/+)/g, '$1/');
```
kendricktan: I think a better way of checking if `INFURA_URL` has a trailing slash is to use the standard `url` library instead of regexing it e.g.

```javascript
const url = require('url')

url.resolve(process.env.INFURA_URL, process.env.INFURA_API_KEY)
```

kendricktan: Sounds pedantic but I'm not really a regex guru and it makes it harder to reason about 😓

kendricktan: Turns out I'm completely wrong, and that is the correct way to do it!

boolafish: oh interesting, is there any document or reference on this?

migrations/4_init_rootchain_contract.js
```diff
@@ -4,7 +4,12 @@ const RootChain = artifacts.require('./RootChain.sol')
 module.exports = async function (deployer, network, accounts) {
   const rootChain = await RootChain.deployed()
 
-  if (process.env.AUTHORITY_PASSPHRASE) {
+  // Unlock the deployer account if not using Infura
+  // As infura doesn't support eth_unlockAccount
+  if (process.env.DEPLOYER_PASSPHRASE &&
```
kevsul:
```suggestion
  if (process.env.AUTHORITY_PASSPHRASE &&
```

truffle-config.js
kevsul: Yeah, I started off using `url` as well, but it doesn't seem to do what we want 😞 
I thought the regex was better than pulling in an external package, but feel free to do so if you feel strongly about it.

migrations/1_check_authority_address.js
```diff
@@ -1,10 +1,36 @@
 require('dotenv').config()
-var RootChain = artifacts.require('./RootChain.sol')
+const RootChain = artifacts.require('./RootChain.sol')
+
 
 module.exports = async function (deployer, network, accounts) {
-  if (process.env.USE_EXISTING_AUTHORITY_ADDRESS) {
+  if (process.env.AUTHORITY_PRIVATEKEY && process.env.DEPLOYER_PRIVATEKEY) {
```
kevsul: Hmmm, there seems to be quite a bit of code duplication in this block: checking auth nonce and funding the auth account. 
Can we rearrange the logic to avoid duplication?

truffle-config.js
kendricktan: @boolafish even the most popular url-joining library uses regex, (and its not in the std url lib) https://github.com/jfromaniello/url-join/blob/master/lib/url-join.js

So I presume its the way

migrations/1_check_authority_address.js
kendricktan: Done! https://github.com/omisego/plasma-contracts/pull/220/commits/65d0f9757201122250a4c1e84d6b7c49bc15d0e3

migrations/1_check_authority_address.js
```diff
@@ -1,40 +1,68 @@
...
+const RootChain = artifacts.require('./RootChain.sol')
+
 
 module.exports = async function (deployer, network, accounts) {
-  if (process.env.USE_EXISTING_AUTHORITY_ADDRESS) {
+  // Get the deployer address
+  const deployerAddress = (process.env.DEPLOYER_PASSPHRASE) ?
+    await RootChain.web3.eth.personal.newAccount(process.env.DEPLOYER_PASSPHRASE) :
+    (process.env.DEPLOYER_ADDRESS || accounts[0])
+
```
kevsul: How about setting the DEPLOYER_ADDRESS and AUTHORITY_ADDRESS here, so that you don't have to remember to do the ` || accounts[0]` everywhere?
e.g.
`process.env.DEPLOYER_ADDRESS = deployerAddress`

kendricktan: Imo, environment variables are suppose to be set inside the environment, not the code.

As the migration process gets more complex, constantly mutating `process.env` will make it harder to debug in the long run

kevsul: Yeah, it's common enough practice but I agree that it can be confusing to modify the env.

Hmm, I still think that always having to remember that the DEPLOYER_ADDRESS (and AUTHORITY_ADDRESS) can be set either via `process.env.DEPLOYER_ADDRESS` or `process.env.DEPLOYER_PASSPHRASE` or in `accounts[0]` is kinda clunky and error-prone though. 
How about setting them as global properties instead?

```
global.deployerAddress = (process.env.DEPLOYER_PASSPHRASE) ?
    await RootChain.web3.eth.personal.newAccount(process.env.DEPLOYER_PASSPHRASE) :
    (process.env.DEPLOYER_ADDRESS || accounts[0])
```

kendricktan: Yes, its definitely clunky and `global.<x>` would be a nice compromise 👍

migrations/4_init_rootchain_contract.js
```diff
@@ -12,14 +12,14 @@ module.exports = async function (deployer, network, accounts) {
...
+      global.deployerAddress,
       process.env.AUTHORITY_PASSPHRASE,
       10
     )
   }
   // Call RootChain.init() from the authority account
   return rootChain.init(
     process.env.MIN_EXIT_PERIOD,
-    { from: process.env.AUTHORITY_ADDRESS || accounts[1] }
+    { from: global.authorityAddress || accounts[1] }
```
kevsul: You don't need to `|| accounts[1]` anymore, right?

migrations/5_output_results.js
```diff
@@ -5,6 +5,6 @@ module.exports = async function (deployer, network, accounts) {
   console.log(JSON.stringify({
     contract_addr: `${RootChain.address}`.toLowerCase(),
     txhash_contract: `${RootChain.network.transactionHash}`.toLowerCase(),
-    authority_addr: `${process.env.AUTHORITY_ADDRESS || accounts[1]}`.toLowerCase()
+    authority_addr: `${global.authorityAddress || accounts[1]}`.toLowerCase()
```
kevsul: Same as above, you don't need the `|| accounts[1]` anymore

migrations/4_init_rootchain_contract.js
```diff
@@ -12,14 +12,14 @@ module.exports = async function (deployer, network, accounts) {
   ) {
     // Unlock the authority account if necessary
     await RootChain.web3.eth.personal.unlockAccount(
-      process.env.AUTHORITY_ADDRESS,
+      global.deployerAddress,
```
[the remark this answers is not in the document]
kendricktan: Oops, good catch!

jbunce:
LGTM. @kevsul are you happy to merge this?

PR #222, opened 2019-08-22
feature: piggybackInFlightExit implementation

boolafish:
#### Main changes
1. Implement PiggybackInFlightExit and break the api into two: on input and on output. The main justification of this is that during two cases, some arguments required for output is never used in input (`outputType`, `outputGuardPreimage`). Plus there are some logic difference already.
1. Introduce 'WithdrawData' for InFlightExit inputs/outputs. Those would be the data needed when
process exit. We don't use PaymentOutput because we want to be tx type agnostic.
1. Extract some common operator for InFlightExit into InFlightExitModelLib.

#### Minor fixs
1. Remove OUTPUT_TYPE_ZERO in tests, also blocks the zero output type. This is something we don't need after the introduction of `OutputGuardHandler`. To be consistent, all types starts with 1.
1. Use `supportTxType` in startIFE. (introduced in challengeIFENoncanical).

### Test
```

    piggybackInFlightExitOnInput
      ✓ should fail when not send with the bond value (90ms)
      ✓ should fail when input index exceed max size of tx input (166ms)
      ✓ should fail when no exit to piggyback on (174ms)
      ✓ should fail when first phase of exit has passed (277ms)
      ✓ should fail when the same input has been piggybacked (515ms)
      ✓ should fail when not called by the exit target of the output (229ms)
      When piggyback successfully
        ✓ should enqueue with correct data when it is the first piggyback of the exit on the token
        ✓ should not enqueue when it is not first piggyback of the exit on the same token (122ms)
        ✓ should set the exit as piggybacked on the input index (55ms)
        ✓ should emit InFlightExitInputPiggybacked event

    piggybackInFlightExitOnOutput
      ✓ should fail when not send with the bond value (87ms)
      ✓ should fail when output index exceed max size of tx output (152ms)
      ✓ should fail when no exit to piggyback on (109ms)
      ✓ should fail when first phase of exit has passed (307ms)
      ✓ should fail when the same output has been piggybacked (510ms)
      When piggyback successfully
        ✓ should enqueue with correct data when it is the first piggyback of the exit on the token
        ✓ should not enqueue when it is not first piggyback of the exit on the token (325ms)
        ✓ should set the exit as piggybacked on the output index (42ms)
        ✓ should emit InFlightExitOutputPiggybacked event
      Given the piggyback output is of output type 0
        ✓ should fail when not called by the exit target of the output (496ms)
        ✓ should set the correct withdraw data on the output of exit data (377ms)
      Given the piggyback output is of non 0 output type
        ✓ should fail when output type is wrong and mismatch with output.outputguard (400ms)
        ✓ should fail when output guard data is wrong and mismatch with output.outputguard (373ms)
        ✓ should fail when there is no outputguard parser for the output type (416ms)
        ✓ should fail when not called by the exit target of the output (502ms)
        ✓ should set the correct withdraw data on the output of exit data (455ms)
```


plasma_framework/contracts/mocks/exits/payment/routers/PaymentInFlightExitRouterMock.sol
```diff
@@ -19,11 +23,28 @@ contract PaymentInFlightExitRouterMock is PaymentInFlightExitRouter {
...
+            ife.outputs[i] = exit.outputs[i];
+        }
+    }
+
+    function getInFlightExitInput(uint192 exitId, uint16 inputIndex) public view returns (PaymentExitDataModel.WithdrawData memory) {
         return inFlightExitMap.exits[exitId].inputs[inputIndex];
     }
 
-    function getInFlightExitOutput(uint192 exitId, uint8 outputIndex) public view returns (PaymentOutputModel.Output memory) {
+    function getInFlightExitOutput(uint192 exitId, uint16 outputIndex) public view returns (PaymentExitDataModel.WithdrawData memory) {
```
pnowosie: `uint8 -> uint16` Why?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelLib.sol
```diff
@@ -0,0 +1,88 @@
...
+library PaymentInFlightExitModelLib {
+    using Bits for uint256;
+
+    uint8 constant public MAX_INPUT_NUM = 4;
+    uint8 constant public MAX_OUTPUT_NUM = 4;
+
+    function setInputPiggybacked(ExitModel.InFlightExit storage ife, uint16 index)
+        internal
+    {
+        ife.exitMap = ife.exitMap.setBit(uint8(index));
```
pnowosie: `uint8(index)` it potentially loose precision. Are we sure values here will fit 8-bit only?

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelLib.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    function setOutputPiggybacked(ExitModel.InFlightExit storage ife, uint16 index)
+        internal
+    {
+        uint8 indexInExitMap = uint8(index + MAX_INPUT_NUM);
```
pnowosie: Hmm we assume that `2 * MAX_INPUT_INDEX` is 8-bit integer, but declare index as twice as big :question: Seems incorrect or I'm missing the point.... :nerd_face:

plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
```diff
@@ -0,0 +1,259 @@
...
+const { buildUtxoPos, utxoPosToTxPos } = require('../../../helpers/positions.js');
+const {
+    addressToOutputGuard,
+} = require('../../../helpers/utils.js');
+const { PaymentTransactionOutput, PaymentTransaction } = require('../../../helpers/transaction.js');
+
+contract('PaymentInFlightExitRouter', ([_, alice, bob]) => {
+    const PIGGYBACK_BOND = 31415926535; // wei
+    const ETH = constants.ZERO_ADDRESS;
+    const MIN_EXIT_PERIOD = 60 * 60 * 24 * 7; // 1 week
```
pnowosie: I miss this all the time :nerd_face:. And all the time I read such is... yyy wait :joy: 
```suggestion
    const MIN_EXIT_PERIOD = 60 * 60 * 24 * 7; // 1 week in seconds
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentPiggybackInFlightExitOnOutput.sol
```diff
@@ -0,0 +1,146 @@
...
+import "../../utils/ExitId.sol";
+import "../../utils/OutputGuard.sol";
+import "../../../framework/PlasmaFramework.sol";
+import "../../../framework/interfaces/IExitProcessor.sol";
+import "../../../transactions/outputs/PaymentOutputModel.sol";
+import "../../../transactions/PaymentTransactionModel.sol";
+import "../../../utils/IsDeposit.sol";
+import "../../../utils/UtxoPosLib.sol";
+
+library PaymentPiggybackInFlightExitOnOutput {
```
pnowosie: I'm not sure the split to handle Inputs & Outputs separately was justified. Both looks very similar except the end of the `run`... I'm not sure what solidity offers to abstract just this difference - maybe simple `if` isn't that bad?

plasma_framework/contracts/mocks/exits/payment/routers/PaymentInFlightExitRouterMock.sol
boolafish: we bump the general `outputIndex` from uint8 to uint16 from this issue: https://github.com/omisego/plasma-contracts/issues/168. As a result, I prefer uint16 in general when we are pointing to outputIndex.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelLib.sol
boolafish: `setBit` itself have 2^8 (256) bits to be set/flag actually. And the index of PaymentTx is bounded by `MAX_INPUT_SIZE` (4 for now) so even though I have a preference of uint16 to represent index in general, it is more than safe to claim it enough for PaymentTx. However, as the `setBit` only handles 2^8 bits, it's capped to uint8 when choosing the position to flag.

Aw, I personally treat storage as optimization so sometimes we store things with smaller size. But I just feel being consistent in other places might be easier?

plasma_framework/contracts/src/exits/payment/controllers/PaymentPiggybackInFlightExitOnOutput.sol
boolafish: The strongest argument was the argument is different as during input there is no need to handle outputGuard with preimage. So some args would always be empty during piggyback input.

However, in #227 we decides to make piggyback more consistent in logic.  I will take a look again.

boolafish: Merged back to one.

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
```diff
@@ -39,8 +41,10 @@ library PaymentExitDataModel {
          */
         uint256 exitMap;
         uint256 position;
-        PaymentOutputModel.Output[MAX_INPUT_NUM] inputs;
-        PaymentOutputModel.Output[MAX_OUTPUT_NUM] outputs;
+        // TODO: need to save the output guard during IFE
+        bytes32[MAX_INPUT_NUM] outputGuardForInputs;
```
boolafish: @pgebal I end up still need the `outputGuard` for inputs, ah.........(just realized that getting output guard from unknown tx type is......fun/complex, so I think it might be better to do that only in startIFE)

Now I re-think, probably it worth to keep inconsistence and let the input save the exitTarget so we don't need extra storage? : (

pgebal: @boolafish ok, I'll code that tomorrow. I'll change WithdrawData to store exit target instead of output guard. For in-flight transaction outputs I'll store only outputId, leaving exitTarget with default value of zeroes.

plasma_framework/contracts/src/exits/payment/controllers/PaymentPiggybackInFlightExitOnOutput.sol
boolafish: ah....because of [this decision](https://github.com/omisego/plasma-contracts/pull/222#discussion_r318611936), it ends up still being inconsistence. So still break it to functions as I think it is more clean. However, I put them in same controller Lib so we can be more dry on helper functions. Also the complexity of that lib is still okay.

boolafish:
@pnowosie can you review again on this? I've updated with some changes after merging with the master, also the result after discussion with @pgebal.

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelLib.sol
```diff
@@ -0,0 +1,81 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "../../utils/Bits.sol";
+import { PaymentExitDataModel as ExitModel } from "./PaymentExitDataModel.sol";
+
+library PaymentInFlightExitModelLib {
```
pnowosie: I wonder about more descriptive name to show its purpose...
`PaymentIFEModelUtils`, `PaymentIFEModelCommon`, `PaymentIFEModelHelper` ...

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
```diff
@@ -50,8 +51,10 @@ library PaymentChallengeIFENotCanonical {
         );
         require(address(condition) != address(0), "Spending condition contract not found");
 
+        // FIXME: move to the finalized interface as https://github.com/omisego/plasma-contracts/issues/214
+        // Also, the tests should verify the args correctness
         bool isSpentByInFlightTx = condition.verify(
-            ife.inputs[args.inFlightTxInputIndex].outputGuard,
+            bytes32(''), // tmp solution, we don't need outputGuard anymore for the interface of 214
```
pnowosie: I was surprised seeing just a number without realizing its related to prev comment
```suggestion
            bytes32(''), // tmp solution, we don't need outputGuard anymore for the interface of :point-up: GH-214
```

plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelLib.sol
```diff
@@ -0,0 +1,67 @@
...
+        uint256 periodTime = minExitPeriod / 2;
+        return ((block.timestamp - ife.exitStartTimestamp) / periodTime) < 1;
+    }
+
+    function isFinalized(ExitModel.InFlightExit memory ife)
+        internal
+        pure
+        returns (bool)
+    {
+        return Bits.bitSet(ife.exitMap, 255);
```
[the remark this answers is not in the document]
boolafish: just realize this function would be removed after the introduction of `isFinalized` bool : )

PR #226, opened 2019-08-23
feat: replace OutputGuardParser with OutputGuardHandler

boolafish:
### Note
In previous implementation, we are bind to using output type 0 as some special
type that is not using outputGuard mechanism, which is hashing the preImages.
This commit tries to abstract that out and remove such assumption.

We introduce OutputGuardHandler that would take in OutputGuardModel.Data which
includes all essential data to verify a outputguard is valid or not. OuputGuardModel.Object
contains: guard, outputType, and preImage. Also, we use the OutputGuardHandler to retrieve
the exit target that is needed in ExitGame.


plasma_framework/contracts/src/exits/payment/controllers/PaymentStartStandardExit.sol
```diff
@@ -99,28 +100,23 @@ library PaymentStartStandardExit {
         bool isTxDeposit = controller.isDeposit.test(utxoPos.blockNum());
         uint192 exitId = ExitId.getStandardExitId(isTxDeposit, args.rlpOutputTx, utxoPos);
         (bytes32 root, uint256 blockTimestamp) = controller.framework.blocks(utxoPos.blockNum());
-        bytes memory outputGuardData = args.outputType == 0 ? bytes("") : args.outputGuardData;
-
-        address payable exitTarget;
-        IOutputGuardParser outputGuardParser;
-        if (args.outputType == 0) {
```
boolafish: This is the main cleanup, no longer need to rely on a preserved "output type 0".

plasma_framework/contracts/mocks/exits/payment/outputGuardHandlers/PaymentOutputGuardHandlerWrapper.sol
```diff
@@ -0,0 +1,43 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "../../../../src/exits/models/OutputGuardModel.sol";
+import "../../../../src/exits/payment/outputGuardHandlers/PaymentOutputGuardHandler.sol";
+
+/**
+For unkown reason truffle is not able to pass in the struct of OutputGuardModel.Data as input args directly.
```
pik694: Try passing a tuple instead

plasma_framework/test/src/exits/payment/outputGuardHandlers/PaymentOutputGuardHandler.test.js
```diff
@@ -0,0 +1,52 @@
...
+            const guard = addressToOutputGuard(alice);
+
+            await expectRevert(
+                this.handler.isValid(guard, TEST_OUTPUT_TYPE, nonEmptyPreimage),
+                'Pre-imgage of the output guard should be empty',
+            );
+        });
+
+        it('should fail when output type mismatch', async () => {
+            const preimage = '0x';
```
pik694: is it necessary to explicitly put `'0x'` in this case? cannot it be just `''` as empty preimage?

boolafish: truffle would not translate `string` from js to `bytes` in solidity, however, it translates `hex` to `bytes`. That's why our tests uses hex to represent bytes input.

plasma_framework/contracts/mocks/exits/payment/outputGuardHandlers/PaymentOutputGuardHandlerWrapper.sol
boolafish: after testing, tuple failed but array works. 
I also tried, if we return a struct from contract, it would be represented as an `Array`.

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -36,8 +37,8 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
     const ETH = constants.ZERO_ADDRESS;
     const INITIAL_ERC20_SUPPLY = 10000000000;
     const DEPOSIT_VALUE = 1000000;
-    const OUTPUT_TYPE_ZERO = 0;
     const EMPTY_BYTES = '0x';
+    const PAYMENT_OUTPUT_TYPE = 1;
```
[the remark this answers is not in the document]
boolafish: Seems like js doesn't have enum: https://stijndewitt.com/2014/01/26/enums-in-javascript/
do you mean like the following?

```
const TYPE_ENUM = {
    PAYMENT_OUTPUT = 1,
    PAYMENT_TX = 1,
};

// and then use it like the following
TYPE_ENUM.PAYMENT_OUTPUT
TYPE_ENUM.PAYMENT_TX
```

boolafish:
I will merge this as approval given and wants to use this on the piggyback PR so we don't need to refactoring back and forth. The only comment I am not sure is about enums. If you think it worth handling that I will file another small PR for that afterward @pik694 .

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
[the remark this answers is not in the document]
pik694: Od this is the only option it is still better than having lots of constants flying around

plasma_framework/test/src/exits/payment/outputGuardHandlers/PaymentOutputGuardHandler.test.js
pik694: so what about extracting a const such as  `EMPTY_BYTES='0x'` ?

Issue #214, opened 2019-08-19
A more generalized SpendingCondition Registry


boolafish:
### Why
At the time a developing Standard Exit, `PaymentSpendingConditionRegistry` was introduced to cover the spending condition concept (or the output predicate). However, in SE, it only cares about the outputting case `PaymentOutput -> Some tx` and does not take input into consideration: `Some Tx -> Payment Tx`.

When start IFE PR is out, I sensed that the original design is not really designed for the input scenario. The registry naming and the interface might not be generalized enough if we want to re-use that for a more global use.

### Proposal interface of registry
```
function verify(
    bytes inputTx,
    uint16 outputIndex,
    uint256 inputTxPos,
    bytes spendingTx,
    uin16 inputIndex,
    bytes witness,
    bytes optionalArgs,
) external view returns (bool);
```
With `inputTx` + `outputIndex`, one should be able to get the `output` data. Why passing in `inputTx` instead of `output` though? In the exit game, it is more common that the function would accept encoded tx instead of encoded output. As a result, if we are passing in encoded output, we would need to re-encode in our exit game function somewhere.

Also, with `inputTx` + `inputTxPos` + `outputIndex`, we can build both `utxoPos` and `outputId` depends on our need. However, `inputTxPos` must be proven correct somewhere outside the spending condition.

Optional args is there to cover things that is to be customized:
1. output guard pre-image might be a good fit there.
1. in the case of DEX settlement, where inputs/outputs of tx might just be a merkle root when decoded, we need to pass in the original output data + merkle proof and not get those from decoded tx.
We can pass in an `abi.encoded(customStruct)` and in our contract.

### Proposed Functionality of the registry
Instead of using the origin `(outputType, spendingTxType)` pair, we would like to add another `inputTxType` to isolate the registration process. Biggest reason is that we want to `Freeze` the registry of an input tx type to accept new spending txs. That is, we first register spending conditions of a input tx type, and then we would freeze future registration on such tx type. We only deploy our exit game contract after such tx type is frozen so user knows how the tx can be spent and know it would not be changed.

Also, this means now (`inputTxType`, `outputType`) should become the unique type identifier of an output. Same `outputType` in different `inputTxType` can mean different things now.

Some basic functionalities of the registry:

```
register(inputTxType, outputType, spendingTxType, spendingConditionContract);
freezeInputTxType(inputTxType);
```

### Where to put this Registry?
[...]

PR #227, opened 2019-08-26
feature: make starting in-flight exit input transaction type agnostic

### LINKED ISSUE #205: Start Inflight Exit should be input tx type agnostic
### What
In start IFE PR: https://github.com/omisego/plasma-contracts/pull/196, it directly uses `PaymentTransaction` as input transaction. We would like to not hardcode input tx in the `PaymentInflightExitable`.

### Why
We want to be able to **not** audit again the main code when we are adding DEX feature or extend to any other tx that can be input of `Payment transaction`.

### Proposal solution
We could probably have a `verifyStateTransition` predicate for IFE. This would basically check given input txs and the tx with the proof/witness, the state transition in such transaction is valid.

We would swap the predicate when we are changing/adding new input txs. Dependency tx types can be hardcoded there.

### Successful Critieria
- The `PaymentInflightExitable` should not hardcode `PaymentOuput` as input. (one example [here](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitable.sol#L243), should be more in current code base)
- In `PaymentExitDataModel`, `inputs` of `InflightExit` should not be `PaymentOutput`: [code here](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol#L34). (I guess need to be sth like `bytes`, or if there is better idea?)

pgebal:
Closes #205


plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -212,18 +213,18 @@ library PaymentStartInFlightExit {
...
+                );
         }
     }
 
-    function verifyInputsSpendingCondition(StartExitData memory exitData) private view {
+    function verifyInputsSpent(StartExitData memory exitData) private view {
         for (uint i = 0; i < exitData.inputTxs.length; i++) {
             uint16 outputIndex = exitData.inputUtxosPos[i].outputIndex();
-            bytes32 outputGuard = exitData.inputTxs[i].outputs[outputIndex].outputGuard;
+            WireTransaction.Output memory output = WireTransaction.getOutput(exitData.inputTxs[i], outputIndex);
```
pgebal: Here a "cheat" is done.
Objective of the PR is to abstract from input transaction types. 
This line assumes, that we can extract some "output" data from transaction knowing only that it conforms to wire transaction format (as in Wire Transaction Format Proposal doc).
It would be very handy and make life easier, if all our transaction conforms to that format.
If we don't want to assume that. We could use an interface here, like we do with state transition verifier.

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
```diff
@@ -25,6 +25,13 @@ library PaymentExitDataModel {
         mapping (uint192 => PaymentExitDataModel.StandardExit) exits;
     }
 
+    struct WithdrawData {
+        bytes32 outputGuard;
```
boolafish: This must be `exitTarget` or, something represents the address to withdraw.

plasma_framework/contracts/mocks/exits/payment/routers/PaymentInFlightExitRouterMock.sol
```diff
@@ -19,11 +20,11 @@ contract PaymentInFlightExitRouterMock is PaymentInFlightExitRouter {
         inFlightExitMap.exits[exitId].exitMap = Bits.setBit(inFlightExitMap.exits[exitId].exitMap, 255);
     }
 
-    function getInFlightExitInput(uint192 exitId, uint8 inputIndex) public view returns (PaymentOutputModel.Output memory) {
+    function getInFlightExitInput(uint192 exitId, uint8 inputIndex) public view returns (PaymentExitDataModel.WithdrawData memory) {
```
boolafish: minor: for `inputIndex ` and `outputIndex` below, probably bump it to uint16 to be consistent with our current index size.

plasma_framework/contracts/src/exits/IStateTransitionVerifier.sol
```diff
@@ -0,0 +1,17 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+interface IStateTransitionVerifier {
```
boolafish: I would put this under `.../src/exits/interfaces/IStateTransitionVerifier.sol`. I think we will have a few interfaces for to be put there: `IOutputGuardHandler` (#227 ), `ISpendingCondition`(#214 ) and this one (and potentially more).

plasma_framework/contracts/src/transactions/WireTransaction.sol
```diff
@@ -0,0 +1,35 @@
...
+    using RLP for RLP.RLPItem;
+
+    struct Output {
+        uint256 amount;
+        bytes32 outputGuard;
+        address token;
+    }
+
+    /**
+    * @notice Returns output for transaction in wire format.
```
boolafish: I would add the assumption of using this function in @dev here.
Like the `outputs` must be a field on the second index and should be a list which is the format of {amount, outputGuard, token} accordingly.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -309,14 +270,39 @@ library PaymentStartInFlightExit {
...
+    }
+
+    function setInFlightExitOutputs(
+        PaymentExitDataModel.InFlightExit storage ife,
+        StartExitData memory exitData
+    )
+        private
+    {
+        for (uint i = 0; i < exitData.inFlightTx.outputs.length; i++) {
+            bytes32 outputId = OutputId.computeNormalOutputId(exitData.inFlightTxRaw, i);
```
boolafish: I would add a comment that deposit tx cannot in-flight exit.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -309,14 +270,39 @@ library PaymentStartInFlightExit {
...
+            ife.inputs[i] = PaymentExitDataModel.WithdrawData(
+                output.outputGuard,
+                exitData.outputIds[i],
+                output.token,
+                output.amount
+            );
+        }
+    }
+
+    function setInFlightExitOutputs(
```
boolafish: haha okay, if we are setting outputs here, I will think again whether I want to separate the piggyback into two functions or not in PR #222 . I personal have preference over simplicity than optimization so I do like to set outputs here to make logic consistent. Let's see how the gas goes.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
boolafish: hmm, with #214 the new interface probably wouldn't need to getOutput here. However, we do need at least some way to get the `WithdrawData` in `setInputs(...)`. 

For DEX related txs as input tx this assumption seems fine as all tx outputs would be using same format anyway. For `ERC721` I think it *can* work because we are only getting output here, but I doubt the tx format would be as described in the document. I feel when we introduce atomic swap tx for fungible and non fungible token, it would be more or less like the following:

```
{
    txType: int.
    inputs: [bytes32, ...],
    fungibleTokenOutputs: [{ }, { },...],
    nonfungibleTokenOutputs: [{ }, { }, ...],
    // other fields
}
```

Reason of two output lists is that they need different data structure. I don't think current {amount, outputGuard, token} could fit non-fungible token output. And, at least in solidity, you cannot have one outputs array with multiple struct within it.

So, yes, it *can* work as long as we put the fungible token outputs in the second indexed position. However, whether it is a good idea to do would be more controversial. I think it strongly depends on what `WireTransaction` means and what assumption we are talking about when we are introducing such tx format.

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
```diff
@@ -0,0 +1,119 @@
...
+            inputs[i] = output;
+        }
+
+        WireTransaction.Output[] memory outputs = new WireTransaction.Output[](inputTxs.length);
+        PaymentTransactionModel.Transaction memory transaction = PaymentTransactionModel.decode(inFlightTx);
+        for (uint i = 0; i < transaction.outputs.length; i++) {
+            outputs[i] = WireTransaction.Output(transaction.outputs[i].amount, transaction.outputs[i].outputGuard, transaction.outputs[i].token);
+        }
+
+        return _isCorrectStateTransition(inputs, outputs);
```
boolafish: Since you're reverting on length mismatch earlier, probably we can consider reverting only strategy to keep consistent or we return false on the earlier. I am on one hand not too much a fan of this reverting strategy as the boolean value become weird, but on the other hand reverting here provides better customized revert message. Also, for unknown reason, revert + return true at the end seems to be the common practice in smart contract (?) 

Anyway, I will leave this to your decision.

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
```diff
@@ -0,0 +1,119 @@
...
+            if (outputs[i].token == token) {
+                outputsWithToken[j] = outputs[i];
+                j += 1;
+            }
+        }
+
+        return outputsWithToken;
+    }
+
+    function isCorrectSpend(
```
boolafish: Minor: would `isCorrectSumRelationship` better describe the function?

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -101,18 +111,18 @@ library PaymentStartInFlightExit {
...
         exitData.inputTxsInclusionProofs = args.inputTxsInclusionProofs;
         exitData.inFlightTxWitnesses = args.inFlightTxWitnesses;
-        exitData.outputIds = getOutputIds(controller, exitData.inputTxsRaw, exitData.inputUtxosPos);
+        exitData.outputIds = getOutputIds(controller, exitData.inputTxs, exitData.inputUtxosPos);
         return exitData;
     }
 
     function decodeInputTxsPositions(uint256[] memory inputUtxosPos) private pure returns (UtxoPosLib.UtxoPos[] memory) {
-        require(inputUtxosPos.length <= MAX_INPUT_NUM, "To many input transactions provided");
+        require(inputUtxosPos.length <= MAX_INPUT_NUM, "To many transactions provided");
```
boolafish: typo: "Too" many....

plasma_framework/test/src/transactions/WireTransaction.test.js
```diff
@@ -0,0 +1,39 @@
...
+        const encoded = web3.utils.bytesToHex(transaction.rlpEncoded());
+
+        const actual = await this.test.getOutput(encoded, 0);
+
+        expect(actual.outputGuard).to.equal(OUTPUT_GUARD);
+        expect(new BN(actual.amount)).to.be.bignumber.equal(new BN(AMOUNT));
+        expect(actual.token).to.equal(constants.ZERO_ADDRESS);
+    });
+
+    it('should decode custom transaction output', async () => {
```
boolafish:
```suggestion
    it('should decode custom transaction output that fulfills wire transaction format', async () => {
```

plasma_framework/test/src/exits/payment/PaymentStartInFlightExit.test.js
```diff
@@ -429,11 +431,24 @@ contract('PaymentInFlightExitRouter', ([_, alice, bob, carol]) => {
...
+                    MIN_EXIT_PERIOD, DUMMY_INITIAL_IMMUNE_VAULTS_NUM, INITIAL_IMMUNE_EXIT_GAME_NUM,
+                );
+                this.exitGame = await PaymentInFlightExitRouter.new(
+                    this.framework.address,
+                    this.spendingConditionRegistry.address,
+                    this.stateTransitionVerifierReject.address,
+                );
+            });
+
+            it('should fail when in-flight transaction is an invalid state transistion', async () => {
```
boolafish:
```suggestion
            it('should fail when the in-flight transaction has an invalid state transistion', async () => {
```

boolafish: I might try to test both verifier returns false and reverts. I think both are potential results in such verifier.

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
boolafish: leave some discussion from slack:

rambo.pawel Today at 4:47 PM
@boolafish I think we can't store beneficiaries of exits when starting IFE, because that means we would have to provide output guard data to recover output owners from output guard, and that's private and may not be known by someone starting ife. We need to go with storing output gaurd, and then interested users provide output guard data on piggybacks @pepesza, is that correct? Maybe I am missing something about that privacy. (edited)
2 replies

rambo.pawel  < 1 minute ago
I talked to Andy, and output guard data is available for input transactions (already mined, included in chain), and may not be available for in-flight transaction itself. So we'll require to provide output guard data on piggyback.

rambo.pawel  < 1 minute ago
when starting IFE output guard data for input transactions is still needed for checking signatures (edited)

pepesza  1 minute ago
Yes, when owner is piggybacking, she needs to reveal the output guard data. This is a moment we fill in the blanks

pepesza  < 1 minute ago
Also, this mechanism only provides some privacy for funds at rest. When funds are being spend or are exited - data is being made public

---
Some notes agreement:
1. startIFE passes in the output guard pre-image. However, only used to check confirm sig (and would not be implemented in this PR). Beneficiaries is not checked at that time.
1. piggyback (input/output) checks the real beneficiaries of exit from output guard preimage and store it. Though input *can* do that while startIFE, move it to piggyback to make piggyback input/output logic more consistent.

plasma_framework/test/src/exits/payment/PaymentStartInFlightExit.test.js
pgebal: done

plasma_framework/test/src/transactions/WireTransaction.test.js
pgebal: done

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
pgebal: done

plasma_framework/contracts/src/transactions/WireTransaction.sol
pgebal: done

plasma_framework/contracts/src/exits/IStateTransitionVerifier.sol
pgebal: done

plasma_framework/contracts/mocks/exits/payment/routers/PaymentInFlightExitRouterMock.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
pgebal: Let's wait for clarification. We'll update that if needed. As of now I would leave it like it is, as it conforms to Wire Transaction Format doc.

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
pgebal: Done. I went for always returning a boolean. It's more flexible (but on the other hand what can a caller do if it returns false?, will probably always reverse).

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
boolafish: sigh) I end up realize it is such a hassle to get output guard from unknown input tx. So I think probably we should only do such complex thing once. In my piggyback PR I end up asking for an extra array of outputGuards for input. 

like to discuss with you on this: https://github.com/omisego/plasma-contracts/pull/222/files#r318546401

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol
boolafish: As we agree on having `exitTarget` set during inputs, let's sync on the data structure of our PRs so we can decrease merging conflicts. Below is my proposal:

```
    struct WithdrawData {
         bytes32 outputId;
         address payable exitTarget;
         address token;
         uint256 amount;
     }
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
boolafish: sure!

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
boolafish: haha not really done XD but this is minor anyway

boolafish:
For some reason I am still awake lol. The change for output guard is outdated. Please check the newer code with outputguardHandler

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
boolafish: Let's just take in WireTransaction as it is for this PR. I am just noting down more concerns with current approach.

During the call with @paulperegud and @pdobacz yesterday, we are talking about that even the `output` for a DEX settlement might not follow this original structure, where we have to make `outputType` explicit on the output itself. So sth like `(outputType, outputGuard/owner, amount, token)`.

TL;DR we might need to have 2 kinds of output in DEX where one send all funds back to user and another keep fund in exchange due to the attack scenerio described [here](https://github.com/omisego/docs/blob/0773484b2a16872e8c730a4f8648d1dcb317a9b1/docs/restricted_custody_mvp1_spec.md#alice-wants-to-withdraw-an-initial-deposit-from-an-offline-venue-but-settlement-occurs). To be able to differentiate which mechanism to parse the `outputGuard/owner` as ch-ch would need to check when collecting txs, we need a flag thus we need the `outputType` being explicit.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -309,14 +284,50 @@ library PaymentStartInFlightExit {
...
     {
-        for (uint i = 0; i < inputTxs.length; i++) {
-            uint16 outputIndex = inputUtxosPos[i].outputIndex();
-            ife.inputs[i] = inputTxs[i].outputs[outputIndex];
+        for (uint i = 0; i < exitData.inputTxs.length; i++) {
+            uint16 outputIndex = exitData.inputUtxosPos[i].outputIndex();
+            WireTransaction.Output memory output = WireTransaction.getOutput(exitData.inputTxs[i], outputIndex);
+
+            address payable exitTarget;
+            if (exitData.inputUtxosTypes[i] == 0) {
```
boolafish: You'll need to change this when merging with master. As I stated in another comment we changed the way of handling outputGuard

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
```diff
@@ -15,6 +16,7 @@ library PaymentInFlightExitRouterArgs {
         bytes[] inputTxs;
         uint256[] inputUtxosPos;
         uint256[] inputUtxosTypes;
+        bytes[] outputGuardDataPreImages;
```
boolafish: Minor: I usually use `outputGuardPreimage` in other places. But I will leave this to you.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
pgebal: done

PR #232, opened 2019-08-29
Output guard pr comment

boolafish:
This is a follow up PR that handles @pik694 's PR comments here: https://github.com/omisego/plasma-contracts/pull/226


boolafish:
So the original comment is more on having a helper constants file to handle the share values for tests instead of this. Closing this and would find time to that refactor.

PR #239, opened 2019-09-03
Separation of operator from authority address

pnowosie:
Fixes #233

PR #241, opened 2019-09-04
Separation of maintainer and authority addresses

### LINKED ISSUE #233: Separation of operator address
### Why
We rely on the authentication address (operator address) to be fresh for re-org protection.
https://github.com/omisego/elixir-omg#managing-the-operator-address

However, we need to limit to an address for register new contracts. This seems to conflict with the nonce mechanism of tracking submitBlock only as the registration would break the assumption.

### Current Implementation
Operator address is set on deployment now. This must be change to respect the nonce mechanism. However, this would not be enough for the registry.

# Solution description

## Terminology
- __Operator__ - organization, not an Ethereum address which holds other addresses
- __Authority__ - Ethereum address used exclusively to submit plasma blocks, requires `nonce = 0` as noted below.
- __Deployer__ - Ethereum address used to deploy Plasma framework for the first time
- __Maintainer__ - Ethereum address (can be the same as deployer or Multisig contract) that is allowed to register upgrades (tx types, output types, vaults, exit games, ...)

## Deployment process

Authority address sends __its very first__ transaction that calls `BlockController.initAuthority()` function. This function sets a caller address as Authority address in `BlockController` which allows for block submission and sets a Authority address nonce - so the Plasma blocks numbers will correspond to the address nonces (`1000 * nonce = blknum`).

## Changing an Authority address
To allow #118 it's possible to replace `authority` address by calling `BlockController.setAuthority(address)` from the `maintainer`. 

## Deployer - maintainer separation
Further separation of `deployer - maintainer` is continued in #242

pnowosie:
Fixes #233 
This :pr: separates `authority` address from the `operator` address.
Where:
- `authority` - an address which publishes new plasma blocks on the Ethereum
- `operator` - an address which deployed `PlasmaFramework` contract and is allow to register components to it such as Exit games, Vaults ...

It's worth to further separate `operator aka maintainer` from the `deployer` address - but this effort failed because of circular dependencies see: #239

### DIFF
--- a/plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
+++ b/plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
@@ -81,6 +81,7 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
 
     const setupContracts = async () => {
         this.framework = await PlasmaFramework.new(MIN_EXIT_PERIOD, INITIAL_IMMUNE_VAULTS, INITIAL_IMMUNE_EXIT_GAMES);
+        await this.framework.init();
 
         const ethDepositVerifier = await EthDepositVerifier.new();
         this.ethVault = await EthVault.new(this.framework.address);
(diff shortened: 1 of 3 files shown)


plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -80,6 +80,7 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
 
     const setupContracts = async () => {
         this.framework = await PlasmaFramework.new(MIN_EXIT_PERIOD, INITIAL_IMMUNE_VAULTS, INITIAL_IMMUNE_EXIT_GAMES);
+        await this.framework.init();
```
boolafish: Looking at this, I think the function name `init()` might be misleading.
Perhaps make the function name `initAuthority` instead?

PR #249, opened 2019-09-05
feat: implement challenge IFE input spent

kevsul:
## Description
Implements challenge IFE input spent.

## Tests
```
  Contract: PaymentChallengeIFEInputSpent
    challenge in-flight exit input spent
      after successfully challenged IFE input spent
        ✓ should emit InFlightExitInputBlocked event
        ✓ should remove the input from piggybacked (93ms)
        ✓ should pay the piggyback bond to the challenger
      check exitMap before and after challenge
        ✓ should remove the input from piggybacked (202ms)
      failures
        ✓ should fail when ife not started (71ms)
        ✓ should fail when the indexed input has not been piggybacked (79ms)
        ✓ should fail when spending tx is the same as in-flight one (80ms)
        ✓ should fail when the spending transaction input index is incorrect (132ms)
        ✓ should fail when the spent input is not the same as piggybacked input (184ms)
        ✓ should fail when spending condition for given output is not registered (187ms)
        ✓ should fail when spending condition is not met (282ms)
```


plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,87 @@
...
+            args.spendingTxInputIndex,
+            args.spendingTxWitness
+        );
+        require(isSpentByInFlightTx, "Spent input spending condition is not met");
+
+        // Remove the input from the piggyback map
+        ife.clearInputPiggybacked(args.inFlightTxInputIndex);
+
+        // Pay out the bond.
+        self.ethVault.withdraw(msg.sender, PIGGYBACK_BOND);
```
boolafish: so far the current implementation would actually keep the bond fund on the ExitGame contract directly instead of using vault since the "value" in transaction would by default goes to ExitGame when start exits...(and we do not explicit move funds to vault so far)

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,87 @@
...
+        EthVault ethVault;
+    }
+
+    event InFlightExitInputBlocked(
+        address indexed challenger,
+        bytes32 txHash,
+        uint8 inputIndex
+    );
+
+    function run(
```
boolafish: Just noticed when I am doing new spending condition interface for challenge SE.
(I think) We need to check `isProtocolFinalized()` whenever calling spending condition. This is new and previously no action need to be done due to every tx is MoreVP....

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,87 @@
...
+        require(ife.exitStartTimestamp != 0, "In-flight exit doesn't exist");
+
+        require(ife.isInputPiggybacked(args.inFlightTxInputIndex), "The indexed input has not been piggybacked");
+
+        require(
+            keccak256(args.inFlightTx) != keccak256(args.spendingTx),
+            "The spending transaction is the same as the in-flight transaction"
+        );
+
+        PaymentTransactionModel.Transaction memory spendingTx = PaymentTransactionModel.decode(args.spendingTx);
```
boolafish: Hmm...I think we cannot really assume the spending tx would be a Payment tx. Imagine a ERC721 tx is your input tx.

However, I think this check + the following (checking input pointing to the right output) should be done in SpendingCondition.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: Oh, I didn't notice that 🤦‍♂ 
So bonds don't ever go into the vault. Is that on purpose? The amounts are low anyway, so I guess it's simpler?

boolafish: it's not on purpose, we can change that if that seems to simply things?
It is just currently we didn't have extra code in start exit to move fund from ExitGame contract to vault, and I don't know which is better to be honest.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: So should the spendingTx type be passed in as a parameter?

BTW is your argument true for this line as well?
```
IPaymentSpendingCondition condition = self.spendingConditionRegistry.spendingConditions(
```
If we don't know it's a Payment tx, then we can't use a PaymentSpendingCondition?

boolafish: hmmm, yeah I guess we would need the type to be passed in.
As for second question, that is why this issue: https://github.com/omisego/plasma-contracts/issues/214 and that is what I am working on right now.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: Ok. For now I'll change this to be consistent with other bonds (i.e. keep it in the contract),
and create an issue to decide if we want to move it to the vault

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: Ok.
However, we still need to decode the spendingTx to be able to check that the specified input really is the same as the piggybacked input. Should we use WireTransaction for this?

boolafish: This is what I am thinking:
1. In ExitGame, build the `outputId` using the `inputTx + outputIndex(of input tx) + utxoPos of inputTx` (Spending Condition input). Verify the outputId is same as what we saved in exit data of IFE (we have `outputId` in WithdrawData). This makes sure we are using the same output of the IFE input.
1. Inside SpendingCondition, it would verify whether the spending tx is pointing to the output as expected. So we would make sure the spending tx is pointing to the output from the data given (`inputTx, outputIndex, txPos`)
1. So instead of directly checking spendingTx and IFE tx have same pointer, we verify both are pointing to same output. Though one is done inside ExitGame and the other done in SpendingCondition.

kevsul: Sorry, not really sure what you're getting at 🤔 
Which `inputTx` you're referring to here?

- In PaymentStartInFlightExit the `outputId` of the IFE tx input is built and put into WithdrawData
- In PaymentChallengeIFEInputSpent, the inFlightTx is passed in so `outputId` of the IFE tx input can be retrieved using the `exitId`.
- Are you saying that the the caller has to pass in the `outputId` of the spending tx input, so that it can be passed into SpendingCondition.verify?

boolafish: > Which inputTx you're referring to here?

Ah...I was in the mind of using new generalized spending condition interface. Wasn't realized it's never showed up in current code here. That is the input tx of ife tx and spending tx.

> Are you saying that the the caller has to pass in the outputId of the spending tx input, so that it can be passed into SpendingCondition.verify?

Ah, again, I was in the mind set of #214.
So not really the `outputId` directly, but the original data for `outputId` (txBytes, outputIndex, utxoPos).
(can be build from spending condition args: bytes inputTx, uint16 outputIndex,  uint256 inputTxPos)

---
hmm, let's me rephrase a bit. So the main problem here is that both input tx (of ife) and spending tx (of input tx) can be not `PaymentTx`. However, we want to be able to check whether the spending tx is pointing to the same input as the in-flight exit tx is pointing to.

instead of
```
require(ifeInput == spendingInput, "Spent input is not the same as piggybacked input")
```

I am saying we try to the following in exit game
```
// outputId is from Withdraw data, has same value of "ifeInput" as above 
// building the outputId itself does not require the need to decode transaction, so it is fine to do this in ExitGame
require(outputId == buildedOutputIdFromSpendingConditionArgs, "....");
```
and then the spending condition would do the sth like
```
// buildedOutputIdentifierFromSpendingConditionArgs can be outputId or utxoPos, depending on spendingTx type
require(buildedOutputIdentifierFromSpendingConditionArgs == spendingTx.inputs[index], "....")
```

SInce `buildedOutputIdFromSpendingConditionArgs` and `buildedOutputIdentifierFromSpendingConditionArgs` are both from same args of spending condition, this makes sure they are indicating same thing. So with the two requirements above, it would represent the same thing as `outputId == spendingTx.inputs[index]`.

kevsul: Ok, thanks for clarifying.
I'll hold off on this until SpendingCondition is done...

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,129 @@
...
+        bytes32 spendingTxInputOutputId = data.controller.isDeposit.test(utxoPos.blockNum())
+                ? OutputId.computeDepositOutputId(data.args.inputTx, utxoPos.outputIndex(), utxoPos.value)
+                : OutputId.computeNormalOutputId(data.args.inputTx, utxoPos.outputIndex());
+
+        require(ifeInputOutputId == spendingTxInputOutputId, "Spent input is not the same as piggybacked input");
+    }
+
+    function verifySpendingCondition(ChallengeIFEData memory data) private view {
+        ISpendingCondition condition = data.controller.spendingConditionRegistry.spendingConditions(
+            data.args.spendingTxInputOutputType, data.args.spendingTxType
```
pgebal: We need to be careful here. I think we should verify that output type and tx type from args match spending tx type and output type of challenged input.

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
```diff
@@ -70,4 +70,27 @@ library PaymentInFlightExitRouterArgs {
...
+     * @param spendingTx RLP encoded spending transaction.
+     * @param spendingTxInputIndex Index of spent input in spending transaction.
+     * @param spendingTxInputOutputType Output type of spent input.
+     * @param spendingTxWitness Witness for spending transaction.
+     */
+    struct ChallengeInputSpentArgs {
+        bytes inFlightTx;
+        uint8 inFlightTxInputIndex;
+        bytes spendingTx;
+        uint8 spendingTxType;
```
pgebal: We can get that from spending tx wire format. No need to provide it as argument.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: Ok. To verify the output type, we need to pass in the outputGuardPreimage and check OutputGuardHandler.isValid(), right?

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,156 @@
...
+        PlasmaFramework framework;
+        IsDeposit.Predicate isDeposit;
+        SpendingConditionRegistry spendingConditionRegistry;
+        OutputGuardHandlerRegistry outputGuardHandlerRegistry;
+    }
+
+    event InFlightExitInputBlocked(
+        address indexed challenger,
+        bytes32 txHash,
+        uint8 inputIndex
```
boolafish: I would use `uint16` for all indexes

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
```diff
@@ -0,0 +1,156 @@
...
+        bytes32 challengingTxInputOutputId = data.controller.isDeposit.test(utxoPos.blockNum())
+                ? OutputId.computeDepositOutputId(data.args.inputTx, utxoPos.outputIndex(), utxoPos.value)
+                : OutputId.computeNormalOutputId(data.args.inputTx, utxoPos.outputIndex());
+
+        require(ifeInputOutputId == challengingTxInputOutputId, "Spent input is not the same as piggybacked input");
+    }
+
+    function verifyOutputType(ChallengeIFEData memory data) private view {
+        UtxoPosLib.UtxoPos memory utxoPos = UtxoPosLib.UtxoPos(data.args.inputUtxoPos);
+        uint16 outputIndex = UtxoPosLib.outputIndex(utxoPos);
```
boolafish:
```suggestion
        uint16 outputIndex = utxoPos.outputIndex();
```

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
```diff
@@ -80,4 +80,29 @@ library PaymentInFlightExitRouterArgs {
...
+     * @param challengingTxInputIndex Index of spent input in challenging transaction.
+     * @param challengingTxInputOutputType Output type of spent input.
+     * @param challengingTxWitness Witness for challenging transaction.
+     * @param inputTx RLP encoded input transaction.
+     * @param inputUtxoPos Utxo position of input transaction's output.
+     * @param spendingConditionOptionalArgs Optional arguments for spending condition of the input transaction.
+     */
+    struct ChallengeInputSpentArgs {
+        bytes inFlightTx;
+        uint8 inFlightTxInputIndex;
```
boolafish:
```suggestion
        uint16 inFlightTxInputIndex;
```

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
```diff
@@ -80,4 +80,29 @@ library PaymentInFlightExitRouterArgs {
...
+        bytes challengingTx;
+        uint8 challengingTxInputIndex;
```
boolafish:
```suggestion
        uint16 challengingTxInputIndex;
```

plasma_framework/test/helpers/ife.js
```diff
@@ -162,6 +159,14 @@ function createInclusionProof(encodedTx, txUtxoPos) {
     };
 }
 
+function getOutputId(txBytes, utxoPos) {
```
boolafish: I would move this to `utils.js` to be with same file as other `computeXXXOutputId`

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
```diff
@@ -0,0 +1,357 @@
...
+        const piggybackInFlightExit = await PaymentPiggybackInFlightExit.new();
+        const challengeInFlightExitNotCanonical = await PaymentChallengeIFENotCanonical.new();
+        const challengeIFEInputSpent = await PaymentChallengeIFEInputSpent.new();
+
+        await PaymentInFlightExitRouter.link('PaymentStartInFlightExit', startInFlightExit.address);
+        await PaymentInFlightExitRouter.link('PaymentPiggybackInFlightExit', piggybackInFlightExit.address);
+        await PaymentInFlightExitRouter.link('PaymentChallengeIFENotCanonical', challengeInFlightExitNotCanonical.address);
+        await PaymentInFlightExitRouter.link('PaymentChallengeIFEInputSpent', challengeIFEInputSpent.address);
+
+        this.exitIdHelper = await ExitId.new();
```
boolafish: Minor: I tends to put this separate with the others. We can have several `before('xxxx', async () => {...})` so we can express them differently. The reason I'd like to separate is that I found out current test code style would need to add links whenever new controller is introduced. As a result, this is the part of code that needs to alter, and it's easier to keep them the same so we can copy-paste easily :P

example code pointer: [here](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnOutput.test.js#L41-L54)

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
```diff
@@ -0,0 +1,357 @@
...
+                utxoPos: buildUtxoPos(3000, 5, outputIndex),
+            };
+        }
+
+        // Sets up an IFE tx using inputTx as an input, starts the IFE and piggybacks on the input.
+        const buildPiggybackInputData = async (inputTx) => {
+            const outputAmount = 997;
+
+            const firstInput = createInputTransaction([buildUtxoPos(BLOCK_NUMBER, 3, 0)], outputOwner, 334455);
+            const firstInputUtxoPos = buildUtxoPos(3000, 66, 0);
```
boolafish: Minor: lol, `firstInput` is using input with utxo that is of block 5000 (BLOCK_NUMBER) while itself is at block number 3000 XDD.

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
```diff
@@ -0,0 +1,357 @@
...
+                exitMap: 0,
+                position: INFLIGHT_EXIT_YOUNGEST_INPUT_POSITION,
+                bondOwner: alice,
+                oldestCompetitorPosition: 0,
+                inputs: [{
+                    outputId: web3.utils.sha3('dummy output id'),
+                    outputGuard: web3.utils.sha3('dummy output guard'),
+                    exitTarget: inputOwner,
+                    token: ETH,
+                    amount: 999,
```
boolafish: Minor: The amount here does not match your input tx amount though. Same on another input amount.

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
```diff
@@ -0,0 +1,357 @@
...
+                    outputGuard: web3.utils.sha3('dummy output guard'),
+                    exitTarget: outputOwner,
+                    token: ETH,
+                    amount: outputAmount,
+                }, emptyWithdrawData, emptyWithdrawData, emptyWithdrawData],
+            };
+
+            const exitId = await this.exitIdHelper.getInFlightExitId(rlpInFlightTxBytes);
+
+            const argsInputOne = {
```
boolafish: I would name this `piggybackArgsInputOne` and same apply for another as piggyback is not the main testing target in this test file. However, I think we can even skip the piggyback part and manipulate the `exitMap` directly in test to mimic the state after piggyback directly IMO, which might/might not?  simpiflies the tests a bit.

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
kevsul: Well this one doesn't have to match because it's an unrelated random input.
The other one should probably match though...

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
kevsul: Good idea, it does simplify the test a bit

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
boolafish: @kevsul I think you missed to handle this?

plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
boolafish: okay, I just have the tendency to make the test data as real as possible unless it's intended to test edge case so that whoever reads the tests won't be confused. But anyway, I am fine with that and it's not a big issue as long as test works.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
kevsul: Ah, I did miss it 🤦‍♂ 
Will do

boolafish:
Leave some notes of solving gas limit via optimizer: 
(btw, there is an issue of using optimizer now: https://github.com/omisego/plasma-contracts/issues/286. I need to comment off the `PaymentStandardExitRouterMock` file)

By turn on optimizer with run=1: (tested by running `PaymentExitGame.test.js`)
```
|  Deployments                                                ·                                       ·  % of limit  ·             │
······························································|·············|·············|···········|··············|··············
|  Erc20DepositVerifier                                       ·          -  ·          -  ·  1166898  ·      17.4 %  ·       4.81  │
······························································|·············|·············|···········|··············|··············
|  Erc20Vault                                                 ·    1237010  ·    1237074  ·  1237070  ·      18.4 %  ·       5.10  │
······························································|·············|·············|···········|··············|··············
|  EthDepositVerifier                                         ·          -  ·          -  ·  1095912  ·      16.3 %  ·       4.52  │
······························································|·············|·············|···········|··············|··············
|  EthVault                                                   ·    1052066  ·    1052130  ·  1052126  ·      15.7 %  ·       4.34  │
······························································|·············|·············|···········|··············|··············
|  OutputGuardHandlerRegistry                                 ·          -  ·          -  ·   275672  ·       4.1 %  ·       1.14  │
······························································|·············|·············|···········|··············|··············
|  PaymentExitGame                                            ·    4607337  ·    4607465  ·  4607431  ·      68.6 %  ·      19.01  │
······························································|·············|·············|···········|··············|··············
|  PaymentOutputGuardHandler                                  ·          -  ·          -  ·   385319  ·       5.7 %  ·       1.59  │
······························································|·············|·············|···········|··············|··············
|  PaymentOutputToPaymentTxCondition                          ·    1830802  ·    1830866  ·  1830862  ·      27.2 %  ·       7.55  │
······························································|·············|·············|···········|··············|··············
|  PaymentProcessInFlightExit                                 ·          -  ·          -  ·  1680281  ·        25 %  ·       6.93  │
······························································|·············|·············|···········|··············|··············
|  PaymentTransactionStateTransitionVerifier                  ·          -  ·          -  ·  1313449  ·      19.5 %  ·       5.42  │
······························································|·············|·············|···········|··············|··············
|  PlasmaFramework                                            ·          -  ·          -  ·  4362865  ·      64.9 %  ·      18.00  │
······························································|·············|·············|···········|··············|··············
|  SpendingConditionRegistry                                  ·          -  ·          -  ·   492772  ·       7.3 %  ·       2.03  │
·-------------------------------------------------------------|-------------|-------------|-----------|--------------|-------------·
```

Optimizer, run=200
```
|  Deployments                                                ·                                         ·  % of limit  ·             │
······························································|·············|·············|·············|··············|··············
|  Erc20DepositVerifier                                       ·          -  ·          -  ·    1185586  ·      17.6 %  ·       4.90  │
······························································|·············|·············|·············|··············|··············
|  Erc20Vault                                                 ·    1289546  ·    1289610  ·    1289606  ·      19.2 %  ·       5.33  │
······························································|·············|·············|·············|··············|··············
|  EthDepositVerifier                                         ·          -  ·          -  ·    1130864  ·      16.8 %  ·       4.67  │
······························································|·············|·············|·············|··············|··············
|  EthVault                                                   ·    1099135  ·    1099199  ·    1099195  ·      16.4 %  ·       4.54  │
······························································|·············|·············|·············|··············|··············
|  OutputGuardHandlerRegistry                                 ·          -  ·          -  ·     284070  ·       4.2 %  ·       1.17  │
······························································|·············|·············|·············|··············|··············
|  PaymentExitGame                                            ·    4738035  ·    4738163  ·    4738129  ·      70.5 %  ·      19.59  │
······························································|·············|·············|·············|··············|··············
|  PaymentOutputGuardHandler                                  ·          -  ·          -  ·     409836  ·       6.1 %  ·       1.69  │
······························································|·············|·············|·············|··············|··············
|  PaymentOutputToPaymentTxCondition                          ·    1890827  ·    1890891  ·    1890887  ·      28.1 %  ·       7.82  │
······························································|·············|·············|·············|··············|··············
|  PaymentProcessInFlightExit                                 ·          -  ·          -  ·    1702057  ·      25.3 %  ·       7.04  │
······························································|·············|·············|·············|··············|··············
|  PaymentTransactionStateTransitionVerifier                  ·          -  ·          -  ·    1343082  ·        20 %  ·       5.55  │
······························································|·············|·············|·············|··············|··············
|  PlasmaFramework                                            ·          -  ·          -  ·    4541165  ·      67.6 %  ·      18.77  │
······························································|·············|·············|·············|··············|··············
|  SpendingConditionRegistry                                  ·          -  ·          -  ·     515503  ·       7.7 %  ·       2.13  │
·-------------------------------------------------------------|-------------|-------------|-------------|--------------|-------------·
```

PR #252, opened 2019-09-06
162 respond to noncanonical challenge

### LINKED ISSUE #162: [PaymentExitGame] respondToNonCanonicalChallenge

pnowosie:
Fixes #162 

```
  ...
    response to non-canonical challenge
      when successfully responded to non-canonical challenge
        ✓ should emit InFlightExitChallengeResponded event (40ms)
        ✓ should set isCanonical back to true (66ms)
        ✓ should set bond owner to caller (59ms)
        ✓ should set oldest competitor position to response position (59ms)
      is unsuccessful and
        ✓ fails when in-flight exit does not exists
        ✓ fails when in-flight transaction is not younger than competitor
        ✓ fails when in-flight transaction is not included in block (44ms)
```

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
@@ -32,7 +32,7 @@ library PaymentChallengeIFENotCanonical {
         uint256 challengeTxPosition
     );
 
-    function run(
+    function challenge(
         Controller memory self,
         PaymentExitDataModel.InFlightExitMap storage inFlightExitMap,
         PaymentInFlightExitRouterArgs.ChallengeCanonicityArgs memory args
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
@@ -102,11 +102,11 @@ library PaymentChallengeIFENotCanonical {
     {
         uint192 exitId = ExitId.getInFlightExitId(inFlightTx);
         PaymentExitDataModel.InFlightExit storage ife = inFlightExitMap.exits[exitId];
-        require(ife.exitStartTimestamp != 0, "In-fligh exit doesn't exists");
+        require(ife.exitStartTimestamp != 0, "In-flight exit doesn't exists");
 
         require(
             ife.oldestCompetitorPosition > inFlightTxPos,
-            "In-flight transaction has to be older than competitors to respond to noncanonical challenge.");
+            "In-flight transaction has to be younger than competitors to respond to non-canonical challenge.");
 
         UtxoPosLib.UtxoPos memory utxoPos = UtxoPosLib.UtxoPos(inFlightTxPos);
         (bytes32 root, ) = self.framework.blocks(utxoPos.blockNum());
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
@@ -26,6 +26,12 @@ library PaymentChallengeIFENotCanonical {
         uint256 challengeTxPosition
     );
 
+    event InFlightExitChallengeResponded(
+        address challenger,
+        bytes32 txHash,
+        uint256 challengeTxPosition
+    );
+
     function run(
         Controller memory self,
         PaymentExitDataModel.InFlightExitMap storage inFlightExitMap,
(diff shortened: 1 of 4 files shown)


plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
```diff
@@ -313,4 +313,138 @@ contract('PaymentInFlightExitRouter', ([_, ifeOwner, inputOwner, outputOwner, co
...
+                this.competingTxBlock.blockTimestamp,
+            );
+
+            await this.exitGame.challengeInFlightExitNotCanonical(
+                this.challengeArgs, { from: challenger },
+            );
+        });
+
+        describe('when successfully responded to non-canonical challenge', () => {
+            beforeEach('include in-flight Tx in a previous block', async () => {
```
pik694: `Tx` - > `tx`

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
```diff
@@ -85,6 +91,35 @@ library PaymentChallengeIFENotCanonical {
...
+    function respond(
+        Controller memory self,
+        PaymentExitDataModel.InFlightExitMap storage inFlightExitMap,
+        bytes memory inFlightTx,
+        uint256 inFlightTxPos,
+        bytes memory inFlightTxInclusionProof
+    )
+        public
+    {
+        uint192 exitId = ExitId.getInFlightExitId(inFlightTx);
```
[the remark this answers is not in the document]
boolafish: > In case when we are using calls to libraries from multiple places, there can happen a situation when part of the code is deployed with one library linked and another part with a different (let's say malicious) version.

I am not fully understanding this comment. The `ExitId` library is currently designed to not need to be linked (as functions are `internal`), so what is the concern here?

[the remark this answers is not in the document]
pik694: Oh, sorry, I was not aware of that. I had in my mind the same model that worked for RootChain

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
```diff
@@ -85,6 +91,35 @@ library PaymentChallengeIFENotCanonical {
...
+        PaymentExitDataModel.InFlightExit storage ife = inFlightExitMap.exits[exitId];
+        require(ife.exitStartTimestamp != 0, "In-flight exit doesn't exists");
+
+        require(
+            ife.oldestCompetitorPosition > inFlightTxPos,
+            "In-flight transaction has to be younger than competitors to respond to non-canonical challenge.");
+
+        UtxoPosLib.UtxoPos memory utxoPos = UtxoPosLib.UtxoPos(inFlightTxPos);
```
boolafish: this line looks weird, why casting a `txPos` to become `utxoPos`.

And I just realized in `RootChain.sol` it is saving `txPos` to the `competitorPosition` and not `utxoPos`. I think we should fix that for the `run` function too. (Also, I'd like to recommend to rename `run` to `challenge` if you'd like to put the two functions in same file)

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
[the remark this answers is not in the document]
pnowosie: As off-site discussion emerged, we decided to create an issue (#253) to rethink remedy for `code duplication` (not related to the :point_up:  original comment)

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
pnowosie: Aaaa :man_facepalming: you're absolutely right :exclamation: Nice catch :exclamation: 
 :+1: rename `run -> challenge`

pnowosie: Eeee, no :man_facepalming: it's totally misleading @boolafish  - client [provides UtxoPos to the contract](https://github.com/omisego/elixir-omg/blob/6df52caa9f7aee079cb9b7f78e63ee8d1d44a80d/apps/omg_watcher/lib/omg_watcher/exit_processor/canonicity.ex#L147) but `RootChain.sol` tricks you it's `txPos`. Oh my ... :man_facepalming:

boolafish: hmmm....but for the case of competitor position, shouldn't it be considering the position of the tx (thus txPos)? because you're competing the priority as a tx not as an output, isn't it?

boolafish: 🤦‍♂.....but yeah, it seems like RootChain.sol is using variable name `txPos` but being a `utxoPos` @@

boolafish: but then you can challenge or respond multiple time with different outputIndexs if we're using utxoPos. As the lower outputIndex can override the one with higher outputIndex number during challenge and respond. Probably nothing would be harmed? However, it seems to me txPos is more logically correct?

pnowosie: Luckily in canonicity game we are interested in oldest competitor so the one with lowest `TxoPos`
so if one challenge with `(blknum, txindex, oindex) = (1000, 222, 999)` you can take his bond easily providing the same data but `TxoPos = (1000, 222, 0)`  which corresponds to the same Transaction position (where output index is ignored)

boolafish: yeah, I agree with that so probably nothing would be harmed. However, this means our client should know that, hey, even we are asking a utxoPos, but please just always give as a txPos + 0, otherwise you might not get your bond......

I don't see this as better design as why we are forcing our client to remember always passing a 0 that should never be used? To me it seems like `RootChain.sol` have the right interface and var naming but implemented wrong instead of the opposite though.

pnowosie: Anyway - I create #254 to avoid/limit spreading the confusion. This is HUGE security thread if we mess with this and use both interchangeably...

boolafish: ah...I know this is merge but I think we should rename the `inFlightTxPos` to `inFlightUTxoPos` to not introduce more confusion. Whether txPos or utxoPos we can leave it to the issue #254.

pnowosie: Yeah, there are more them as `competitorTxPos` and so. I'm leaving it for #254.

PR #257, opened 2019-09-10
Fix failing submit block test

pik694:
Fixes failing tests on master e.g. https://circleci.com/gh/omisego/plasma-contracts/2015

### DIFF
--- a/testlang/testlang.py
+++ b/testlang/testlang.py
@@ -135,7 +135,7 @@ class TestingLanguage:
         blknum = self.root_chain.nextChildBlock()
         block = Block(transactions, number=blknum)
         signed_block = block.sign(signer.key)
-        self.root_chain.functions.submitBlock(signed_block.root).transact({'from': signer.address})
+        self.root_chain.submitBlock(signed_block.root, **{'from': signer.address})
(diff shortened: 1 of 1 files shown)


testlang/testlang.py
```diff
@@ -135,7 +135,7 @@ def submit_block(self, transactions, signer=None, force_invalid=False):
         blknum = self.root_chain.nextChildBlock()
         block = Block(transactions, number=blknum)
         signed_block = block.sign(signer.key)
-        self.root_chain.functions.submitBlock(signed_block.root).transact({'from': signer.address})
+        self.root_chain.functions.submitBlock(signed_block.root).transact({'from': signer.address, 'gas': DEFAULT_GAS})
```
boolafish: so this one does not goes into the Convenience wrapper that would put a gas cost automatically?

pik694: In this case we are bypassing the wrapper, but I do not really remember why we do it. Let me check.

pik694: It seems that there is no reason to bypass the wrapper in here. It must have been left during refactors. I have applied the change to use the wrapper ;)

PR #258, opened 2019-09-11
Set number of allowed solhint warnings to 0

thec00n:


boolafish:
Minor one: we've been following this commit message style

https://omisego.atlassian.net/wiki/spaces/ENG/pages/63111218/Commit+Message+Conventions+OMG+Network

PR #261, opened 2019-09-11
feat: emit events for sucessful Ethereum and ERC20 deposits

### LINKED ISSUE #260: Deposit in ALD does not emit an event
## Current Behavior
When deposit to the vault is made, no event is emitted, hence child chain application does not know about deposits.

## Expected Behavior
When deposit is created a proper event is emitted.

## Proposed Solution
Each vault emits and event for a successful deposit.

pgebal:
Closes #260

As plasma framework contract does not understand strucutre of deposit events are emitted by specific vaults.

### DIFF
--- a/plasma_framework/contracts/src/vaults/Erc20Vault.sol
+++ b/plasma_framework/contracts/src/vaults/Erc20Vault.sol
@@ -16,6 +16,13 @@ contract Erc20Vault is Vault {
         uint256 amount
     );
 
+    event DepositCreated(
+        address indexed depositor,
+        uint256 indexed blknum,
+        address indexed token,
+        uint256 amount
+    );
+
     constructor(PlasmaFramework _framework) public Vault(_framework) {}
 
     /**
@@ -28,7 +35,9 @@ contract Erc20Vault is Vault {
 
         IERC20(token).safeTransferFrom(owner, address(this), amount);
 
-        super._submitDepositBlock(_depositTx);
+        uint256 blknum = super._submitDepositBlock(_depositTx);
+
+        emit DepositCreated(msg.sender, blknum, token, amount);
     }
 
     /**
--- a/plasma_framework/contracts/src/vaults/EthVault.sol
+++ b/plasma_framework/contracts/src/vaults/EthVault.sol
@@ -10,6 +10,13 @@ contract EthVault is Vault {
         uint256 amount
     );
 
+    event DepositCreated(
+        address indexed depositor,
+        uint256 indexed blknum,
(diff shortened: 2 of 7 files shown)


plasma_framework/contracts/src/vaults/EthVault.sol
```diff
@@ -10,6 +10,13 @@ contract EthVault is Vault {
         uint256 amount
     );
 
+    event DepositCreated(
+        address indexed depositor,
+        uint256 indexed blknum,
+        address indexed token,
```
boolafish: hmmm, do we really need this? I think the client would be forced to listen event from different contracts anyway, keeping the interface for both vault might not bring much value?

plasma_framework/contracts/src/vaults/Erc20Vault.sol
```diff
@@ -16,6 +16,13 @@ contract Erc20Vault is Vault {
         uint256 amount
     );
 
+    event DepositCreated(
```
paulperegud: Why not just "Deposit"?

plasma_framework/contracts/src/vaults/EthVault.sol
```diff
@@ -18,8 +25,9 @@ contract EthVault is Vault {
      */
     function deposit(bytes calldata _depositTx) external payable {
         IEthDepositVerifier(getEffectiveDepositVerifier()).verify(_depositTx, msg.value, msg.sender);
+        uint256 blknum = super._submitDepositBlock(_depositTx);
 
-        super._submitDepositBlock(_depositTx);
+        emit DepositCreated(msg.sender, blknum, address(0), msg.value);
```
pik694: I would also try to move emitting this event in `_submitDepositBlock`

plasma_framework/contracts/src/vaults/Erc20Vault.sol
pik694: This `DepositCreated` is a legacy from `RootChain`, but I do agree that event `Deposit` or `Deposited` would be probably better.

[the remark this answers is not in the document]
boolafish: no this is wrong direction, at least current interface would fail for ERC721. Unless the event is emitting with more general data like `txBytes`.

plasma_framework/contracts/src/vaults/EthVault.sol
pgebal: I did it to be consistent with previous version of the contracts.

PR #276, opened 2019-09-17
Makes metadata field mandatory for rlp-encoded tx

### LINKED ISSUE #189: Structural hash allows collisions
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior

We can easily generate 2 transaction with different identifiers (`keccak(txbytes)`) which both has equal structural hash (EIP-712) which signature schema is using.

At first it seems to be just bug. However as @pdobacz pointed out https://github.com/omisego/elixir-omg/issues/827#issuecomment-515338201 it has more serious security implications.
Including
* prove non-canonicality of the transaction
* challenge input piggybacking in IFE which prevents funds exit

## Expected Behavior

Transactions with different identifiers **MUST HAVE** different hashes used in signing

## Steps to Reproduce

1. Create transaction without `metadata`
2. Create :point_up: corresponding transaction with `metadata = 0` (32-zero bytes) 
3. These transactions differs on identifier and matches on struct hash

- Full output of error:
- Command that caused error:
- Code that caused error:
Limitation in EIP-712 implementation. 32-zero bytes are used for transaction's metadata when no metadata is contained

## Suggested Fix

EIP-712 domain should specify 2 types of transactions 
* with metadata
* without metadata

Structural hash calculation should identify which type of transaction is provided.

**NOTE:** We can also make `metadata` field mandatory

## Motivation for Change

Security issue

## System Specs

Solidy 0.5
Elixir code changes tracked in [elixir-omg#827](https://github.com/omisego/elixir-omg/issues/827)

pnowosie:
Fixes #189 

This change requires that rlp-encoded transaction contains 4 fields: type, inputs, outputs & metadata.
Metadata is decoded to `bytes32`.

## Testing
By an accident we always passing 32-zero bytes as [metadata from truffle tests](https://github.com/omisego/plasma-contracts/blob/ae4545c504ce7ca680d26cbccd36e575d66c59a7/plasma_framework/test/helpers/transaction.js#L30) so no changes requires in tests.

### DIFF
--- a/plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
+++ b/plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
@@ -13,7 +13,6 @@ library PaymentTransactionModel {
     uint8 constant public MAX_OUTPUT_NUM = 4;
 
     uint8 constant private ENCODED_LENGTH_WITH_METADATA = 4;
-    uint8 constant private ENCODED_LENGTH_WITHOUT_METADATA = 3;
 
     struct Transaction {
         uint256 txType;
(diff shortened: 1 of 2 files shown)


boolafish:
Oh then shall we test that it would fail if without metadata?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -13,7 +13,6 @@ library PaymentTransactionModel {
     uint8 constant public MAX_OUTPUT_NUM = 4;
 
     uint8 constant private ENCODED_LENGTH_WITH_METADATA = 4;
```
pik694: Since there no longer is `ENCODED_LENGTH_WITHOUT_METADATA`, you can simple rename this constant into `ENCODED_LENGTH`

Issue #273, opened 2019-09-17
Rename `token` in PlasmaFramework


boolafish:
### Why
We are using the word `token` in our framework layer. The main reason we need that concept in the framework layer is that we need to isolate the exit queue for each external contracts outside the Plasma system.

`token` makes sense for ERC20, however, as the `PlasmaFramework` is designed to be able to support multiple ERCs, then it does not represent that well for others. For instance, in ERC721 (non-fungible token), when people think of `token` it might link to the `uid` of the token instead of the ERC721 implementation contract address.

### Proposal
Rename `token` to `ercConctract` in PlasmaFramework. (Or if there is any better word choice)

Some code pointer: [here](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/framework/ExitGameController.sol#L49)

boolafish:
we end up decided to keep the name `token` as there is no general consensus on the naming and `token` is used in etherscan for like ERC721 too.

PR: https://github.com/omisego/plasma-contracts/pull/285
Discussion: https://git.io/Je3qD

PR #277, opened 2019-09-18
feat: rename 'token' to 'ercContract' in framework

boolafish:
### Note
We are renaming 'token' to 'ercContract' as the word 'token' might not be appropriate when it is
not using for ERC20 tokens on `PlasmaFramework`.

closes issue: https://github.com/omisego/plasma-contracts/issues/273

(PS. if anyone has better naming idea please let me know)

### DIFF
--- a/plasma_framework/contracts/mocks/exits/SpyPlasmaFrameworkForExitGame.sol
+++ b/plasma_framework/contracts/mocks/exits/SpyPlasmaFrameworkForExitGame.sol
@@ -10,7 +10,7 @@ contract SpyPlasmaFrameworkForExitGame is PlasmaFramework {
     mapping (uint256 => BlockModel.Block) public blocks;
 
     event EnqueueTriggered(
-        address token,
+        address ercContract,
         uint64 exitableAt,
         uint256 txPos,
         uint256 exitId,
@@ -24,13 +24,13 @@ contract SpyPlasmaFrameworkForExitGame is PlasmaFramework {
     }
 
     /** override for test */
-    function enqueue(address _token, uint64 _exitableAt, TxPosLib.TxPos calldata _txPos, uint192 _exitId, IExitProcessor _exitProcessor)
+    function enqueue(address _ercContract, uint64 _exitableAt, TxPosLib.TxPos calldata _txPos, uint192 _exitId, IExitProcessor _exitProcessor)
         external
         returns (uint256)
     {
         enqueuedCount += 1;
         emit EnqueueTriggered(
-            _token,
+            _ercContract,
             _exitableAt,
             _txPos.value,
             _exitId,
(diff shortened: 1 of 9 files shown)


pdobacz:
> (PS. if anyone has better naming idea please let me know)

I don't, but `ercContract` is bothering me too :(. I was thinking `currency` - which is what we use interchangeably (sigh) in code. Maybe a native English speaker could weigh in on how bad it sounds in the context of an ERC721 contract.

But please, consider keeping it to "Token", c.f. https://etherscan.io/token/0x06012c8cf97bead5deae237070f9587f8e7a266d and https://etherscan.io/tokens-nft

they're both using "token" in the context of the ERC721 contract, and "tokenID" for individual items. I think it is perfectly acceptable even if not strictly correct.

BTW, Is there input from anyone who actually was confused by use of the term "token" or are we just future proofing here?

boolafish:
> BTW, Is there input from anyone who actually was confused by use of the term "token" or are we just future proofing here?

Just future proofing I guess. 

> But please, consider keeping it to "Token", c.f. https://etherscan.io/token/0x06012c8cf97bead5deae237070f9587f8e7a266d and https://etherscan.io/tokens-nft

Oh nice, I was keep wondering how people handle ERC721 contract. I wasn't able to find the word from the ERC implementation.

If they are others using `token`, then sure I don't have strong opinion. I think something from @paulperegud 's recent writing might be a good one: `token_contract_address` so we don't confuse with what it really means. How do you think @pdobacz ?


Btw, this issue/PR is only targeting the renaming on framework layer as `token` still makes perfect sense to the `PaymentTx` and it's clear and simple enough there so I am not proposing to change that. It is just on the framework layer it would be confusing as it is really the ERC contract address and the reason of having that in framework is really just to isolate the exit queue for external contracts.

pdobacz:
`token_contract_address` is quite long and still can be just as confusing, if someone isn't fully aware of how ERC721 is constructed ("since it's called token_contract_address, do individual items have their contract addresses, huh?")

I don't have a strong opinion either, but the pragmatic side of things is:
 - let's be lax on this
 - let's be good to ourselves and avoid long names for things (we'll probably still be saying "token" in our conversations...)
 - let's not change stuff that settled in, if there's no strong reasons to do so ;)
 - etherscan is kinda source of truth and strong reference, consistency with that will help.

But no strong opinion here too, in the end. Maybe let's not bikeshed and just toss a coin :)

boolafish:
okay, in general I am convinced that `token` might be able to be enough for a good enough while. Though personally I do hope to make it something like `tokenContract` at least so we'd know it's really talking about the external contract from framework's point of view.

btw, I haven't dig into this deep yet but there do is sth like this: https://blog.enjincoin.io/erc-1155-the-final-token-standard-on-ethereum-a83fce9f5714 which has multiple token protocol under one contract (ouch....). Anyway, we might never tough that anyway XD

boolafish:
close this in flavor of #285 and stick with name `token`.

PR #278, opened 2019-09-18
274 use output index in state trasition verifier instead

### LINKED ISSUE #274: Use `outputIndexOfInputTxs` instead of `inputUtxosPos` for `IStateTransitionVerifier`
### What
Current `IStateTransitionVerifier` interfaces requires the `utxoPos` as input. However, it seems like what it really needs are the `outputIndex` for `inputTx` so the contract can know which output would be used as input.

https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/interfaces/IStateTransitionVerifier.sol#L12

### Why the change
Using `utxoPos` would confuse reader whether a more global plasma status is checked inside or not. It is not designed to be checked there so removal of unnecessary info would make the intention more clear (IMHO).

boolafish:
### Notes
- `utxoPos` --> `outputIndex` in StateTransitionVerifier: closes #274 
- use `StateTransitionVerifierMock` and replace the existing `StateTransitionVerifierAccept/Reject/Reverse` to simplify tests
- add tests in startIFE testing the expected args is called with the stateTransitionVerifier

### Test
```
    startInFlightExit
      when calling start in-flight exit succeed with valid arguments
        ✓ should call the StateTransitionVerifier with correct arguments (745ms)
```


plasma_framework/contracts/mocks/exits/StateTransitionVerifierMock.sol
```diff
@@ -0,0 +1,66 @@
...
+    bool public shouldRevert;
+    Args public expectedArgs;
+
+    struct Args {
+        bytes inFlightTx;
+        bytes[] inputTxs;
+        uint16[] outputIndexOfInputTxs;
+    }
+
+    /** mock what would "isCorrectStateTransition()" returns */
```
pgebal: We can remove 'would'.

plasma_framework/contracts/mocks/exits/StateTransitionVerifierMock.sol
```diff
@@ -0,0 +1,66 @@
...
+    function mockResult(bool result) public {
+        expectedResult = result;
+    }
+
+    /** when called, the "isCorrectStateTransition" function would always revert on purpose */
```
pgebal: would always revert -> reverts

plasma_framework/contracts/mocks/exits/StateTransitionVerifierMock.sol
```diff
@@ -0,0 +1,66 @@
...
+        // only run the check when "shouldVerifyArgumentEqauals" is called
+        if (expectedArgs.inFlightTx.length > 0) {
+            require(keccak256(inFlightTx) == keccak256(expectedArgs.inFlightTx), "in-flight tx is not as expected");
+
+            require(inputTxs.length == expectedArgs.inputTxs.length, "input txs array length mismatch expected data");
+            for (uint i = 0; i < expectedArgs.inputTxs.length; i++) {
+                require(keccak256(inputTxs[i]) == keccak256(expectedArgs.inputTxs[i]), "input tx is not as expected");
+            }
+
+            require(outputIndexOfInputTxs.length == expectedArgs.outputIndexOfInputTxs.length, "outputIndex array length mismatch expected data");
```
pgebal: mismatch -> mismatches

plasma_framework/contracts/src/exits/interfaces/IStateTransitionVerifier.sol
```diff
@@ -9,7 +9,7 @@ interface IStateTransitionVerifier {
     function isCorrectStateTransition(
         bytes calldata inFlightTx,
         bytes[] calldata inputTxs,
-        uint256[] calldata inputUtxosPos
+        uint16[] calldata outputIndexOfInputTxs
```
pgebal: Could you add a doc for this function? `@param`

plasma_framework/test/src/exits/payment/PaymentStartInFlightExit.test.js
```diff
@@ -167,6 +166,22 @@ contract('PaymentInFlightExitRouter', ([_, alice, richFather, carol]) => {
...
             });
 
+            it('should call the StateTransitionVerifier with correct arguments', async () => {
+                const expectedArgs = {
+                    inFlightTx: this.args.inFlightTx,
+                    inputTxs: this.args.inputTxs,
+                    outputIndexOfInputTxs: this.args.inputUtxosPos.map(utxo => new UtxoPos(utxo).outputIndex),
+                };
+
+                // test would fail if called but not with the expected arguments
```
pgebal: if called with unexpected arguments

boolafish:
close this in flavor of #285

plasma_framework/contracts/src/exits/interfaces/IStateTransitionVerifier.sol
boolafish: okay, I add the doc as following:
I renamed `inFlightTx` to `txBytes` as from this verifier's point of view, it does not care that transaction is in-flight exiting or not.
```
/**
    * @notice Verifies state transition logic.
    * @param txBytes the targeting transaction that is doing the state transition verification
    * @param inputTxs input transactions of the targeting transaction checking state transition
    * @param outputIndexOfInputTxs the output index of the input txs that the transaction input is pointing to
    */
    function isCorrectStateTransition(
        bytes calldata txBytes,
        bytes[] calldata inputTxs,
        uint16[] calldata outputIndexOfInputTxs
    )
        external
        view
        returns (bool);
```

PR #281, opened 2019-09-18
fix: restrict exit game processExit calls to only be called from exit game controller

### LINKED ISSUE #271: IExitProcessor.processExits implementations should not allow anyone in (plasma framework/ALD)
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior

Anyone can call https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/PaymentExitGame.sol#L24 now. This is expected to be called from within `PlasmaFramework.processExits`, which cares about the priority queues.

## Expected Behavior

Only the allowed contract should call. Also suggestion: rename `IExitProcessor.processExits` to something different, indicating this is should be protected + provide more documentation with expectations and behavior of the abstract function, for the sake of future implementation authors.

## Steps to Reproduce

N/A

pgebal:
Closes #271

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEOutputSpent.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEOutputSpent.sol
@@ -69,7 +69,7 @@ library PaymentChallengeIFEOutputSpent {
 
         verifyInFlightTransactionStandardFinalized(controller, args);
         verifyOutputType(controller, args);
-        verifyChallengingTransactionSpendsOutput(controller, args, ife);
+        verifyChallengingTransactionSpendsOutput(controller, args);
 
         ife.clearOutputPiggybacked(outputIndex);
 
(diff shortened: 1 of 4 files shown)


plasma_framework/contracts/src/exits/OnlyFromExitGameController.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.5.0;
```
pik694: Instead of creating an abstract contract with just a modifier and one field and call them respectively `OnlyFrom...`, do not we want to create a one abstract contract `OnlyFrom` which would not have any fields, but just a single modifier that would take a parameter: 
``` modifier onlyFrom(address caller){...}``` 

In this case we can just call `onlyFrom(exitGameController)` and there is so much less code boilerplate

File: plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFEOutputSpent.sol [hunk 2]
```diff
@@ -69,7 +69,7 @@ library PaymentChallengeIFEOutputSpent {
 
         verifyInFlightTransactionStandardFinalized(controller, args);
         verifyOutputType(controller, args);
-        verifyChallengingTransactionSpendsOutput(controller, args, ife);
+        verifyChallengingTransactionSpendsOutput(controller, args);
```
pik694: Are you sure that this should me a part of this PR?

I believe I approved those changes earlier today.

pgebal: Solved by a merge.

Issue #286, opened 2019-09-19
Fix ALD code non-able to compile with optimizer on


boolafish:
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior
`plasma_framework/`  code cannot compile with optimizer on.

## Expected Behavior
Should be able to compile.

## Steps to Reproduce
in `truffle-config.js`, turn on the optimizer:
```
          settings: {          // See the solidity docs for advice about optimization and evmVersion
                optimizer: {
                    enabled: true,
                    runs: 1,
                },
                evmVersion: "byzantium",
            },
```
And then run: `truffle compile`.
It would throw the error:
```
InternalCompilerError: Assembly exception for bytecode

Compilation failed. See above.
Truffle v5.0.21 (core: 5.0.21)
Node v9.2.1
```

## Suggested Fix
So far, I wasn't able to find a fix yet. However, I can pin-point the file that causes the error is `PaymentStandardExitRouterMock` (by deleting file, compile and passes). Luckily it's just the mock that causes this.

boolafish:
[Note]

So if we remove the part
```
      PaymentStandardExitRouter(
            plasmaFramework,
            ethVault,
            erc20Vault,
            outputGuardHandlerRegistry,
            spendingConditionRegistry
        )
```
in constructor it would compile.....however I don't know why and how to solve yet XDD

pdobacz:
Hi, not sure if this helps, but with `solc v0.5.11` and contracts from commit `67b90e5175cd1e4d503ad0160cb069565ef91136`, branch: `150_python_tests_on_plasma_framework` it compiles fine with optimizer on, 1 run:

My invocation `"~/Downloads/solc-static-linux openzeppelin-solidity=openzeppelin_solidity --overwrite --abi --bin --optimize --optimize-runs 1 -o ...`

pdobacz:
oh, I just after noticed that it's a single file that fails, which I am not compiling :man_facepalming:, oh well, so scratch my comment

PR #291, opened 2019-09-20
269 migration script

boolafish:
Currently a draft to show what's the ordering of deployment and registration would be like.

One issue I see when writing this migration:
The contract is designed to re-use in the future. As a result, we might need to deploy one contract with potentially different args multiple times in the future. However, truffle migration using  `deployer.deployed(ContractName)` might not be able to differentiate multiple deployment. https://github.com/trufflesuite/truffle/issues/237

Potentially we can probably use `new` for the contracts that has the possibility to re-deploy?


boolafish:
@kendricktan here's the draft! BTW, current contract would exceed normal gas size so I use 8000000 for now. If you'd like to try, please initiate instance with higher gas.

cc @thec00n as you might be interested in migration script.

thec00n:
@boolafish @kendricktan Could we consider the following as part of this PR:

- Fix the version of `openzeppelin-solidity` to 2.3.0 and manually upgrade. The contracts are injected in various parts of the contract system. It is safer to chose a manual upgrade path in this case. 
- Fix the version of the Solidity compiler 0.5.11. Consider to fix the pragma version in the contracts as well. 
- Decide if the contracts are deployed with optimizations. If so let's set the option by default and make sure that all tests run in an environment with optimizations set.

kendricktan:
@thec00n 

- (Y) How are you injecting `openzepplin-solidity`? From the looks of it it just seems like you're loading it from `node_modules`
- (Y)
- (Y) It'll need to be run with optimizations otherwise it'll keep throwing the `gas required exceeds allowance` error

thec00n:
> @thec00n
> 
> * 👍  How are you injecting `openzepplin-solidity`? From the looks of it it just seems like you're loading it from `node_modules`

I think truffle takes care of looking in the right place when it compiles the contracts.

kendricktan:
@thec00n even with optimizations, the gas limit still exceeds the allowance. Any ideas? :\

thec00n:
@boolafish or @kevsul might know?

pdobacz:
@kendricktan Sorry to jump in, but maybe this will help:

> even with optimizations, the gas limit still exceeds the allowance. Any ideas? :\

If we're talking about exceeding the allowance on contract deployment, try optimizations on but with 1 run. This works for me (at least for a subset of the required deployments).

1 run optimizes for compactness of the bytecode, thereby minimizing the deployment gas cost

kevsul:
If you switch on optimisations, be aware of this issue: https://github.com/omisego/plasma-contracts/issues/286
Deleting `PaymentStandardExitRouterMock.sol` work around it for now.

With that I've been able to deploy to Rinkeby. I've had to add an explicit `{ gas: 6500000 }` when deploying `PaymentExitGame` as well.

boolafish:
Do we really need to base on the `150_python_tests_on_plasma_framework` branch? BTW, why elixir uses that branch instead of master @pdobacz ? Also, if it's necessary to base on that branch, can @pik694 find a point that we can merge the PR first? and we keep on migrate other python codes with other PRs so we don't block ourselves or doing code review cross PRs.

kasima:
Agree with @boolafish here. If we need to wait for that branch, let's merge it first so things don't get too messy.

kendricktan:
@kasima @boolafish so should I cull the merges and restart migrations from master in this case?

boolafish:
hmmm probably, if it's not too hard to move back to master (?)
Also, I am still not following why we need to do it on top of the python test branch

pik694:
I will rebase today and try to prepare a PR :)

pdobacz:
> BTW, why elixir uses that branch instead of master @pdobacz ?

b/c I'm integrating elixir to what I know works with the old python tests, so I know how things might break and what part of contract functionality I can rely on.

+1 on merging what's mergable to `plasma-contracts/master` in general and moving on.

kasima:
Thanks for untangling these dependencies, everyone. 🙌

PR #293, opened 2019-09-20
fix: use bondsize lib for ifes

kevsul:


plasma_framework/test/src/exits/payment/PaymentChallengeIFEInputSpent.test.js
```diff
@@ -184,17 +182,20 @@ contract('PaymentChallengeIFEInputSpent', ([_, alice, inputOwner, outputOwner, c
             // Create the input tx
             const inputTx = buildInputTx();
 
+            await this.framework.setBlock(BLOCK_NUMBER, web3.utils.sha3('dummy root'), 0);
+
+            this.startIFEBondSize = await this.exitGame.startIFEBondSize();
```
boolafish: I would probably not change the tests for `challenge` (IFEInputSpent, IFEOutputSpent, IFENotCanonical) because this can be misleading actually. It would be reading from storage of exit anyway and not dependent on what current bond size is, especially current bond size might change afterward, and potentially there can be piggybacks of same exit with different bond sizes.

PR #294, opened 2019-09-21
Remove misleading node files, set node version

pnowosie:
If I'm correct we're using truffle just for plasma_framework, so files in root dir are misleading and useless.
Also we're missing default node version, hence I'm setting it safely to `10.14`.


boolafish:
hmmm.... the outside layer of truffle is used for deployment actually

kevsul:
☝️  That's right - truffle is used to deploy the current RootChain.sol and related contracts

boolafish:
I think we can close this. If I am wrong please re-open this @pnowosie

PR #297, opened 2019-09-24
chore: turn on optimizer + reenable e2e tests

### LINKED ISSUE #286: Fix ALD code non-able to compile with optimizer on

boolafish:
### Note
Previously the e2e tests cannot be run due to gas limit on contract deployment. With optimizer
turned on, it can be deployed within gas limit.

Also, with latest compiler (0.5.11) it can compile without the previous issue tested in 0.5.8

closes #286

### DIFF
--- a/plasma_framework/truffle-config.js
+++ b/plasma_framework/truffle-config.js
@@ -92,15 +92,14 @@ module.exports = {
     // Configure your compilers
     compilers: {
         solc: {
-            // version: "0.5.1",    // Fetch exact version from solc-bin (default: truffle's version)
-            // docker: true,        // Use "0.5.1" you've installed locally with docker (default: false)
-            // settings: {          // See the solidity docs for advice about optimization and evmVersion
-            //  optimizer: {
-            //    enabled: false,
-            //    runs: 200
-            //  },
-            //  evmVersion: "byzantium"
-            // }
+            version: '0.5.11',
+            settings: {
+                optimizer: {
+                    enabled: true,
+                    runs: 200,
+                },
+                evmVersion: 'byzantium',
+            },
         },
     },
 };
--- a/plasma_framework/truffle-config.js
+++ b/plasma_framework/truffle-config.js
@@ -98,7 +98,6 @@ module.exports = {
                     enabled: true,
                     runs: 200,
                 },
-                evmVersion: 'byzantium',
             },
         },
     },
--- a/plasma_framework/contracts/mocks/Imports.sol
+++ b/plasma_framework/contracts/mocks/Imports.sol
@@ -3,3 +3,7 @@ pragma solidity ^0.5.0;
 // Import contracts from third party Solidity libraries to make them available in tests.
 
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20Mintable.sol";
+
+contract Import {
+    // dummy empty contract to allow the compiler not always trying to re-compile this file.
+}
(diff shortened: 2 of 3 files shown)


kevsul:
Awesome! But where are you setting the compiler to 0.5.11?

plasma_framework/truffle-config.js
```diff
@@ -92,15 +92,14 @@ module.exports = {
...
-            //    enabled: false,
-            //    runs: 200
-            //  },
-            //  evmVersion: "byzantium"
-            // }
+            version: '0.5.11',
+            settings: {
+                optimizer: {
+                    enabled: true,
+                    runs: 200,
```
kevsul: Should we set `runs: 1` to optimise for size?

plasma_framework/truffle-config.js
```diff
@@ -92,15 +92,14 @@ module.exports = {
...
-            // version: "0.5.1",    // Fetch exact version from solc-bin (default: truffle's version)
-            // docker: true,        // Use "0.5.1" you've installed locally with docker (default: false)
-            // settings: {          // See the solidity docs for advice about optimization and evmVersion
-            //  optimizer: {
-            //    enabled: false,
-            //    runs: 200
-            //  },
-            //  evmVersion: "byzantium"
-            // }
+            version: '0.5.11',
```
boolafish: > t where are you setting the compiler to 0.5.11

Here :)

plasma_framework/truffle-config.js
boolafish: hmm, I don't know which is better. But if it is deploy-able, do we need to keep optimize for deploy cost and not optimize for transaction?

plasma_framework/truffle-config.js
kevsul: 🤦‍♂

plasma_framework/truffle-config.js
kevsul: I'm surprised it's deployable! :)
But yeah, let's leave it like this and change it later if necessary

plasma_framework/truffle-config.js
```diff
@@ -92,15 +92,14 @@ module.exports = {
...
+            settings: {
+                optimizer: {
+                    enabled: true,
+                    runs: 200,
+                },
+                evmVersion: 'byzantium',
```
kevsul: Why 'byzantium'? The latest is 'petersburg'  (which is the default).

boolafish: ah....I just uncomment it and thought that's the latest @@. I think I would just remove then, unless there's reason to explicitly state that.

Issue #254, opened 2019-09-06
`TxPosLib` library HAS TO be removed !!!


pnowosie:
If I understand `RootChain.sol` correctly there never be a mention of `Transaction position` used the way `TxPosLib` computes. Everywhere `TxoPos == Transaction Output position` was used even there where `oindex` was ignored or assumed to be zero.

There is only one place where RootChain's `getTxPos` was used and it's `getStandardExitPriority` where a lot information is packed into bignum with bit operations :face_with_head_bandage: 

Therefore `TxPosLib` shouldn't be exposed publicitly as it makes confusion and misleads to never used concept.  Also we would want to rename all confusing `TxPos` to `TxoPos` - to make them more clear. 

Also there might be a good idea to make broader use of UtxoPosLib.UtxoPos structure in the code.
So external functions would accept uint256 utxoPos value, which will be converted to the :point_up: UtxoPos struct and passed in that form only. Eg. also to the merkle Lib and so... 

How about ExitData stores structure instead of simple uint256 :question:

boolafish:
I actually think it was implemented wrong, at least not as intentional in `RootChain.sol` though.

See my comments in the PR: https://github.com/omisego/plasma-contracts/pull/252#discussion_r321729009

pnowosie:
> Also, whenever we are doing merkle proof check, we are only checking the `blockNum + txIndex` is correct (that is actually the info of `txPos`), as `outputIndex` is not checked there.

Obviously one doesn't need `oindex` value to precisely knew Tx position which is used for merkle proof. However if we allow contract to accept UtxoPos in few places and TxPos in some other places we're asking that someone will finally compare to uint256 number to check which position is older - whatever huge TxPos will be it will be always smaller than UtxoPos.

boolafish:
As a successful criteria of this issue, we need to solve:
1. Whether we'd like to pass in `txPos` or `utxoPos` in canonicity challenge
1. make sure we are consistent in var names in code, see comment: https://github.com/omisego/plasma-contracts/pull/252#discussion_r321783002

PR #303, opened 2019-09-25
283 Return to old exit priority schema

### LINKED ISSUE #283: Return to exit priority and exit id calculation schema from RootChain
We should return to the exit priority calculation schema that was used in RootChain contract, as we should favor the solutions that do not change the API,  there is also no apparent reason to stay with the current schema, moreover the current schema costs more gas.

### RootChain schema
Exit priority used to be computed with the following schema:
```
|255                     214|213     160|159                      0|
|---------------------------|-----------|--------------------------|
|    exitable timestamp     | tx pos    |         exit id          |
```

### Current schema
```
|255                     192|191              64|63    0|
|---------------------------|-------------------|-------|
|    exitable timestamp     |    tx position    | nonce |
```

The changes of the exit priority does not affect the priority itself, as the exitable timestamp is still the most significant field, but such the new schema changes the API (the old priority differs from the new one givent the same transaction) and the current schema enforces to introduce a mapping from exit priority to exit id, in order to process the exit.   

The field nonce is stored in the `ExitGameController` contract and is updated each enqueuing - which increases the overall cost of `startExit` transaction.

# priority -> Exit mapping in ExitGameController
Having changed the exit priority, we can extract exit id directly from the priority, therefore the `mapping (uint256 => Exit) public exits;` can be changed as follows: `mapping (uint256 => IExitProcessor) public exits`, which is also more gas efficient.

pik694:
Closes #283. Please look to the issue for the explanation of the changes.

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -8,15 +8,9 @@ import "./utils/ExitPriority.sol";
 import "../utils/TxPosLib.sol";
 
 contract ExitGameController is ExitGameRegistry {
-    uint64 public exitQueueNonce = 1;
-    mapping (uint256 => Exit) public exits;
+    mapping (uint256 => IExitProcessor) public delegations;
     mapping (address => PriorityQueue) public exitsQueues;
     mapping (bytes32 => bool) public isOutputSpent;
-
-    struct Exit {
-        IExitProcessor exitProcessor;
-        uint192 exitId; // The id for exit processor to identify specific exit within an exit game.
-    }
 
     event TokenAdded(
         address token
@@ -72,64 +66,58 @@ contract ExitGameController is ExitGameRegistry {
      * @param _exitProcessor The exit processor contract that would be called during "processExits"
      * @return a unique priority number computed for the exit
      */
-    function enqueue(address _token, uint64 _exitableAt, TxPosLib.TxPos calldata _txPos, uint192 _exitId, IExitProcessor _exitProcessor)
+    function enqueue(address _token, uint64 _exitableAt, TxPosLib.TxPos calldata _txPos, uint160 _exitId, IExitProcessor _exitProcessor)
         external
         onlyFromNonQuarantinedExitGame
         returns (uint256)
     {
         require(hasToken(_token), "Such token has not been added to the plasma framework yet");
-
         PriorityQueue queue = exitsQueues[_token];
 
-        uint256 uniquePriority = ExitPriority.computePriority(_exitableAt, _txPos, exitQueueNonce);
-        exitQueueNonce++;
-
+        uint256 uniquePriority = ExitPriority.computePriority(_exitableAt, _txPos, _exitId);
+       
         queue.insert(uniquePriority);
-
-        exits[uniquePriority] = Exit({
-            exitProcessor: _exitProcessor,
-            exitId: _exitId
-        });
+        delegations[uniquePriority] = _exitProcessor;
 
         emit ExitQueued(_exitId, uniquePriority);
-
         return uniquePriority;
     }
 
     /**
      * @notice Processes any exits that have completed the challenge period.
      * @param _token Token type to process.
-     * @param _topUniquePriority Unique priority of the first exit that should be processed. Set to zero to skip the check.
+     * @param _topExitId Unique priority of the first exit that should be processed. Set to zero to skip the check.
      * @param _maxExitsToProcess Maximal number of exits to process.
      * @return total number of processed exits
      */
-    function processExits(address _token, uint256 _topUniquePriority, uint256 _maxExitsToProcess) external {
+    function processExits(address _token, uint160 _topExitId, uint256 _maxExitsToProcess) external {
         require(hasToken(_token), "Such token has not been added to the plasma framework yet");
 
         PriorityQueue queue = exitsQueues[_token];
         require(queue.currentSize() > 0, "Exit queue is empty");
 
         uint256 uniquePriority = queue.getMin();
-        require(_topUniquePriority == 0 || uniquePriority == _topUniquePriority,
-            "Top unique priority of the queue is not the same as the specified one");
+        uint160 exitId = ExitPriority.parseExitId(uniquePriority);
+        require(_topExitId == 0 || exitId == _topExitId,
+            "Top exit id of the queue is not the same as the specified one");
 
-        Exit memory exit = exits[uniquePriority];
+        IExitProcessor processor = delegations[uniquePriority];
         uint256 processedNum = 0;
 
         while (processedNum < _maxExitsToProcess && ExitPriority.parseExitableAt(uniquePriority) < block.timestamp) {
-            delete exits[uniquePriority];
+            delete delegations[uniquePriority];
             queue.delMin();
             processedNum++;
 
-            IExitProcessor processor = exit.exitProcessor;
-            processor.processExit(exit.exitId, _token);
+            processor.processExit(exitId, _token);
 
             if (queue.currentSize() == 0) {
                 break;
             }
 
             uniquePriority = queue.getMin();
-            exit = exits[uniquePriority];
+            exitId = ExitPriority.parseExitId(uniquePriority);
+            processor = delegations[uniquePriority];
         }
 
         emit ProcessedExitsNum(processedNum, _token);
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -28,7 +28,7 @@ contract ExitGameController is ExitGameRegistry {
     );
 
     event ExitQueued(
-        uint192 indexed exitId,
+        uint160 indexed exitId,
         uint256 uniquePriority
     );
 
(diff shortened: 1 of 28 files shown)


plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -9,15 +9,10 @@ import "../utils/TxPosLib.sol";
 
 contract ExitGameController is ExitGameRegistry {
     uint64 public exitQueueNonce = 1;
-    mapping (uint256 => Exit) public exits;
+    mapping (uint256 => IExitProcessor) public exitProcessors;
```
paulperegud: Please find a better name. `Delegations`?

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -3,21 +3,32 @@ pragma solidity ^0.5.0;
...
-     * The combinition of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
-     * 'exitableAt' only provide granuality of block, thus add 'txPos' to provide priority of transaction.
+     * @dev Given a utxo position and a unique ID, returns an exit priority.
+     * The combination of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
+     * 'exitableAt' only provide granularity of block, thus add 'txPos' to provide priority of transaction.
+     * @param exitId Unique exit identifier.
+     * @return An exit priority.
+     *   Anatomy of returned value, most significant bits first
+     *   42 bits - timestamp (exitable_at); unix timestamp fits into 32 bits
+     *   54 bits - blknum * 10^9 + txindex; to represent all utxo for 10 years we need only 54 bits
```
boolafish: this is incorrect since it is using `txPos` instead of `utxoPos` so actually have 10*4 more spaces here compare to the original one in `RootChain.sol`.

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -3,21 +3,32 @@ pragma solidity ^0.5.0;
...
-     * The first 64 bit for exitableAt, following 128 bits of txPos and then 64 bits of nonce.
-     * The combinition of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
-     * 'exitableAt' only provide granuality of block, thus add 'txPos' to provide priority of transaction.
+     * @dev Given a utxo position and a unique ID, returns an exit priority.
+     * The combination of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
+     * 'exitableAt' only provide granularity of block, thus add 'txPos' to provide priority of transaction.
+     * @param exitId Unique exit identifier.
+     * @return An exit priority.
+     *   Anatomy of returned value, most significant bits first
+     *   42 bits - timestamp (exitable_at); unix timestamp fits into 32 bits
```
boolafish: > unix timestamp fits into 32 bits

Since 32 bits timestamp has the 2038 issue, I think it might be better to comment the similar way like we do in `txPos` on how long can this schema hold.

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
pik694: but we really want here only tx position - which output exits is determined in exit id

boolafish: > to represent all utxo for 10 years we need only 54 bits

ah...I mean the comment itself is incorrect, not the txPos

pik694: heh, right!

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
pik694: I will fix the comments :+1:

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -3,21 +3,32 @@ pragma solidity ^0.5.0;
...
-     * The combinition of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
-     * 'exitableAt' only provide granuality of block, thus add 'txPos' to provide priority of transaction.
+     * @dev Given a utxo position and a unique ID, returns an exit priority.
+     * The combination of 'exitableAt' and 'txPos' is the priority for Plasma M(ore)VP protocol.
+     * 'exitableAt' only provide granularity of block, thus add 'txPos' to provide priority of transaction.
+     * @param exitId Unique exit identifier.
+     * @return An exit priority.
+     *   Anatomy of returned value, most significant bits first
+     *   42 bits - timestamp (exitable_at); we can represent dates until 2109 
+     *   54 bits - blknum * 10^9 + txindex; to represent all transactions for 10 years we need only 54 bits
```
boolafish: This is still wrong though. 
should be `blknum * 10^5 + txindex`, and of would have much longer life span than 10 years.

pdobacz: it might be worthwhile to add a reference (or a quick dissection in here) towards a source of this calculation and the 10^5 part. I take it it comes from here:

https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/utils/TxPosLib.sol#L11

and here:

https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/utils/TxPosLib.sol#L11

Also note a related issue https://github.com/omisego/plasma-contracts/issues/254, so that we not forget to update those docs, in case this modification gets through.

I think it's very important to keep all those offsetting and bitwise operations as transparent and documented as possible.

pik694: I made some more caution maths and it the truth appears to be yet another thing :joy:

### tx pos
Considering a UTXO_POS schema:
https://github.com/omisego/plasma-contracts/blob/c4aad906bb948bc59d1d7eda653a8219263d44e3/plasma_framework/contracts/src/utils/UtxoPosLib.sol#L6

We get than in each block there are at most `BLOCK_OFFSET / TX_OFFSET = 10^5` transactions, but we should also take into account `CHILD_BLOCK_INTERVAL` as each block number is multiplied by it (not counting deposits).
Right now `CHILD_BLOCK_INTERVAL` is set to `10^3`, and it determines number of deposits that can be made between the two consecutive blocks submitted be the operator (kudos to @pdobacz for pointing this out).

Therefore, we should reserve `10^8` per child chain block. Assuming that we commit a new child chain block in each root chain block (which happens every ~15 seconds), we must reserve `10^8 * 4 * 60 * 24 * 365` space a year (for simplicity let's don't take into account leap years). (please point out if my math is wrong there)

Having this said and having reserved 54-bits for tx pos, we will be out of space in ~85 years after the initial block - if we started now we would be out of space in ~2104.

### exitable timestamp
There is another bug with the timestamp: 42-bits will finish in year 141431 :joy:
Calculating the 2109, I took `2^42`, put it into https://www.epochconverter.com and received value 2109, but I did not notice that the smart website assumed the the value provided was in _milliseconds_ instead of seconds.

### Another solutions
I may just amend the comments, but there is also one more possible solution. Having done the math, we see that timestamp has ridiculously much space, on the contrary to tx pos; we can move some bits from timestamp to tx pos, in such a way that the years they exhaust are more "balanced". We could use 37-bits for timestamp and 59-bits for tx pos, in such case they will end in years 6328 and 4760 (2741 years since start) respectively.

We can also do one more thing: use  33-bits for timestamp and 55-bits for tx pos ( timestamp finishes in 2242 and tx pos has space to run for 171 years) and move the additional 8 bits to exit id and end up with 168-bits for exit id.

boolafish: > Right now CHILD_BLOCK_INTERVAL is set to 10^3

oh......I definitely missed this! Meanwhile, I think [Eth2.0 would make the 15 sec -> 6 sec](https://our.status.im/9-things-you-didnt-know-about-ethereum-2-0/), so would be half to 1/3 of our current calculation.

>  in milliseconds instead of seconds.

ah....I missed this whenever I was doing the calculation too @@ So yeah, it seems like it has more than enough bits. Let's re-balance this.

> 168-bits for exit id.

Though I really want the exitId to be longer so that it has at least full 160 bits for hash collision protection, but since there is no type `uint168`, I would prefer we stick to either `uint192` or `uint160`. Otherwise passing `uint192` with a real limit of `uint168` just make things more complex.

pik694: > I would prefer we stick to either `uint192` or `uint160`.

Unfortunatelly we cannot use 192-bits since we cannot fit both timestamp and tx pos in only 64 bits.

pik694: >BTW, all this complexity is what I think original ALD schema is better actually. It is just more easy to analyze relatively. And, either way we are approaching it is breaking the client anyway @@....

That's why I suggest just changing the documentation and explicitly clarify the numbers there.

The problem with the ALD schema is the overall cost - we store more bytes, we use nonces which we update each time which costs a loooot of gas.

PR #306, opened 2019-09-25
fix: use SafeMath for user input arithmetic operations

### LINKED ISSUE #296: Use SafeMath for all tainted inputs
## Issue Type 

```
[ x ] bug report
[ ] feature request
```

## Current Behavior
`SafeMath` is not consistently used across the contract system. Input from malicious users could potentially overflow or underflow variables and change the business logic in undesirable ways. 

In the following example this could happen theoretically when calculating the sum of outputs. While this is unlikely to happen practically in this case it still highlights the importance of taking every precaution. Further information on the nature of the issue can be found at https://consensys.github.io/smart-contract-best-practices/known_attacks/#integer-overflow-and-underflow or for examples check https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-101. 

https://github.com/omisego/plasma-contracts/blob/3eda481acb53ca74ff2420d73eb1fe5528e1eb6a/plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol#L107-L113 

## Suggested Fix
Review all arithmetic operations in the contract system and replace them with safe alternatives from OpenZeppelin's SafeMath library. Any arithmetic operation that contains at least one tainted value (can be influenced by any user in the system) should be changed to use the safe version.

pgebal:
Closes #296


thec00n:
Changes look. Let's have a look at the rest of the contracts and see where we need to add SafeMath operations.

thec00n:
@pik694 I would suggest that we do review individually all contracts and then we can can discuss context and what we think needs change. WDYT?

pik694:
This is a great idea, the more eyes the more bigger changes we see sth :)

pgebal:
> This is a great idea, the more eyes the more bigger changes we see sth :)

I'm cool with that. I've been looking through the code and that seems to  be it. But I'll go through it once more to make sure.

pik694:
This may cause problems in the future (when we have many inputs and outputs):
https://github.com/omisego/plasma-contracts/blob/9ca1b188dfe096a26ea100cb4e86bbbecc5b0811/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelUtils.sol#L26

pgebal:
> This may cause problems in the future (when we have many inputs and outputs):
> 
> https://github.com/omisego/plasma-contracts/blob/9ca1b188dfe096a26ea100cb4e86bbbecc5b0811/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelUtils.sol#L26

That's a separate issue and a limitation that I didn't think about.

As of now we use uint256 for exitMap, so uint8 is fine for indexing it. As long as we have less then 128 inputs/outputs we're good. I asked on slack how many outputs we are planning to support.

pik694:
This may cause problems if maintainer were malicious:
https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/utils/BondSize.sol#L73 (sorry for not attaching a permalink, but GH is having some issues right now)

pgebal:
> This may cause problems if maintainer were malicious:
> https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/utils/BondSize.sol#L73 (sorry for not attaching a permalink, but GH is having some issues right now)

right, amended and pushed changes

pgebal:
> > This may cause problems in the future (when we have many inputs and outputs):
> > https://github.com/omisego/plasma-contracts/blob/9ca1b188dfe096a26ea100cb4e86bbbecc5b0811/plasma_framework/contracts/src/exits/payment/PaymentInFlightExitModelUtils.sol#L26
> 
> That's a separate issue and a limitation that I didn't think about.
> 
> As of now we use uint256 for exitMap, so uint8 is fine for indexing it. As long as we have less then 128 inputs/outputs we're good. I asked on slack how many outputs we are planning to support.

Actually it would be good, to check there if inputIndex is not greater than 128 and fail on that explicitly.

pgebal:
@thec00n I think we are done. If you find something please put a comment. If there will be no comments I'll merge that tomorrow.

PR #310, opened 2019-09-26
300 test coverage

boolafish:
### Main changes
- Install and use the 0.7.0 beta version of solidity coverage. Somehow was not able to setup the coverage with 0.6 version in current master and it seems like others are migrating to the 0.7 beta already (such as open-zepplin)
- Update truffle to latest as the solidity coverage requires 0.5.31 which is higher then existing one.
- Update openzeppelin-test-helpers to latest as new truffle web3 dependency is incompatible with old one.
- Add "coverage" command in package.json. Can run coverage by "npm run coverag
- add coverage batch in README under `plasma_framework`
- install coverall

### Side changes
- use `package-lock` as circle ci cache key

### Manual setup for CI
- register `plasma-contract` to coverall
- link the folder `plasma_framework` to be the folder to get source code in coverall (done with coverall web UI)
- add `COVERALLS_REPO_TOKEN` as ENV var in Circle CI for `plasma-contract`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ commands:
           command: echo 'export PATH="$(npm bin):$PATH"' >> $BASH_ENV
       - restore_cache:
           keys:
-            - v2-dependencies-{{ checksum "plasma_framework/package.json" }}
+            - v2-dependencies-{{ checksum "plasma_framework/package-lock.json" }}
       - run:
           name: Install dependencies
           working_directory: ~/repo/plasma_framework
@@ -31,7 +31,7 @@ commands:
       - save_cache:
           paths:
             - plasma_framework/node_modules
-          key: v2-dependencies-{{ checksum "plasma_framework/package.json" }}
+          key: v2-dependencies-{{ checksum "plasma_framework/package-lock.json" }}
 
   setup_python_env:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,17 +21,17 @@ commands:
           command: echo 'export PATH="$(npm bin):$PATH"' >> $BASH_ENV
       - restore_cache:
           keys:
-            - v2-dependencies-{{ checksum "plasma_framework/package-lock.json" }}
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -21,17 +21,19 @@ commands:
...
       - restore_cache:
           keys:
-            - v2-dependencies-{{ checksum "plasma_framework/package.json" }}
+            - v2-dependencies-{{ checksum "plasma_framework/package-lock.json" }}
       - run:
           name: Install dependencies
           working_directory: ~/repo/plasma_framework
           command: |
+            # FIXME: workaround for the error on old web3 version using git: https://circleci.com/gh/omisego/plasma-contracts/2973
+            # also the 'rm -rf node_modules/**/.git' workaround does not solve this
```
boolafish: I've tried to use `rm -rf node_modules/**/.git` which is stronger than the original one. However, it still have the problem in CI (though locally my Mac would be able to `npm install`).

Since it is only having issue in that specific package, I `rm` that single package instead. Also, as sb pointed out, rm the `.git` file is probably not a good practice: https://github.com/OpenZeppelin/openzeppelin-sdk/issues/929#issuecomment-499271999

pik694: > Also, as sb pointed out, rm all .git file is probably not a good practice
The practice is acceptable, though, as we are running it in a container and there we never have some local copies of our repos that cannot be restored.

boolafish: ah, yeah I agree. But anyway that workaround is not working now so do we still want to keep that?

pik694: How about this: let me think a little about the issue - shouldn't I find a solution till Monday, then we will remove the problematic dependency from the cache, therefore we will be able to remove the workaround (as we will replace it with another :joy:).

cgewecke:
@boolafish Hi, I help maintain solidity-coverage. Thanks for trying out the beta - if you run into any difficulties with it or see room for improvement please just lmk :)

boolafish:
@pik694 I've cleaned up the git commit a bit. If you think it's ready to go please approve it :)

PR #313, opened 2019-09-26
fix: bytes32 owner -> bytes20 outputGuard in eip712 TX_TYPE_HASH

kevsul:
@pgebal I found a `bytes32 owner` in the precomputed eip712 TX_TYPE_HASH


pgebal:
Good catch Kevin. Feel free to merge that after solving issues with Python tests.

PR #319, opened 2019-09-30
301 upgradeable protocol

### LINKED ISSUE #301: Document the upgradeability of `Protocol`/`TxFinalization` (MVP vs MoreVP)
EDIT: this becomes mainly a documentation request issue, scroll down to comments for details

## Issue Type 

```
[ ] bug report
[x] feature request (design)
```

## Current Behavior

Protocols are baked into the PlasmaFramework and non-upgradeable (e.g. the implementation of validation of confirmation signatures)

## Expected Behavior

This should be upgradeable, and can by by relying on the txtype to differentiate protocols and SpendingCondition/ExitGame contracts to deliver the implementation of the protocols. Then one can introduce a new protocol via an upgrade.

For reference, the original convo from Slack (curated):

> 
> Question:
> ```
>      * @param _protocol The protocol of the transaction, 1 for MVP and 2 for MoreVP.
>      */
>     function registerExitGame(uint256 _txType, address _contract, uint8 _protocol) public onlyOperator {
> ```
> is the `_protocol` bit really needed? Shouldn't we rather have the transaction type drive this parameter and the rest be agnostic of it? I recall a conversation with Pepesza, where it seemed that it would be possible for the transaction type / exit game to carry the burden of this abstraction. It would be desirable, as then MVP vs MoreVP is upgradeable and the rest is agnostic of this
> 
> pepesza  6 days ago
> That is true for cases that we have right now. I.e. payments are clearly better with MoreVP than MVP. And dex settlements are better with MVP than with plain MoreVP.
> Will we have cases where users would want to run different protocol for the same tx type?
> 
> piotr  6 days ago
> I think it's more about upgradeability and separation of concerns than about MVP vs MoreVP (actually, not having to decide on which "VP"s we support is the case for keeping this part abstracted out & upgradeable)
> 
> pepesza  6 days ago
> Agree with your last point.
> 
> piotr  6 days ago
> another argument for upgradeability of this:
> https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/utils/TxFinalization.sol#L100
> here we check only the root, what if we want to check root and blknum was signed (which I think we recently talked about was necessary to prevent attacks from operator)? We'd need an upgrade of this impl.
> I fear this baking in 2 particular impls of the protocol. We put so much effort into flexible things like SpendingCondition whatnot, but this stays very rigid
> 
> piotr  6 days ago
> Wouldn't it be possible to resolve the confirm sigs via spending condition + special witnesses that contain the required confirmation signature? Then the MVP implementation (confirm sigs) is internal to the ExitGame abstraction and can be swapped out for a new one (edited) 
> 
> pepesza  6 days ago
> EDIT: discussed this problem with @piotr. It seems to be working. Also, my doubts around combinatorial explosion were dispelled - the particular problem is solved by spendingCondition.
> Boolafish what do you think about Piotr's proposal?
> 
> boolafish  6 days ago
> no spending condition itself is not enough, I thought of that as well.
> 
> during IFE, you need to make sure your input tx itself is finalized, and that would need to check the confirm sig of input tx. I guess your proposal would only check the confirm sig of spending tx
> 
> well…I thought the assumption of the TxFinalization would be pretty solid toward future so that’s why I use single library for that. But if there is concern we can always switch to interface-register based implementation
> 
> haha, likewise I am still very concerned about the WireTransaction library implementation. XDD
> 
> Wouldn’t it be possible to resolve the confirm sigs via spending condition + special witnesses that contain the required confirmation signature?
> oh, wait potentially we can check inputTx is standardFinalized + spendingTx is protocolFinalized in spending condition…. do you mean this? (edited) 
> 
> piotr  4 days ago
> ok, after talking about the output type public, let's go back to this one :sweat_smile:
> lemme refresh the terminology in my mind...
> 
> uhm, I'm not sure I parse this well and adjust to the terms, so I'll talk around:
> on startIFE:
> - you prove inclusion of input txs
> - you provide the IFE tx
> - you "connect" IFE txs with the input txs. This connecting consist in checking the spending conditions for all the IFE tx's inputs. So if any input happens to be MVP, this spending condition will include providing a suitable witness (one with confirm sig) and validating against block number/confirmer etc.
> So I think yes, this sounds like this:
> check inputTx is standardFinalized + spendingTx is protocolFinalized in spending condition
> just the standard/protocolFinalized terminology isn't fully sinked in with me
> 
> boolafish  4 days ago
> Some downside of this is that the inputTx inclusion proof would be re-checked when doing challengeSE
> 
> Also when challenge non canonical and challenge piggy back input
> 
> piotr  4 days ago
> no sure what you mean re-checked here?
> 
> boolafish  4 days ago
> Oh so it will be checked once when starting exit
> 
> Might not be a big deal though,just non optimizing
>

boolafish:
### Note
1. TxFinalization --> ITxFinalizationVerifier + TxFinalizationModel + Implementation of interface.
2. remove the code around MVP
3. Dependecy inject with ITxFinalizationVerifier so next ExitGame can inject with the updated one.

closes #301


plasma_framework/contracts/src/exits/utils/TxFinalizationVerifier.sol
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            // TODO: solhint disabled for now due to bug, https://github.com/protofire/solhint/issues/157
+            // solhint-disable-next-line reason-string
+            revert("invalid protocol value");
+        }
+    }
+
+    /**
+    * @notice Checks a transaction is "protocol finalized" or not
+    * @dev MVP: need to be standard finalzied.
```
kevsul:
```suggestion
    * @dev MVP: need to be standard finalized.
```

Issue #290, opened 2019-09-20
Make output_type public in the output data structure


pdobacz:
Following our discussion on a call, I'm suggesting that the `output_type`, now planned to be included in the obfuscated data structure (`output_guard`), is made public, similar to how `tx_type` is handled.

Below is the description of the issue and suggested change to the ALD.

Let's take some time to think about the implications.

## Issue Type 

```
[ ] bug report
[x] feature request (design)
```

## Current Behavior

The current design keeps the `output_type` included in the `output_guard_preimage` and is delivered in the transaction data obfuscated. This means:
 - we're conflating two distinct mechanism - the exchange deposit obfuscation mechanism and the ALD's `output_type` discovery
 - we're limiting the discovery of `output_type` only if you have the remainder of the preimage (think: a query to get all outputs of certain type)
 - we're making the discovery of the `owner` of outputs harder (think: a query to get all outputs of all types belonging to someone - you need to know all the possible types)
 - we're making modifications to the decoding of an output harder - one needs the witness data (preimage) to discover the output type (think: changing the encoding of an output entirely)
 - we're making the so-called 1:1 binding more difficult: since `output_type` determines `input_pointer_type`, child chain/watcher wouldn't know how to store the output in their UTXO set. (1:1 binding makes UTXOset management easier and more abstract and also double-spend detection easier, see [here](https://github.com/omisego/research/issues/93#issuecomment-517734509) for more info)
 - we're forcing ourselves to include things like `confirmer_address` in the `output_guard_preimage` too, thereby obfuscating it and leading to similar problems as written above

## Expected Behavior

`output_type` is delivered similar to `tx_type` as the first n bit field of the structure (presumably RLP encodable array), in public. The obfuscation mechanism is set separate from the `output_type` typing - `output_guard` may contain either the hashed preimage or (hashed) owner's address, **keeping the `output_type` the same**. Then the witness to unlock such output may contain either the preimage (+other data if applicable) or the owners signature.

## Motivation for Change

If the `output_type` is public:
 - we can dispatch do different decoding schemes if the current one turns out to be insufficient. An example of this is that we can then include `confirm_address` of an MVP output in the public section of the output data structure
 - we're keeping the two mechanisms (dispatch & obfuscation) separate
 - we still can have >1 `output_type` for a single creating `tx_type`, which seems to be required
 - we have the same dispatch/typing mechanism for output and transaction

paulperegud:
LGTM

[...]

PR #321, opened 2019-09-30
290 add output type to tx

kevsul:
## Description

Make outputType publicly readable in the transaction by moving it out from the outputGuard into the output field. (as discussed in #290 ) 

Many of the contract APIs no longer need to be passed outputType as an explicit param, as they can now parse the tx to get it.


plasma_framework/test/helpers/transaction.js
```diff
@@ -1,24 +1,24 @@
...
+const { EMPTY_BYTES_32, OUTPUT_TYPE } = require('../helpers/constants.js');
 
-const EMPTY_BYTES32 = `0x${Array(64).fill(0).join('')}`;
 const TransactionTypes = {
     PLASMA_DEPOSIT: 1,
 };
 
 class PaymentTransactionOutput {
-    constructor(amount, owner, token) {
+    constructor(amount, owner, token, type = OUTPUT_TYPE.PAYMENT) {
```
boolafish: minor: might be better and more intuitive to use if we put the args with the same order of the output. so `type, owner, token, amount`?

kevsul: yeah, I agree... and I started that way, but then took the easy way out above because it meant much less changes to the existing tests 😂 
But I'll go back and make the changes - it is better to have the correct order

boolafish:
`PaymentEip712Lib.sol` need to change the output structure too :sweat:

sadly it is not auto detected/covered by test on the mismatch of structures....

Issue #282, opened 2019-09-18
Transparent handling of token standards (ERC20, ERC721, etc)


paulperegud:
We can handle different token standards transparently - without introducing a specialized exit game or specialized transaction types. **Note - we can do this only now!** Adding such functionality later would not be possible via upgrade mechanism (discussed with @boolafish) since it requires a change in output format.

#### Vaults today
At this moment vaults implement two callbacks (argument list is the proposed one, not the one implemented):
* `withdraw(token_contract_address, owner, standard_specific_data): void` (throw if something is wrong)
* `process_deposit(token_contract_address, owner, standard_specific_data): void` (throw if something has gone wrong)

Logic of choosing what vault is the proper one is tied to transaction type. That means that we have one tx type for ERC20 and one for ERC721 and one for mix of them. We can do better.

#### Required change in vaults
We add one more callback that each vault needs to implement. It looks like this.
isPrintingMoney(this_vault_specific_inputs: [output()], this_vault_specific_outputs: [output()]): boolean

#### Required change in output format
We add one more field and we drop `amount` in favor of `standard_specific_data`. Modified output format looks like this:
* `output_guard` - ....
* `token_standard` <- new field, `uint`
* `token_contract_address` - `address`
* `standard_specific_data` - `bytes`

Last field being just `bytes` hides internal structure of `standard_specific_data`. This is important since Solidity does not support dynamic types and we want to avoid leaking of information about the types and number of elements inside all over the codebase. We take whatever contents of that field and RLP encode them.

All operations on standard_specific_data would be done inside of the 3 callbacks above - callback would be the only one that unpacks that field. Contract would choose proper vault by reading the `token_standard` value. E.g. when `isPrintingMoney` needs to be called, inputs and outputs need to be grouped by `token_standard` value and for each group we call `isPrintingMoney` once.

Example of the way token specific data can be encoded is below.
For ERC721:
* `token_standard` = 2 -> erc721 vault implementation
* `standard_specific_data` = `RLP.encode([uint256 token_uid, ...]): bytes`

For ERC20:
* `token_standard` = 1 -> 
* `standard_specific_data` = `RLP.encode(uint256 amount): bytes`

For buggy ERC20 implementation:
* `token_standard` = x -> 
* `standard_specific_data` = `RLP.encode(uint256 amount): bytes`).

boolafish:
yeah, I think this is the real `WireTransaction` data structure that can survive foreseeable future.

[...]

PR #324, opened 2019-10-01
feat: process exits by (vaultId, token) pair

pgebal:
Implements optional part from: https://github.com/omisego/plasma-contracts/issues/282#issuecomment-535429760

We introduce an exit queue per (vault id, token) pair to isolate processing exits that are funded by different vaults.


plasma_framework/contracts/src/exits/payment/controllers/PaymentPiggybackInFlightExit.sol
```diff
@@ -27,6 +27,9 @@ library PaymentPiggybackInFlightExit {
     uint8 constant public MAX_INPUT_NUM = 4;
     uint8 constant public MAX_OUTPUT_NUM = 4;
 
+    // we use a single exit processing queue
```
pgebal: I don't like that, but since we didn't decide to implement transparent tokens in outputs we don't have a way to enqueue exit to the right queue. That's the reason we put everything into a single queue (no matter what the vault id is). That's problematic, since after we deploy a tx format and exit game that enqueues in the right way (using a proper vault_id), we'll have a discrepancy (some exits will be enqueued with vault_id 1, even though non-id 1 vaults fund that).

boolafish: I would inject the right vault id of `EthVault` and `Erc20Vault` and remove the injection of address and fetch on construction instead.

plasma_framework/contracts/mocks/exits/SpyPlasmaFrameworkForExitGame.sol
```diff
@@ -24,12 +25,15 @@ contract SpyPlasmaFrameworkForExitGame is PlasmaFramework {
     }
 
     /** override for test */
-    function enqueue(address _token, uint64 _exitableAt, TxPosLib.TxPos calldata _txPos, uint160 _exitId, IExitProcessor _exitProcessor)
+    function enqueue(uint256 _vaultId, address _token, uint64 _exitableAt,
```
boolafish: Minor: can we make each var a line

plasma_framework/contracts/mocks/framework/DummyExitGame.sol
```diff
@@ -40,8 +40,10 @@ contract DummyExitGame is IExitProcessor {
         exitGameController = ExitGameController(_contract);
     }
 
-    function enqueue(address _token, uint64 _exitableAt, uint256 _txPos, uint160 _exitId, IExitProcessor _exitProcessor) public {
-        uniquePriorityFromEnqueue = exitGameController.enqueue(_token, _exitableAt, TxPosLib.TxPos(_txPos), _exitId, _exitProcessor);
+    function enqueue(uint256 _vaultId, address _token, uint64 _exitableAt, uint256 _txPos, uint160 _exitId, IExitProcessor _exitProcessor)
+    public
```
boolafish: Minor: I think the solidity style is that the `public` needs one more tab/4-spaces

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -48,6 +48,8 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
     const DEPOSIT_VALUE = 1000000;
     const INITIAL_IMMUNE_VAULTS = 2; // ETH and ERC20 vault
     const INITIAL_IMMUNE_EXIT_GAMES = 1; // 1 for PaymentExitGame
+    const ETH_VAULT_ID = 1;
+    const ERC20_VAULT_ID = 2;
```
boolafish: minor: consider putting these two into `constants.js` (For all the other `VAULT_ID` in other tests too)

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -8,14 +8,21 @@ import "./utils/ExitPriority.sol";
 import "../utils/TxPosLib.sol";
 
 contract ExitGameController is ExitGameRegistry {
+
     mapping (uint256 => IExitProcessor) public delegations;
-    mapping (address => PriorityQueue) public exitsQueues;
+    mapping (uint256 => bool) public registeredVaults;
```
boolafish: Instead of having two stuffs: vault, token with two sets of actions, I would merge into one and have the following functions/events instead:

```
event ExitQueueAdded(vaultId, token)
addExitQueue(vaultId, token)
hasExitQueue(vaultId, token)
```

and wouldn't need this extra mapping anymore.

I think this is much more simpler and easier to understand.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -120,7 +155,7 @@ contract ExitGameController is ExitGameRegistry {
             processor = delegations[uniquePriority];
         }
 
-        emit ProcessedExitsNum(processedNum, _token);
+        emit ProcessedExitsNum(processedNum, token);
```
boolafish: please add `vaultId` to this event too, or replace `token` with `exitQueueKey `

plasma_framework/contracts/mocks/framework/ReentracyExitGame.sol
```diff
@@ -22,10 +26,12 @@ contract ReentrancyExitGame is IExitProcessor {
...
     // This would call the processExits back to mimic reentracy attack
     function processExit(uint160, address) public {
-        exitGameController.processExits(testToken, 0, reentryMaxExitToProcess);
+        exitGameController.processExits(EXIT_FUNDING_VAULT_ID, testToken, 0, reentryMaxExitToProcess);
     }
 
-    function enqueue(address _token, uint64 _exitableAt, uint256 _txPos, uint160 _exitId, IExitProcessor _exitProcessor) public {
-        exitGameController.enqueue(_token, _exitableAt, TxPosLib.TxPos(_txPos), _exitId, _exitProcessor);
+    function enqueue(uint256 _vaultId, address _token, uint64 _exitableAt, uint256 _txPos, uint160 _exitId, IExitProcessor _exitProcessor)
+    public
```
boolafish: Minor: I think the solidity style is that the public needs one more tab/4-spaces

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
```diff
@@ -72,10 +72,18 @@ contract PaymentInFlightExitRouter is IExitProcessor, Operated, OnlyWithValue {
             supportedTxType
         );
 
+        uint256 ethVaultId = framework.vaultToId(address(ethVault));
```
boolafish: I would replace injection of vault address -> vault id for `Router` too. And fetch vault contract here instead. It is more client friendly for them to inject with `id` instead of `address` which would differs on different deployment environment.

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
```diff
@@ -72,10 +72,18 @@ contract PaymentInFlightExitRouter is IExitProcessor, Operated, OnlyWithValue {
...
+        require(ethVaultId != 0, "Invalid ETH vault");
+
+        uint256 erc20VaultId = framework.vaultToId(address(erc20Vault));
+        require(ethVaultId != 0, "Invalid ERC20 vault");
```
boolafish: ethVaultId -> erc20VaultId.

plasma_framework/contracts/src/framework/ExitGameController.sol
pgebal: I did so a vault can only be registered by operator. That's why there is registeredVaults map. Do you think that's not an issue that users will be allowed to register vaults and we will get some bogus vaultIds?

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
pgebal: That's a good idea!

plasma_framework/contracts/src/framework/ExitGameController.sol
boolafish: hmmm, I don't *think* it should be worry-some as we are not limiting `addToken` to operator too. Seems to me who create the queue is not important, and also whether that queue is of some non even existing vault id does not matter.

But let get another pair of eye on this: @paulperegud @pdobacz @kevsul

pdobacz: What would be the consequence of someone `addExitQueue` for an important token for all `vaultIds` (is that even possible? is the `vaultID` for new vault chosen by the operator or predetermined (can't add vault 13 if there's no vault 12 yet))

If the above is a non-issue then I think we're cool.

If it is an issue then that would require moving the `PriorityQueues` to be anchored in vaults rather than `PlasmaFramework` which sounds like a uge change. (that is one does `IVault.addToken(token)` instead)

pgebal: @pdobacz Vault id is chosen by operator when registering a vault in VaultRegistry. Right now we allow anyone to create a queue for a token, but vaults can only be registered by operator. My idea was to mirror that behavior.

boolafish: > What would be the consequence of someone addExitQueue for an important token for all vaultIds

My claim is that it would be fine and doesn't matter as long as our ExitGame contracts code it *correctly* when putting things into exit queue.

> is that even possible?

In my proposal it is. It basically allow anyone to generate any arbitrary exit queue.

pdobacz: ok, so if the vault id is at the discretion of the operator when they're registering the vault, then I think the above scenario is a non-issue :+1:

pgebal: ok

plasma_framework/contracts/mocks/exits/SpyPlasmaFrameworkForExitGame.sol
pgebal: done

plasma_framework/contracts/mocks/framework/DummyExitGame.sol
pgebal: done

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
pgebal: done

plasma_framework/contracts/mocks/framework/ReentracyExitGame.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
pgebal: done

tests/contracts/root_chain/test_tokens.py
```diff
@@ -1,28 +1,9 @@
...
-def test_token_adding(token, plasma_framework):
-    assert not plasma_framework.hasToken(token.address)
-    plasma_framework.addToken(token.address)
-    assert plasma_framework.hasToken(token.address)
-    with pytest.raises(TransactionFailed):
-        plasma_framework.addToken(token.address)
-
-
-def test_token_adding_gas_cost(w3, plasma_framework):
+def test_exit_queue_adding_gas_cost(w3, plasma_framework):
```
pgebal: Removed other tests as it's not crucial and already tested in truffle End-To-End tests.

plasma_framework/contracts/src/framework/ExitGameController.sol
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
         uint256 processedNum = 0;
 
-        while (processedNum < _maxExitsToProcess && ExitPriority.parseExitableAt(uniquePriority) < block.timestamp) {
+        while (processedNum < maxExitsToProcess && ExitPriority.parseExitableAt(uniquePriority) < block.timestamp) {
             delete delegations[uniquePriority];
             queue.delMin();
             processedNum++;
 
-            processor.processExit(exitId, _token);
+            processor.processExit(exitId, token);
```
boolafish: I think we should change the interface for `processExit` too. So we should call with `processExit(exitId, vaultId, token)`.

plasma_framework/test/helpers/constants.js
```diff
@@ -14,6 +14,9 @@ const OUTPUT_TYPE = {
 const EMPTY_BYTES = '0x';
 const EMPTY_BYTES_32 = '0x0000000000000000000000000000000000000000000000000000000000000000';
 
+const ETH_VAULT_ID = 1;
```
boolafish: I would probably do the following instead:
```
VAULT_ID = {
    ETH: 1,
    ERC20: 2,
}
```

It is more consistent with how we treat `TX_TYPE` and `OUTPUT_TYPE`

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -107,8 +107,11 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
         this.erc20Vault = await Erc20Vault.new(this.framework.address);
         await this.erc20Vault.setDepositVerifier(erc20DepositVerifier.address);
 
-        await this.framework.registerVault(1, this.ethVault.address);
-        await this.framework.registerVault(2, this.erc20Vault.address);
+        await this.framework.registerVault(ETH_VAULT_ID, this.ethVault.address);
+        await this.framework.registerVault(ERC20_VAULT_ID, this.erc20Vault.address);
+
+        await this.framework.addExitQueue(ETH_VAULT_ID, ETH);
+        await this.framework.addExitQueue(ERC20_VAULT_ID, this.erc20Vault.address);
```
boolafish: There is a test of `Given ERC20 token added to the PlasmaFramework` and would add the queue for ERC20 token there. So it shouldn't be done here. However, just realize we can add the queue twice seems like a **bug**!

(wait....I see the code should actually block this, how come this works.........

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -107,8 +107,11 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
         this.erc20Vault = await Erc20Vault.new(this.framework.address);
         await this.erc20Vault.setDepositVerifier(erc20DepositVerifier.address);
 
-        await this.framework.registerVault(1, this.ethVault.address);
-        await this.framework.registerVault(2, this.erc20Vault.address);
+        await this.framework.registerVault(ETH_VAULT_ID, this.ethVault.address);
+        await this.framework.registerVault(ERC20_VAULT_ID, this.erc20Vault.address);
+
+        await this.framework.addExitQueue(ETH_VAULT_ID, ETH);
```
boolafish: Minor: consider add similar test as `Given ERC20 token added to the PlasmaFramework` instead of initiating here.
But I am fine without doing this because at one point we might want to change the setup here using our migration script result instead (so more real e2e) and our migration script might want to add ETH by default.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
-     * @return bool whether the queue for a token was created.
+     * @notice Adds queue to the plasma framework.
+     * @dev the queue is created as a new contract instance.
+     * @param vaultId Id of the vault
+     * @param token Address of the token.
      */
-    function hasToken(address _token) public view returns (bool) {
-        return address(exitsQueues[_token]) != address(0);
+    function addExitQueue(uint256 vaultId, address token) external {
+        require(vaultId != 0, "Invalid vault id");
```
boolafish: minor: I think the message can be more clear on not able to be 0

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -17,16 +17,18 @@ contract ExitGameController is ExitGameRegistry {
     // exit priority => IExitProcessor
     mapping (uint256 => IExitProcessor) public delegations;
     // token => PriorityQueue
-    mapping (address => PriorityQueue) public exitsQueues;
+    mapping (bytes32 => PriorityQueue) public exitsQueues;
```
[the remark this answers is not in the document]
pgebal: Done. You are right, one always need to remember to update those comments.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
+     * @notice Checks if queue for particular token was created.
+     * @param vaultId Id of the vault that handles the token
+     * @param token Address of the token.
+     * @return bool represents whether the queue for a token was created.
      */
-    function addToken(address _token) external {
-        require(!hasToken(_token), "Such token has already been added");
+    function hasExitQueue(uint256 vaultId, address token) public view returns (bool) {
+        bytes32 key = exitQueueKey(vaultId, token);
+        return address(exitsQueues[key]) != address(0);
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/mocks/framework/ReentracyExitGame.sol
```diff
@@ -9,6 +9,10 @@ import "../../src/vaults/EthVault.sol";
 import "../../src/utils/TxPosLib.sol";
 
 contract ReentrancyExitGame is IExitProcessor {
+
+    // we use a single exit processing queue
+    uint256 constant public EXIT_FUNDING_VAULT_ID = 1;
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
+    }
 
-        exitsQueues[_token] = new PriorityQueue();
-        emit TokenAdded(_token);
+    function hasExitQueue(bytes32 queueKey) private view returns (bool) {
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
         returns (uint256)
     {
-        require(hasToken(_token), "Such token has not been added to the plasma framework yet");
-        PriorityQueue queue = exitsQueues[_token];
+        bytes32 key = exitQueueKey(vaultId, token);
+        require(hasExitQueue(key), "Such token has not been added to the plasma framework yet");
+        PriorityQueue queue = exitsQueues[key];
 
-        uint256 uniquePriority = ExitPriority.computePriority(_exitableAt, _txPos, _exitId);
+        uint256 uniquePriority = ExitPriority.computePriority(exitableAt, txPos, exitId);
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -39,89 +41,97 @@ contract ExitGameController is ExitGameRegistry {
...
+     * @param maxExitsToProcess maximal number of exits to process.
      * @return total number of processed exits
      */
-    function processExits(address _token, uint160 _topExitId, uint256 _maxExitsToProcess) external {
-        require(hasToken(_token), "Such token has not been added to the plasma framework yet");
-
-        PriorityQueue queue = exitsQueues[_token];
+    function processExits(uint256 vaultId, address token, uint160 topExitId, uint256 maxExitsToProcess) external {
+        bytes32 key = exitQueueKey(vaultId, token);
+        require(hasExitQueue(key), "Such token has not been added to the plasma framework yet");
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
pgebal: done

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
pgebal: That's what I thought, so I'll leave it as it is.

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
pgebal: Thankfully there is no bug and we have it covered by a unit test.
I messed up a bit, those are not the same calls, the first one should be removed:
`
await this.framework.addExitQueue(ERC20_VAULT_ID, this.erc20Vault.address);
await this.framework.addExitQueue(ERC20_VAULT_ID, this.erc20.address);
`

boolafish: ahhh......`erc20Vault` vs `erc20`

plasma_framework/test/helpers/constants.js
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
pgebal: done

plasma_framework/contracts/mocks/framework/ReentracyExitGame.sol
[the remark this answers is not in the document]
boolafish: minor: I guess we can remove this line then.

plasma_framework/contracts/src/framework/ExitGameController.sol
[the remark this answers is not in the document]
boolafish: lol, it is not changed....but it is minor stuff anyway

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -168,7 +181,16 @@ contract ExitGameController is ExitGameRegistry {
...
-        return exitsQueues[_token].getMin();
+    function exitQueueKey(uint256 vaultId, address token) private pure returns (bytes32) {
+        return keccak256(abi.encodePacked(vaultId, token));
+    }
+
+    function hasExitQueue(bytes32 queueKey) private view returns (bool) {
+        return address(exitsQueues[queueKey]) != address(0);
+    }
+
+    function getNextExit(uint256 vaultId, address token) external view returns (uint256) {
```
boolafish: minor: This is actually external, so should be grouped slight above.

plasma_framework/package.json
```diff
@@ -31,7 +31,7 @@
   "scripts": {
     "test": "truffle test",
     "coverage": "truffle run coverage",
-    "docgen": "[[ -z $(git status --porcelain) ]] && rm -rf contracts/mocks && solidoc ./ ./docs && git checkout -- contracts/mocks || echo 'Please make sure all files are committed and the git status is clean' ",
+    "docgen": "bash docgen.sh",
```
boolafish: I guess we can just put `./docgen.sh` ? (not sure whether that `bash` would work on my side since I am using `zsh`...)

eg. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/package.json#L13
Meanwhile, shall we put it under `scripts/` ?

pgebal: done

PR #328, opened 2019-10-02
Some Minor fixes

### LINKED ISSUE #296: Use SafeMath for all tainted inputs

boolafish:
### Note
- remove some solhint workaround as latest version fixes the bug. @kevsul 's PR has been merged and released.
- take care of BondSize boundary number, closes #296. End up not using `SafeMath` as it is really meaning less since it only protects `uint256` but we are using `uint128`. Instead, I add tests on that we support edge case of upgrade to max num of `uint128`.
- Add boundary of the BondSize to not be 0 during update. Thus with min size 1.
- fix bug around comparing time in BondSize. There was a bug when `now == effectiveUpdateTime` the update would be broken. Found this while trying to add tests for coverage : )
- make `depositVerifier` be updated right at `newDepositVerifierMaturityTimestamp`

---
- ~upgrade to solidity 0.5.12~ see this comment: https://github.com/omisego/plasma-contracts/pull/328#issuecomment-537399670
- ~make compiler happy when optimizer is off. Somehow it failed to compile when the optimizer is tuned off with the change on BondSize. (Whenever there is a `require` inside `buildParams` it fails to compile....). Found this because the [coverage tool forces optimizer to be off](https://github.com/sc-forks/solidity-coverage/issues/417). Seems like it cannot be called in a constructor. So I change the constructor for `Routers` to become `init` function instead.~ Reverting this as it end up increase deployment gas quite a lot, and we don't really have security issue as the params are fixed in PaymentExitGame contract now.

### DIFF
--- a/plasma_framework/contracts/src/exits/utils/BondSize.sol
+++ b/plasma_framework/contracts/src/exits/utils/BondSize.sol
@@ -50,7 +50,7 @@ library BondSize {
     function updateBondSize(Params storage self, uint128 newBondSize) internal {
         validateBondSize(self, newBondSize);
 
-        if (self.updatedBondSize != 0 && now > self.effectiveUpdateTime) {
+        if (self.updatedBondSize != 0 && now >= self.effectiveUpdateTime) {
             self.previousBondSize = self.updatedBondSize;
         }
         self.updatedBondSize = newBondSize;
--- a/plasma_framework/contracts/src/exits/utils/BondSize.sol
+++ b/plasma_framework/contracts/src/exits/utils/BondSize.sol
@@ -26,7 +26,7 @@ library BondSize {
         uint16 upperBoundMultiplier;
     }
 
-    function buildParams(uint128 _initialBondSize, uint16 _lowerBoundDivisor, uint16 _upperBoundMultiplier)
+    function buildParams(uint128 initialBondSize, uint16 lowerBoundDivisor, uint16 upperBoundMultiplier)
         internal
         pure
         returns (Params memory)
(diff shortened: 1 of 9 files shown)


pdobacz:
is there a requirement to go solc `0.5.12` now? If `0.5.11` works just fine, could we stick to it for just a while longer to limit the number of things in flux?

boolafish:
I think we should stick with latest version before we start the audit. I don't mind to upgrade that a bit later though. Would this impact your side?

plasma_framework/test/src/utils/BondSize.test.js
```diff
@@ -6,87 +6,195 @@ contract('BondSize', () => {
...
+                expect(bondSize).to.be.bignumber.equal(this.initialBondSize);
+
+                // Wait the remaining waiting period
+                await time.increase(HALF_WAITING_PERIOD);
+                bondSize = await this.contract.bondSize();
+                expect(bondSize).to.be.bignumber.equal(secondNewBondSize);
+            });
+        });
+
+        describe('with boundary size of numbers', () => {
```
boolafish: @thec00n here is the new tests for boundary

pdobacz:
@boolafish 

> @pdobacz I think we should stick with latest version before we start the audit. I don't mind to upgrade that a bit later though. Would this impact your side?

we'd probably like to upgrade to the same version, which impacts the builder/deployer images for `elixir-omg` and the deployment process a bit.

(or we could just yolo and use different versions for now, but if sth funny happens we'll have one more thing to debug).

If we can upgrade a bit later, I'd appreciate that. Maybe after the branch-entanglement and deployment changes are behind us?

But if time comes and you need to, pls upgrade and let us know, thx :pray:

plasma_framework/contracts/src/exits/utils/BondSize.sol
```diff
@@ -26,52 +26,57 @@ library BondSize {
...
     * @param newBondSize the new bond size.
     */
-    function updateBondSize(Params storage _self, uint128 newBondSize) internal {
-        validateBondSize(_self, newBondSize);
+    function updateBondSize(Params storage self, uint128 newBondSize) internal {
+        validateBondSize(self, newBondSize);
 
-        if (_self.updatedBondSize != 0 && now > _self.effectiveUpdateTime) {
-            _self.previousBondSize = _self.updatedBondSize;
+        if (self.updatedBondSize != 0 && now >= self.effectiveUpdateTime) {
```
boolafish: This `now >= self.effectiveUpdateTime` part was a bug! It was `now > self.effectiveUpdateTime` but in `bondSize(self)` it is using `now < self.effectiveUpdateTime`. As a result, when `now == self.effectiveUpdateTime` it would fail to fall into this `if` block.

plasma_framework/test/src/utils/BondSize.test.js
```diff
@@ -6,87 +6,165 @@ contract('BondSize', () => {
...
-        let bondSize = await this.contract.bondSize();
-        expect(bondSize).to.be.bignumber.equal(this.initialBondSize);
-
-        // Wait the remaining waiting period
-        await time.increase(HALF_WAITING_PERIOD);
-        bondSize = await this.contract.bondSize();
-        expect(bondSize).to.be.bignumber.equal(secondNewBondSize);
+    describe('with boundary size of numbers', () => {
+        it('should able to update to max number of uint128 without having overflow issue', async () => {
+            const maxSizeOfUint256 = (new BN(2)).pow(new BN(128)).sub(new BN(1)); // 2^128 - 1
```
pgebal: Uint256 -> Uint128

plasma_framework/test/src/utils/BondSize.test.js
thec00n: I did not appreciate https://github.com/omisego/plasma-contracts/blob/11b3bae4926dabc491dfbce511207b2f2798ba0f/plasma_framework/contracts/src/exits/utils/BondSize.sol#L75 in the beginning but I think it's actually the best solution in this case to ensure that the check does not overflow and scenario where we get stuck with a new bond size.

plasma_framework/test/src/utils/BondSize.test.js
```diff
@@ -6,87 +6,165 @@ contract('BondSize', () => {
...
-    beforeEach(async () => {
-        this.initialBondSize = new BN(20000000000);
-        this.lowerBoundDivisor = 2;
-        this.upperBoundMultiplier = 3;
-        this.contract = await BondSizeMock.new(
-            this.initialBondSize,
-            this.lowerBoundDivisor,
-            this.upperBoundMultiplier,
-        );
+    describe('in general...', () => {
```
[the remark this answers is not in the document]
boolafish: lol, my bad. rename it to `with normal cases`

PR #331, opened 2019-10-03
fix: remove superfluous space in the eip712 types

pdobacz:
well, as title.

Also a gentle nudge to maybe use (if exists) a sol/js lib for eip712, or roll some rudimentary tool like `elixir-omg` to at least somewhat reduce the chance of such errors.

### DIFF
--- a/plasma_core/utils/eip712_struct_hash.py
+++ b/plasma_core/utils/eip712_struct_hash.py
@@ -2,6 +2,11 @@ from eip712_structs import EIP712Struct, Address, Uint, Bytes, make_domain
 from eth_utils.crypto import keccak
 from plasma_core.constants import NULL_HASH, NULL_ADDRESS
 
+# TODO: if this is still required to be maintained it is missing a couple of items:
+# - outputGuard probably not output_guard in the Output type struct
+# - outputType missing in type and struct
+# - txType missing in type and struct
+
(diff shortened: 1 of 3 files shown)


boolafish:
> some rudimentary tool like elixir-omg

curious, what tool has been used in elixir-omg to avoid this?

kevsul:
omg-js lib builds the typehash dynamically:
https://github.com/omisego/omg-js/blob/master/packages/omg-js-util/src/signHash.js

We could do this in the tests which should catch any errors like this.

pdobacz:
@boolafish 

> curious, what tool has been used in elixir-omg to avoid this?

only the type hash preimages for types are auto generated from more structured specs, but that would cover spaces: https://github.com/omisego/elixir-omg/blob/master/apps/omg/lib/omg/typed_data_hash/types.ex

Maybe an alternative to have like a registry-repo holding those strings somewhere, we pull it in as a dep in every other repo that needs it and not repeat ourselves. (or that repo would hold a tool to generate, which all repos would like to use as a dependency, e.g. a non-py, non-js, non-ex one)

However then I wonder what would happen if that registry/tool broke. Maybe that would need to be caught by a test that tests Metamask conformance, which is kinda the reference impl here? :thinking:

Issue #304, opened 2019-09-25
Refactor controls for administrative roles to reflect role scheme 


thec00n:
## Issue Type 
```
[ ] bug report
[ X ] feature request
```

## Current Behavior
The following administrative roles are conceptually agreed upon but have not been implemented consistently across the contract system. 

- Authority: EOA used exclusively to submit plasma block hashes to the root chain. 
- Deployer: EOA used to deploy all root chain contracts. 
- Maintainer: MultiSig wallet that is allowed to initiate upgrades/updates of specific parts of the root chain contracts. Can also update the exit bond size. 
- Contract: A root chain contract that is authorized to call specific functions in other root chain contracts.

ref: #242 

## Suggested Fix

Suggestions are made as follows on a per contract basis. 

### Operated

- [ ] Rename the contract `Operated` to `Maintainable` in order to make it clear for which role it is used for.  

```
contract Maintainable {
    address public maintainer;

    constructor() public {
        maintainer = msg.sender;
    }

    modifier onlyMaintainer() {
        require(msg.sender == maintainer, "Can only be called by maintainer");
        _;
    }

}
```

### EthVault and Erc20Vault 

Todo: 
- [ ] Rename `Operated` to `Maintainable`
- [ ] Set Maintainer during contract creation 
- [ ] Use `onlyMaintainer` to control access to register new DepositVerifier 

### PlasmaFramework

Todo: 
- [ ] Rename `Operated` to `Maintainable`
- [x] The upgrade mechanism `registerExitGame` and `registerVault` for `VaultRegistry` and `ExitGameRegistry`  is inline with the main product requirements of being able to add components after `MIN_EXIT_PERIOD` has passed.  
- [ ] Instead of inlining the check if the function `submitBlock` is called from Authority create a modifier onlyAuthority() so it's aligned with the scheme of using modifiers for access checks. 
- [ ] The `initAuthority()` function is susceptible to front running and can be called by anyone after contract creation until it is called for the first time. I propose to remove `initAuthority()` and `setAuthority()` and set `authority` during contract creation. 

### PriorityQueue

Todo:
- [ ]  Remove the Ownable contract. 
- [ ] Introduce a new variable `plasmaFramework` that is set at contract creation.
- [ ] Create a new modifier `onlyFromPlasmaFramework` that restricts calling the functions `insert` and `delMin`. 

### PaymentExitGame

Todo: 
[...]

Issue #242, opened 2019-09-04
Separation deployer from maintainer addresses


pnowosie:
Continues on #233

## Terminology
- __Operator__ - organization, not an Ethereum address which holds other addresses
- __Authority__ - Ethereum address used exclusively to submit plasma blocks, requires `nonce = 0` as noted below.
- __Deployer__ - Ethereum address used to deploy Plasma framework for the first time
- __Maintainer__ - (aka __Operated.operator__) Ethereum address (can be the same as deployer or Multisig contract) that is allowed to register upgrades (tx types, output types, vaults, exit games, ...)

## Currently implemented
Separation `maintainer - authority` in #233
Maintainer address which can register `PlasmaFramework` components is passed as `msg.sender` during deployment. There were effort to separate `maintainer - deployer` addresses in #239 - but we found out that we need to untangle inheritance dependencies first (__TODO__: mention task :question: )

## Deployment process
Plasma Framework constructor accepts Maintainer address as a parameter. During deployment 
Deployer address sets Maintainer address. It could be the same address as Deployer for test networks or dedicated address (possibly multisig) responsible for registering components.

Note that: __Authority address is unset yet__. Deployer address can be throwaway address it isn't stored or verified anywhere in the contract.

Maintainer address register appropriate components (sets addresses of component contracts to the registries). Component contracts can be deployed by Deployer or Maintainer it does not matter, but only Maintainer can run `register...` functions on framework.

Authority address sends __its very first__ transaction that calls `BlockController.init()` function. This function sets a caller address as Authority address in `BlockController` which allows for block submission and sets a Authority address nonce - so the Plasma blocks numbers will correspond to the address nonces (`1000 * nonce = blknum`).

boolafish:
depends on whether the migration script goes first or not, but might need to change the migration script to reflect the decoupling of addresses.

boolafish:
ref #304

PR #333, opened 2019-10-04
242 separate maintainer address

boolafish:
### Note
- Replace `Operated` with `OnlyFromAddress`. This is done to separate `authority` from `deployer`. (see #242 and https://github.com/omisego/plasma-contracts/issues/304#issuecomment-538207923)
- Replace `initAuthority` function with `activateChildChain`. see this discussion: https://github.com/omisego/plasma-contracts/issues/304#issuecomment-538285330
- use interface `getMaintainer()` to decouple the maintenance address.
- use `onlyFrom(framework)` in `PriorityQueue`.

closes: #242 
closes: #304


thec00n:
@boolafish overall looks good:

- For PriorityQueue can we set the PlasmaFramework address through the constructor and use OnlyFromAddress to limit access. 
- For SpendingConditionRegistry/OutputGuardHandlerRegistry. Looking at this seems still odd. Until the registries can actually be updated I would suggest to remove the Ownable contract. Make registerSpendingCondition/registerOutputGuardHandler private for now and call them during contract creation.

thec00n:
`I don't like the way that we end up with 3 copies of the same authority address: PlasmaFramework, BlockController and ExitGameController. Same with maintainer address.`
I agree. 
We could use abstraction or we could have `authority` and `maintainer` in a separate contract and inject it into everything that needs it?

boolafish:
>  in a separate contract and inject it into everything that needs it?

Just a note to myself. @thec00n showed me the code what he means as "inject" is by inheritance with a `Inject` base contract that has the above two getters. And then `ExitGameController` can call the `getXXX()` since it is inheritance the base `inject` contract.

boolafish:
One idea bump to me just now on removing the inheritance issue rather easily. One thing make it a lot of work for us is that if we properly inject each contract (ExitGameController, BlockController) as a logic contract, we need to write extra code to expose public functions for old controller functions.

However, if we make our client calling  `plasma.blockController().submitBlock()` instead of `plasma.submitBlock()`, then this change can be done fairly easy. We just need to inject and expose the two controller contracts to PlasmaFramework instead of inheritance them. However, just throwing idea here as this breaks the interface to client more and make our client need to know more detail on our plasma framework components.

boolafish:
So only `maintainer` is used across contracts. `authority` is only used in `BlockController` so I did not use the new pattern for that.

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouter.sol
```diff
@@ -119,6 +121,8 @@ contract PaymentInFlightExitRouter is IExitProcessor, Operated, OnlyWithValue {
     )
         public
     {
+        maintainer = framework.getMaintainer();
```
thec00n: Can we consider using `framework.getMaintainer()` in `onlyFrom()` instead of creating another storage variable?

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
```diff
@@ -40,6 +40,8 @@ contract PaymentStandardExitRouter is
     PaymentChallengeStandardExit.Controller internal challengeStandardExitController;
     BondSize.Params internal startStandardExitBond;
 
+    address private maintainer;
```
thec00n: Can we consider using framework.getMaintainer() in onlyFrom() instead of creating another storage variable?

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -63,8 +78,8 @@ contract BlockController is Operated, VaultRegistry {
      * @dev see discussion in https://github.com/omisego/plasma-contracts/issues/233
      * @param _blockRoot Merkle root of the plasma block.
      */
-    function submitBlock(bytes32 _blockRoot) external {
-        require(msg.sender == authority, "Can be called only by the Authority.");
+    function submitBlock(bytes32 _blockRoot) external onlyFrom(authority) {
+        require(isChildChainActivated == true, "Child chain has not been activate by authority address yet");
```
thec00n: Do we need this check? I am ok if we make this implicit and enforce in child chain

thec00n: I guess it's safer ... :)

boolafish: I added this check in both `submitBlock` and `submitDepositBlock`. So is your proposal to enforce this implicit on child chain for `submitBlock` but keep the `submitDepositBlock`? And we are doing this to save some gas per block right?

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
boolafish: lol, just realized we are not saving `framework` in the `Router` to storage xdd. so...shall we replace the `maintainer` with `framework` instead XDD?

plasma_framework/contracts/src/framework/BlockController.sol
thec00n: Yeah I was thinking about gas ... but safer is always good.

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
thec00n: yeah

plasma_framework/contracts/src/framework/BlockController.sol
kevsul:
```suggestion
        require(isChildChainActivated == true, "Child chain has not been activated by authority address yet");
```

plasma_framework/contracts/src/framework/BlockController.sol
```diff
@@ -85,6 +100,7 @@ contract BlockController is Operated, VaultRegistry {
      * @return the deposit block number
      */
     function submitDepositBlock(bytes32 _blockRoot) public onlyFromNonQuarantinedVault returns (uint256) {
+        require(isChildChainActivated == true, "Child chain has not been activate by authority address yet");
```
kevsul:
```suggestion
        require(isChildChainActivated == true, "Child chain has not been activated by authority address yet");
```

plasma_framework/contracts/src/vaults/Vault.sol
```diff
@@ -2,16 +2,17 @@ pragma solidity 0.5.11;
...
 /**
  * @notice Base contract for vault implementation
  * @dev This is the functionality to swap "deposit verifier".
  *      By setting new deposit verifier, we can upgrade to a new deposit tx type without upgrading the vault.
  */
-contract Vault is Operated {
+contract Vault is OnlyFromAddress {
     event SetDepositVerifierCalled(address nextDepositVerifier);
     PlasmaFramework internal framework;
+    address internal maintainer;
```
kevsul: Is it necessary to store `maintainer`? We could just call `framework.getMaintainer()` instead when it's needed

PR #335, opened 2019-10-04
docs: add integration-doc.md

kevsul:
First stab at integration document.
Plenty still to be fleshed out, but wanted to get this out to make sure it's useful.


boolafish:
can you try `npm run docgen` with that folder under `./docs`? I am not sure whether `solidoc` would remove the existing folder first or just do an update.....

kevsul:
> can you try `npm run docgen` with that folder under `./docs`? I am not sure whether `solidoc` would remove the existing folder first or just do an update.....the new file might potentially be deleted by `solidoc`?

It doesn't delete it 😅 
In any case, I was going to move the generated docs to ./docs/contracts or something. Do you agree?

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -0,0 +1,283 @@
...
+Transactions that have been included in a block have a position which is the number of the block it's in and its index in that block. For example the fourth transaction in block 5000 has a position of `(5000, 3)`.
+
+The position of the outputs of a transaction can be obtained by including the index of the output in the transaction. So the position of the second output of the transaction in the above example would be `(5000, 3, 1)`.
+
+## Transaction type and output type
+
+The Abstract Layer Design introduces the concept of Transaction Type and Transcation Output Type. Each Transaction Type and Transcation Output Type can define different rules about how to spend funds.
+
+## Transaction format
+Even though there are different types of transaction and transaction outputs, all transactions have the same basic format:
```
boolafish: this might mislead that ALD actually requires all transactions have the same data structure which is something not really true and I would not want to limit.

I might consider to use the `Wire Transaction format` to make it easier to link with current `WireTransaction` library. Also, make it clear that this format is expected to support Payment + DEX related tx types (which is of our highest priority) but sadly not supporting ERC721.

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -0,0 +1,283 @@
...
+```
+    event InFlightExitOutputWithdrawn(
+        uint192 indexed exitId,
+        uint16 outputIndex
+    );
+```
+- When an exit was in the exit queue but was not processed (e.g. because it was already processed).
+```
+    event InFlightExitOmitted(
+        uint192 indexed exitId,
```
boolafish: exitId has moved to uint160

thec00n:
@kevsul @boolafish I am wondering if you guys think that documentation on the user/data flows is part of the integration doc. Let's take for example ERC20 deposits. 

1.) User approves the amount of ERC20 tokens that are about to be transferred. 
2.) User creates a RLP encoded transaction, that is decoded in `ERC20Vault`.
3.)`safeTransferFrom` is called which does the ownership transfer of the tokens
4.) The transaction is included in an empty block and Merkle root is used to create a new deposit block. 
5.) The `DepositCreated` event is triggered which does ...

I think documentation that surfaces interaction between the components contracts, child chain, watcher and client would be really useful.

pdobacz:
> 5.) The DepositCreated event is triggered. The child chain listens for this even and does ..
> 
> I think documentation that surfaces interaction between the components contracts, child chain, watcher and client would be really useful.

This is interesting, I'm very supportive of producing this kind of docs. Let us know if you need help with it, when it comes to chch/watcher.

I think, even if this content is written down somewhere, it might be quite dispersed, so there's surely a need for docs that describe this specifically. So far, we only have a general documentation of those interactions in [here](https://github.com/omisego/elixir-omg/blob/master/docs/tesuji_blockchain_design.md).

The question here is where do we draw the line between contracts documentation and child chain/watcher documentation.

My suggestion would be to write only general expectations (or better put - intention of how events and entry-points to the contracts should be used), and go into the specifics of "what happens" on the child chain/watcher docs' end.

E.g. contracts would only have:

> The DepositCreated event is triggered. This event can be listened to to make the deposited funds spendable on the child chain

vs on the chch/watcher:

> DepositCreated is recognised in the child chain after a deposit_finality_margin.
> After that the child chain will allow to spend the given UTXO in transactions, provided a correct utxo position (output id) is given along with a correct signature.
>
> DepositCreated is recognised in the watcher after a deposit_finality_margin. After that, the Watcher will treat transactions correctly spending the deposit as valid...

and so forth...

thec00n:
`My suggestion would be to write only general expectations (or better put - intention of how events and entry-points to the contracts should be used), and go into the specifics of "what happens" on the child chain/watcher docs' end.`

Makes sense also from a security perspective. Then people looking at the security of the system have some context when they look at the actual implementation and look for gaps. 

It's ok to have it in different places and I can see why the teams want to maintain their own. My only request is that they should overlay so if someone wants to understand the entire system they have an easier time with it.

pdobacz:
> they should overlay so if someone wants to understand the entire system they have an easier time with it.

Totally :+1:. I tried to make the example :point_up: overlay. We'll have to find the sweet spot of overlaid-but-not-excessively-redundant

kevsul:
Yes, I think this document should contain everything needed to interact with the contracts, whether you're a regular user, implementing a wallet or even implementing a watcher. 
Enough information to describe the apis, data structures, flows, etc. 
For super detailed explanations on the how and why it all works, you can go to the spec.

> The question here is where do we draw the line between contracts documentation and child chain/watcher documentation.

Do the child chain/watcher docs already exist @pdobacz ?

pdobacz:
@kevsul 

> Do the child chain/watcher docs already exist @pdobacz ?

There's some but dispersed, incomplete and not on the proper level of detail I guess. Docs like [this doc](https://github.com/omisego/elixir-omg/blob/master/docs/tesuji_blockchain_design.md) mentioned above or [this](https://github.com/omisego/elixir-omg/blob/master/docs/exit_validation.md) and comments throughout code.

cduangprom:
Gerhard has asked that I review this content from a language and layout perspective. I've created a google doc out of the content at https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/docs/integration-docs/integration-doc.md and will edit offline with suggested comments and then create a PR for the changes. Is that OK with everyone involved?

Issue #224, opened 2019-08-23
Defensive code when `processExits` calls function in each ExitGame


boolafish:
This is for discussion, want to get some feedback.

### Proposal
Current [code](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/framework/ExitGameController.sol#L122) calls the `processExit` function of ExitGame directly. Though the expected behavior is that user should exit while the ExitGame contract is quarantined if such ExitGame contract has invalid implementation. However, if there are case both us + auditor + user cannot catch a bug when `processExit` is called, and if such bug halt the transaction, now all the following valid exits inside the queue with valid ExitGame contract would be blocked.

To protect further valid exits, we can be defensive on calling back the ExitGame function. That is we use low level `call` and no matter it succeed or not we continue to do the next one. Of course, we can emit some event to trace the error if it ever happens.

In general, I am not a huge fan of defensive code but I feel it might worth putting this here.

Meantime, if we are doing this then there would be no need to have a queue per token anymore. See this previous issue discussion: https://github.com/omisego/plasma-contracts/issues/111. This removes the need of the step of `addToken` and almost introduce no extra UX tradeoff.

(a side discussion, btw, even if we are keeping this, is there better naming than `token`? Such naming seems wrong to me when we will be doing `ERC721` or other `ERC`.... as it is linked to the ERC721 contract address but not the token/assets in such contract...)

### Feedback Please : )
1. Does such defensive code make sense to do? Is it something we'd like to do or not?
1. If we are doing so, shall we just remove "token" concept from the queue?

boolafish:
> Meantime, if we are doing this then there would be no need to have a queue per token anymore.

This is a false statement as external contract can run out of gas. Using low level `call` and relying on the `success` flag of the result would not help. (did some experiments and proven not working)

boolafish:
There are some eyes on processExits already (eg. #385 ). this is more of an issue proposing an idea. closing this.

boolafish:
Just realize a potential issue of doing this, user can give some small gas forcing the low level call to the ExitGame contract to fail while just enough to do the remains job so the while transaction would still be considered successful. Then although the exit should be able to be processed would never be able to run again.

PR #336, opened 2019-10-07
fix: stop using transfer() and send() functions

### LINKED ISSUE #312: Stop using solidity's transfer now
## Issue Type 

```
[ ] bug report
[x] feature request
```

## Current Behavior
using `transfer` everywhere

## Expected Behavior
not using it?

## Suggested Fix
N/A

## Motivation for Change
https://diligence.consensys.net/posts/2019/09/stop-using-soliditys-transfer-now/

pgebal:
Gets rid off `transfer` and `send` calls in favor of `call.value`.
Closes #312.

Introduces two ways of guarding against reentrancy:
1. `ReentrancyGuard.sol` from open-zeppelin, which reverts the top most call and in our case is usable for preventing reentrancy on challenges.
2. `GracefulReentrancyGuard.sol`, which prevents reentrancy but not fails the top most call, we need that, so failed fund transfers do not block exit queues. Used in `EthVault` and for processing standard and in-flight exits.


plasma_framework/contracts/mocks/attackers/PaymentIFEInputChallengeAttacker.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    constructor(PaymentInFlightExitRouter _router, PaymentInFlightExitRouterArgs.ChallengeInputSpentArgs memory _challengeArgs) public {
+        router = _router;
+        challengeArgs = _challengeArgs;
+    }
+
+    /* solhint-disable no-complex-fallback */
+    function () external payable {
+        if (funded) {
+            router.challengeInFlightExitInputSpent(challengeArgs);
```
pgebal: All Payment*Attacker contracts could probably be replaced by a single contract that takes a generic address and encoded function selector with args. Here we would use `call`. I couldn't find a handy way to encode those structs on js side, so I proceeded with just copying the contracts. That's something that could be fixed in a follow-up PR.

plasma_framework/contracts/src/vaults/Erc20Vault.sol
```diff
@@ -49,7 +50,7 @@ contract Erc20Vault is Vault {
     * @param token address of ERC20 token contract.
     * @param amount amount to transfer.
     */
-    function withdraw(address payable receiver, address token, uint256 amount) external onlyFromNonQuarantinedExitGame {
+    function withdraw(address payable receiver, address token, uint256 amount) external onlyFromNonQuarantinedExitGame gracefullyNonReentrant {
```
pgebal: This could possibly be just `nonReentrant` as if token transfer fails then most probably token contract is broken and we won't be able to process exitQueue for that token anyway.

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -68,7 +73,12 @@ library PaymentProcessInFlightExit {
```
thec00n: `PaymentProcessInFlightExit.run()` has several places (basically all `call`s) that would allow for reentrancy. We should refactor the function so it follows Checks-Effects-Interactions pattern -> https://solidity.readthedocs.io/en/develop/security-considerations.html?highlight=check%20effects#use-the-checks-effects-interactions-pattern

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -68,7 +73,12 @@ library PaymentProcessInFlightExit {
 
                 if (shouldWithdrawInput(exit, withdrawal, token, i)) {
                     withdrawFromVault(self, withdrawal);
-                    withdrawal.exitTarget.transfer(withdrawal.piggybackBondSize);
+                    // solhint-disable-next-line avoid-call-value
+                    (bool success, ) = withdrawal.exitTarget.call.value(withdrawal.piggybackBondSize)("");
+                    // we do not want to block a queue if bond return is unsuccessful
+                    if (!success) {
+                        emit InFlightBondReturnFailed(withdrawal.exitTarget, withdrawal.piggybackBondSize);
```
thec00n: Does that mean that the funds get stuck in the contract forever?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -78,7 +88,12 @@ library PaymentProcessInFlightExit {
 
                 if (shouldWithdrawOutput(self, exit, withdrawal, token, i)) {
                     withdrawFromVault(self, withdrawal);
-                    withdrawal.exitTarget.transfer(withdrawal.piggybackBondSize);
+                    // solhint-disable-next-line avoid-call-value
+                    (bool success, ) = withdrawal.exitTarget.call.value(withdrawal.piggybackBondSize)("");
+                    // we do not want to block a queue if bond return is unsuccessful
+                    if (!success) {
+                        emit InFlightBondReturnFailed(withdrawal.exitTarget, withdrawal.piggybackBondSize);
```
thec00n: Does that mean that the funds get stuck in the contract forever?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -89,7 +104,12 @@ library PaymentProcessInFlightExit {
         clearPiggybackOutputFlag(exit, token);
 
         if (allPiggybacksCleared(exit)) {
-            exit.bondOwner.transfer(exit.bondSize);
+            // solhint-disable-next-line avoid-call-value
+            (bool success, ) = exit.bondOwner.call.value(exit.bondSize)("");
+            // we do not want to block a queue if bond return is unsuccessful
+            if (!success) {
+                emit InFlightBondReturnFailed(exit.bondOwner, exit.bondSize);
```
thec00n: Does that mean that the funds get stuck in the contract forever?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessStandardExit.sol
```diff
@@ -22,6 +22,11 @@ library PaymentProcessStandardExit {
         uint160 indexed exitId
     );
```
thec00n: We should refactor the function so it follows Checks-Effects-Interactions pattern -> https://github.com/omisego/plasma-contracts/blob/eb53c85aa821f53f49a2fc2d7b31aa548a219766/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessStandardExit.sol#L69

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
pgebal: yes

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
pgebal: yes

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
pgebal: yes

boolafish:
High level wise: I don't know my proposal is a good idea or not, but potentially we can be defensive in the `processExits` of `ExitGameController` instead of every places.

Meanwhile, for the case low level `call` is not successful, shall we use `assert` instead (if we be defensive on the `ExitGameController`). This makes our intention more obvious that we are not expecting this to happen.

see: https://github.com/omisego/plasma-contracts/issues/224

boolafish:
BTW, seeing the reentrancy attack just let me realize an user can attack by draining out the gas. This would not be protected by the `success` flag of `call`. So we can be attacked by running out of gas and stop the queue from processing further on.

@thec00n

thec00n:
hey @boolafish @pgebal I think the way to go is:

- Use a reentrancy guard for `PaymentExitGame` and all the exit related functions such that it is not possible for a caller of a public/external exit game function to have a callee that is also an exit game related function. 
- Use the Checks-Effects-Interactions pattern consistently across functions that have an external call.

boolafish:
Note for some changes that I added:
1. rename `GracefulReentrancyGuard` to `FailFasteentrancyGuard`. Same code but different purpose of using it.
1. Add `nonReentrant` in exit game interactive public functions. However, left out the process exit as that need a bit more investigation on how to capture reentrant from the `PlasmaFramework` level.
1. Change most reentrant fail test to test by forcing fallback function revert instead. I think it was trying to test two kind of stuff at once.
1. Have another test suite for reentrant protection. Have a test helper function to mimic the action.

plasma_framework/contracts/mocks/attackers/FailFastReentrancyGuardAttacker.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "../../src/utils/FailFastReentrancyGuard.sol";
+
+contract FailFastReentrancyGuardAttacker is FailFastReentrancyGuard {
+
+    event RemoteCallFailed();
+
+    function guardedLocal() public nonReentrant {
```
thec00n: What does this function do?

boolafish: My understanding is that both functions here are just to demonstrate that it would fail

thec00n: Wouldn't this function call itself indefinitely until it runs out of gas?

boolafish: In this implementation of `FailFastReentrancyGuard` it would not. That's a bit like why I choose to use our own one instead of the one from openzepplin so it would fail directly instead of running the function first and then fail.

https://github.com/omisego/plasma-contracts/blob/d9c61822a5359106d87891e87cfa3795e5c80b52/plasma_framework/test/src/utils/FailFastReentrancyGuard.test.js#L10

thec00n: Ok that makes sense.

plasma_framework/contracts/mocks/exits/payment/routers/PaymentInFlightExitRouterMock.sol
```diff
@@ -79,6 +89,32 @@ contract PaymentInFlightExitRouterMock is PaymentInFlightExitRouter {
         framework.flagOutputSpent(_outputId);
     }
 
+    /**
+     * This function helps test reentrant by making this function itself the first call with 'nonReentrant' protection
+     * So all other PaymentExitGame functions that is protected by 'nonReentrant' too would fail as it would be considered as re-entrancy
```
thec00n: So all other PaymentExitGame functions that are protected by 'nonReentrant' too would fail as it would be considered as re-entrancy

plasma_framework/contracts/mocks/exits/payment/routers/PaymentStandardExitRouterMock.sol
```diff
@@ -43,4 +47,20 @@ contract PaymentStandardExitRouterMock is PaymentStandardExitRouter {
     }
 
     function depositFundForTest() public payable {}
+
+    /**
+     * This function helps test reentrant by making this function itself the first call with 'nonReentrant' protection
+     * So all other PaymentExitGame functions that is protected by 'nonReentrant' too would fail as it would be considered as re-entrancy
```
thec00n: So all other PaymentExitGame functions that are protected by 'nonReentrant' too would fail as it would be considered as re-entrancy

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -34,6 +34,11 @@ library PaymentProcessInFlightExit {
         uint16 inputIndex
     );
 
+    event InFlightBondReturnFailed(
```
thec00n: Why do we do this only for inflight bonds?

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity 0.5.11;
+
+/**
+ * @notice Reentrancy guard that fails immediately when a reentrace occurs
+ * @dev Instead of using the one from openzepplin for the following reasons:
+ *      1. Easier to understand
+ *      2. Fail fast when reentracy occurs make testing easier
```
thec00n: 2. Fail fast when reentracy occurs makes testing easier

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity 0.5.11;
+
+/**
+ * @notice Reentrancy guard that fails immediately when a reentrace occurs
+ * @dev Instead of using the one from openzepplin for the following reasons:
```
thec00n: @dev We use a simple mutex instead of https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuard.sol for the following reasons:

plasma_framework/contracts/mocks/attackers/FallbackFunctionFailAttacker.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "../../src/exits/payment/routers/PaymentInFlightExitRouter.sol";
```
pgebal: We don't use those imports.

plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -417,53 +416,6 @@ contract('ExitGameController', () => {
                 });
             });
         });
-
-        describe('When reentrancy happens', () => {
```
boolafish: Can we update to the present the new protection schema?

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
```diff
@@ -1,21 +1,21 @@
 pragma solidity 0.5.11;
 
+import "../framework/PlasmaFramework.sol";
+
 /**
```
boolafish: Might need to revisit this documentation + consider rename (probably don't need) ?

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -43,6 +44,25 @@ contract ExitGameController is ExitGameRegistry {
...
+
+    function unlock() external onlyFromNonQuarantinedExitGame() {
+        mutex = false;
+    }
+
+    /**
+     * @dev Prevents reentrant calls by using a mutex.
+     */
+    modifier nonReentrant() {
+        require(!mutex, "Reentrant call");
```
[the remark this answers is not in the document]
pgebal: Not really, 'lock' and 'unlock' are accessible only by exit games.
But I got rid off this modifier and I just put locking and unlocking into 'processExits' body.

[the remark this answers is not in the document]
boolafish: ah....now I see why

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -43,6 +44,25 @@ contract ExitGameController is ExitGameRegistry {
     {
     }
 
+    function lock() external onlyFromNonQuarantinedExitGame() {
```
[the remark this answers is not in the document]
thec00n: Maybe `activateNonReentrant` and ``deactivateNonReentrant`

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
pgebal: We do it for standard exits. Check 'PaymentProcessStandardExit.sol'.

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
pgebal: done

plasma_framework/contracts/mocks/attackers/FallbackFunctionFailAttacker.sol
pgebal: done

plasma_framework/contracts/src/framework/ExitGameController.sol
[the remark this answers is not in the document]
pgebal: nice names, done

plasma_framework/contracts/src/framework/ExitGameController.sol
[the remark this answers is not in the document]
pgebal: Actaully I've kept this modifer for code clarity reasons.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -43,6 +44,25 @@ contract ExitGameController is ExitGameRegistry {
...
+        require(!mutex, "Reentrant call");
+        mutex = true;
+    }
+
+    function unlock() external onlyFromNonQuarantinedExitGame() {
```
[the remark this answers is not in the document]
pgebal: done

plasma_framework/contracts/src/utils/FailFastReentrancyGuard.sol
pgebal: done

plasma_framework/test/src/framework/ExitGameController.test.js
pgebal: done

paulperegud:
> BTW, seeing the reentrancy attack just let me realize an user can attack by draining out the gas instead of reentrance? This would not be protected by the `success` flag of `call`. So we can be attacked by running out of gas and stop the queue from processing further on. (btw, I am not 100% is this possible :p)

If tx is out of gas, it reverts all of it's changes to Ethereum state, with exception of burned gas and burned account nonce. I don't really see how being out of gas can be exploited with regards to reentrance.

boolafish:
Ah I skip quite a lot of info but basically the concern is that any failure would block the loop for processExits. And this is not protected by isolation of exit queue.

paulperegud:
Are we still in a world where we have 3 groups of contracts?
1. our own code, trusted
2. token code, untrusted, execution related to particular token is isolated by their own exit queue
3. all other code, never called.

If we are in group 1 - we should be able to know how much gas our calls may have. I'm a bit lost with the issue above.

boolafish:
I don’t think it falls to any of your three. It is when I put a contract as the owner of plasma transaction, and when doing transfer/call, the fallback function would be triggered just as how Reentrancy attack could happen

thec00n:
I missed the comments from @boolafish last week. I had a look at this and I think he is right, here are my thoughts https://github.com/omisego/plasma-contracts/issues/385.

Issue #270, opened 2019-09-13
Event emit in library might not be trackable by client


boolafish:
### Why
Current `controllers` code emit event in the library. However, as it is not a real contract that has storage, the event is emit on the behalf of the original contract using the library. But, the original contract might not have the event abi, as a result, the client might not be able to subscribe to the event by name.

solution and reasoning see this: 
1. https://ethereum.stackexchange.com/questions/11137/watching-events-defined-in-libraries
1. https://blog.aragon.org/library-driven-development-in-solidity-2bebcaf88736/  (`Events and libraries` part)

Note: I tested in tests, without doing the above solution it cannot listen to the event via `getPastEvent` on the contract directly.

PR #337, opened 2019-10-07
fix: copy lib events into linked contracts

kevsul:
# Description
 - Copy library events into the linked contracts (*Routers) so that they appear in the contract's abi and so are searchable. See https://blog.aragon.org/library-driven-development-in-solidity-2bebcaf88736/

 - Changed the tests to use `expectEvent.inLogs` instead of `expectEvent.inTransaction`, this makes sure that the events are actually in the logs, and so can be found later by e.g `getPastEvents()`


kevsul:
Closes #270

boolafish:
LGTM. 
Curious is there a way that we can make sure we don't fall into this problem again? Like never use `inTransaction` in tests?

PR #339, opened 2019-10-08
chore: replace RLP.sol with RLPReader.sol

### LINKED ISSUE #317: Multiple overflows/assert violations in RLP decoding library
## Issue Type 

```
[ x ] bug report
[ ] feature request
```

## Current Behavior
The RLP decoding library is foundational component in the root chain contracts as it processes RLP encoded user input in various places and converts them into Solidity native structures. An exploitable vulnerability in the RLP library could have critical consequences for the root chain contracts and potentially put user funds at risk. 

## Steps to Reproduce
Overflows/assert violations shared in Slack. 

## Suggested Fix
The current RLP encoding library has poor test coverage and it uses Solidity assembly extensively which is undesirable from a security perspective. Also there does not seem to be a well tested and audited Solidity RLP encoding library out there except maybe https://github.com/hamdiallam/Solidity-RLP.

 It is recommended to either:

- write the current RLP library from scratch. Use assembly only in places where a Solidity implementation is impossible or outrageously more gas heavy. 
-  Find an existing implementation that has better code quality and that does not rely on assembly as much. 

In any case an RLP encoding library that is production grade needs to be well tested and needs to work according to RLP specifications. Existing test suites from other languages should be leveraged and should all pass for the new implementation:

- https://github.com/ethereum/tests/tree/master/RLPTests
- https://github.com/ethereumjs/rlp/tree/master/test

kevsul:
Closes #317


kevsul:
@thec00n the issue (#317 ) refers to overflows/assert violations in the previous RLP implementation.
Can you test this new implementation, or point me to how to do it?

thec00n:
@kevsul following up here with my current thoughts.

- from a security perspective I would like for any RLP decoding function to check against the spec and revert if it’s invalid before it does anything else.
- Once we have the valid (https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json)/invalid (https://github.com/ethereum/tests/blob/develop/RLPTests/invalidRLPTest.json) data set passing for the smart contracts I would like to ask the elixir-omg team to make sure they pass on their side as well to make sure decoding is consistent.
- The assembly code is limited to mostly read X number of bytes from a dynamic bytes array. One thing that concerns me is when assembly is used you lose the safety harness of solidity. In this case this means that if there is an error in calculating the length then you could read outside of the bytes array that you pass to the function. Solidity does add an assert for this case. If we stick with assembly then we need to add bounds checks that might be missing. I think a function that chops N bytes of bytes[] would probably cover most cases in the lib, not sure what that means in terms of gas increase?
- Can we remove any functions that we currently don't use. This will make code review and comparing test coverage much easier.

PR #345, opened 2019-10-11
fix: do not revert PaymentTransactionStateTransitionVerifier.isCorrec…

### LINKED ISSUE #342: `PaymentExitGame.startInFlightExit` OOGs when #outputs > #inputs
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior

When attempting to start an ife from a transaction that has #outputs > #inputs the transaction seems to be able to consume any amount of gas (when I try with 6M gas allowance it reverts consuming 5820349). Transactions that have #outputs <= #inputs work just as expected.

I haven't been able to have a proper isolated reproduction of this, but judging by Watcher's integration tests from [this revision](https://github.com/omisego/elixir-omg/commit/7ddb42b1ca011277ef14d6ad7110e6f30f1c6196), see lines: https://github.com/omisego/elixir-omg/blob/7ddb42b1ca011277ef14d6ad7110e6f30f1c6196/apps/omg_watcher/test/omg_watcher/integration/in_flight_exit_test.exs#L164-L179

Notice that `tx1` has two outputs whereas `tx2` only one. The IFE for `tx2` is started but for `tx1` is not

The data that's being sent to the contract seems healthy: https://gist.github.com/pdobacz/e0f3b872c18215d9542164ecf84e103c

## Expected Behavior

start ife should work in all combinations of ins/outs. The loop that could cause the OOG should not be possible.

## Steps to Reproduce

WIP, unless you're able to run the watcher tests, then:

```
mix test --only integration test/omg_watcher/integration/in_flight_exit_test.exs --seed 0 --trace
```

on revision cited above.

## System Specs

- python version: N/A
- pyetherem version: N/A
- environment (output of running `pip freeze`): N/A
- operating system: Linux
- solc: 0.5.11
- geth: 1.8.27

pgebal:
Closes #342


plasma_framework/test/src/exits/payment/PaymentTransactionStateTransitionVerifier.test.js
```diff
@@ -106,7 +107,28 @@ contract('PaymentTransactionStateTransitionVerifier', ([alice, bob]) => {
...
             const verificationResult = await this.verifier.isCorrectStateTransition(
-                args.inFlightTx,
+                args.inFlightTxRaw,
+                args.inputTxs,
+                args.outputIndexOfInputTxs,
+            );
+            expect(verificationResult).to.be.true;
+        });
+
+        it('should not revert when number of transaction outputs is greater then number of transaction inputs', async () => {
```
pdobacz: I saw the title of the PR and was somewhat confused. I'd suggest to rename things (PR and test here) more towards the behavior that we're expecting to see, along the lines of:

```
allows to start IFE with different configurations of inputs/outputs in teh transaction
```

where you would test a couple of situations (or all 25 if you prefer 0-4ins x 0-4outs :) )

PR #346, opened 2019-10-15
Minor edit

cduangprom:


plasma_framework/contracts/src/exits/interfaces/IOutputGuardHandler.sol
```diff
@@ -4,19 +4,19 @@ pragma experimental ABIEncoderV2;
...
- *      the privacy deposit mechanism and becomes hash of output type and preimage.
+ * @notice An interface for utils functions, required for processing and retrieving essential data from the `output guard` field.
+ * @dev This is required because there are multiple ways of using the `output guard` field. For example, in normal
+ *      payments the `output guard` field holds the owner's address directly, while for DEX deposits the `output guard` field uses
+ *      the privacy deposit mechanism and becomes the hash of output type and preimage.
  */
 interface IOutputGuardHandler {
     /**
-    * @notice Checks a given output guard data
+    * @notice Checks a given output guard data.
```
kasima: I'd like to follow NatSpec convention here and not use `.` to close these phrases. Here's the NatSpec [example](https://solidity.readthedocs.io/en/v0.5.10/natspec-format.html#documentation-example). 0x also [doesn't use punctuation](https://github.com/0xProject/0x-monorepo/blob/development/contracts/erc721/contracts/src/ERC721Token.sol) in their NatSpec for tags like `@param`. Of course, for multi-sentence sections in a `@notice` or `@dev` tag, punctuation makes sense.

PR #348, opened 2019-10-15
Minor edit

cduangprom:


cduangprom:
Moving the change into PR 358

PR #351, opened 2019-10-15
Minor edit

cduangprom:


cduangprom:
Redoing the change into branch patch 15 and PR 358

PR #352, opened 2019-10-15
Minor edit

cduangprom:


cduangprom:
Redoing this change into PR 358 patch 14 branch

PR #354, opened 2019-10-15
Minor edits

cduangprom:
* Not sure if 'The storage of all in-flight exit data.' should be 'Stores all in-flight exit data
* Not sure if "Challenging transaction does not spent the output" should be "Challenging transaction does not spend the output"


cduangprom:
Redoing the change into PR 358 patch-14 branch

PR #355, opened 2019-10-15
Minor edit

cduangprom:


cduangprom:
Redoing change into PR 358 and patch-14 branch

PR #356, opened 2019-10-15
Edit file

cduangprom:
Not clear what this line means:  "Exit Target for outputs is set in piggyback instead of start in-flight exit."
Is it this? "Exit Target for outputs is set in the piggyback instead of when starting an in-flight exit."

Also find this in several files: The storage of all in-flight exit data.
Should it be: "Stores all in-flight exit data."?


cduangprom:
Redoing this change into patch-14 branch

PR #360, opened 2019-10-15
344 fix flagging outputs spent

### LINKED ISSUE #344: Zero inputs/outputs get through to framework.batchFlagOutputsSpent
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior

When an IFE is processed, it is doing `framework.batchFlagOutputsSpent` as part of the process ([here](https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol#L208)). However `""` `outputIds` are selected, due to the way the list of outputIDs are prepared **for Eth** where `token == 0x000...`.

As a result any non 4in/4out tx IFE finalizing **in Eth** will cause an error: `Should not flag with empty outputId` thrown here: https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/contracts/src/framework/ExitGameController.sol#L170

## Expected Behavior

This finalization should work.

Also consider turning the throwing `revert` into an `assert` - this is a condition that should ever be met.

## Steps to Reproduce

Using the watcher integration tests:

```
mix test --only integration test/omg_watcher/integration/in_flight_exit_test.exs:337
```

as of https://github.com/omisego/elixir-omg/commit/c32071d5c4f4b7c06f507a6fefd5245406b1f1aa (taking workarounds for #342, otherwise you won't reach the line executing the processing)

## Suggested Fix

Add `exit.outputs[i].outputId != "" && ` (or respective alternative for `inputs`) to the `if` that's used when scanning the inputs/outputs. 

**But not sure if this is OK**, rethink carefully! The fact that zero-data is held in static arrays `inputs` `outputs` in an ife seems to be the root cause of the problem - can't these arrays be dynamic instead?

pgebal:
Closes #344 
Implements a fix suggested in the issue description.


plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -173,12 +173,12 @@ library PaymentProcessInFlightExit {
...
-        // if exiting from output, all inputs are consider spent and can only exit from output furthre on.
-        // if exiting from input, for simplicity, we force all users to piggyback the input together at once
-        // instead of re-start the IFE and re-piggyback their input.
+        // if exiting from output, all inputs are consider spent and can only exit from output further on.
+        // if exiting from input, for simplicity, we force users to piggyback inputs together
+        // instead of restarting the IFE and then piggybacking their input.
         uint256 inputNumOfTheToken;
         for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token) {
+            if (exit.inputs[i].token == token && exit.inputs[i].outputId != bytes32(0)) {
```
boolafish: potentially we can have a function `isEmptyInput` in `PaymentInFlightExitModelUtils`. Meanwhile, would it be better to use amount > 0 instead? I remember we have a case that we put in the WithdrawData without `outputId` and set it in `piggyback` so with outputId or not is not so "empty". But I might be in the wrong impression that we are checking empty. 

anyway, this might only make it less clear why we are doing this so I will leave this to you.

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
```diff
@@ -514,5 +518,82 @@ contract('PaymentExitGame - End to End Tests', ([_, richFather, bob]) => {
                 });
             });
         });
+
+        describe('should process in-flight exits', () => {
```
boolafish: I might follow the style with other tests, like
```
Given Alice deposited with ETH
   Given a in-flight tx not mined
        The in-flight tx should be able to in-flight exit
             And then the output owner piggybacks
                  And then someone processes the exits for ETH after a week
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
pgebal: Good catch with `amount` instead of `outputId`. I decided not to introduce `isEmptyInput`. If you think we need it, let me know and I'll code it.

plasma_framework/test/src/exits/payment/PaymentExitGame.test.js
pgebal: done

PR #362, opened 2019-10-16
216 ife with invalid priority

### LINKED ISSUE #216: IFE with deposit tx can have wrong priority
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior
Let's say we have the following three blocks, two consecutive child chain block and another deposit block:
[valid_tx] [invalid_tx] [deposit_tx]

Now a user A owns the `valid_tx` while invalid operator owns `invalid_tx` and `deposit_tx`. The child chain block with `invalid_tx` is withhold by the operator however, user A verified the block with `valid_tx` already and decide to do another valid tx and send to operator.

Now, operator do an IFE with `invalid_tx` and `deposit_tx` as input, and the youngest utxoPos would be `deposit_tx`. The problem is that a deposit tx have 1 week higher priority then normal plasma tx. As a result, even user A tries to IFE with the priority of `valid_tx`, it will still lose the priority over the `deposit_tx` that hides the `invalid_tx`.

## Expected Behavior
Should exit with right priority.

## Suggested Fix
hmm, check the "latest exit time" instead of "youngest utxopos" of inputs?

## Motivation for Change
Fix bug

pik694:
We made sure that the exitable timestamp is not computed based on the youngest input, but we take the furthest exitable timestamp of all inputs.

The issue still persists in the  RootChain contract.

Closes #216

### DIFF
--- a/plasma_framework/contracts/src/exits/utils/ExitableTimestamp.sol
+++ b/plasma_framework/contracts/src/exits/utils/ExitableTimestamp.sol
@@ -7,21 +7,28 @@ library ExitableTimestamp {
         uint256 minExitPeriod;
     }
 
-    function calculate(
+    function calculateDepositTxOutputExitableTimestamp(
         Calculator memory _calculator,
-        uint256 _now,
-        uint256 _blockTimestamp,
-        bool _isDeposit
+        uint256 _now
     )
         internal
         pure
         returns (uint64)
     {
-        uint256 minExitableTimestamp = _now + _calculator.minExitPeriod;
+        // Please that boosting the exitable timestamp for a deposit should be only done in case of a SE.
+        // For the explanation please refer to: https://github.com/omisego/plasma-contracts/issues/216
+        return uint64(_now + _calculator.minExitPeriod);
+    }
 
-        if (_isDeposit) {
-            return uint64(minExitableTimestamp);
-        }
-        return uint64(Math.max(_blockTimestamp + (_calculator.minExitPeriod * 2), minExitableTimestamp));
+    function calculateTxExitableTimestamp(
+        Calculator memory _calculator,
+        uint256 _now,
+        uint256 _blockTimestamp
+    )
+        internal 
+        pure
+        returns (uint64)
+    {
+        return uint64(Math.max(_blockTimestamp + (_calculator.minExitPeriod * 2), _now + _calculator.minExitPeriod));
     }
 }
--- a/plasma_framework/contracts/src/exits/utils/ExitableTimestamp.sol
+++ b/plasma_framework/contracts/src/exits/utils/ExitableTimestamp.sol
@@ -7,6 +7,31 @@ library ExitableTimestamp {
         uint256 minExitPeriod;
     }
 
+    /**
+     * @notice Calculates the exitable timestamp for a mined transaction
+     * @dev This is the main function when asking for exitable timestamp in most cases.
+     *      The only exception is to calculate the exitable timestamp for a deposit output in standard exit.
+     *      Should use the function 'calculateDepositTxOutputExitableTimestamp' for that case.
+     */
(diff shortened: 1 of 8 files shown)


plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -361,6 +362,28 @@ library PaymentStartInFlightExit {
         return youngest;
     }
 
+    function getFurthestExitableTimestamp(StartExitData memory startExitData)
```
boolafish: the function `getYoungestInputUtxoPosition` and this `getFurthestExitableTimestamp` must be pointing to same input. I think currently we can potentially having a youngest position of an input and the furthest excitable timestamp to another input?

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -361,6 +362,28 @@ library PaymentStartInFlightExit {
...
+        private view
+        returns (uint64)
+    {
+        ExitableTimestamp.Calculator memory calculator = startExitData.controller.exitTimestampCalculator;
+        BlockController blockController = startExitData.controller.framework;
```
boolafish: interesting it does not need to cast!

pik694: Solidity 😂

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
pik694: It does not have to point to the same input - youngest input != furthest exitable timestamp.
When we were taking the exitable timestamp of the youngest input, we introduced a critical bug as described in #216. Now, my proposal is to use the last exitable timestamp (compute exitable timestamps of all inputs and choose one) and then still find the youngest input.

boolafish: I am a bit confused now (oops) I thought your proposal is to do what described in [this comment](https://github.com/omisego/plasma-contracts/issues/216#issuecomment-523399915) while we discussed offline.

I found a scenario current implementation seems wrong:

[valid_tx_1, valid_tx_2, invalid_tx, ...] -> [deposit_tx]....

Now operator IFE with input of `invalid_tx` and `deposit_tx`, the `furthestExitableTimestamp` would be of `invalid_tx` while the `youngest_utxo_pos` would be of `deposit_tx`. However, when we enqueue, we are only using the `txPos` instead of `utxoPos` which would be 0 for `deposit_tx` and make this IFE exiting with the same priority of `valid_tx_1` (since txPos would become 0). now `valid_tx_2` would have lower priority.....

pik694: actually `tx pos` consists of `blknum` and `tx index`, which means that the priorities are still good.

plasma_framework/contracts/src/exits/utils/ExitableTimestamp.sol
```diff
@@ -7,21 +7,28 @@ library ExitableTimestamp {
...
-        uint256 _blockTimestamp,
-        bool _isDeposit
+        uint256 _now
     )
         internal
         pure
         returns (uint64)
     {
-        uint256 minExitableTimestamp = _now + _calculator.minExitPeriod;
+        // Please that boosting the exitable timestamp for a deposit should be only done in case of a SE.
```
boolafish: Can you put this on top of the function as the function doc instead?
Also, would be better to link to Kevin's comment directly in that issue.

Issue #323, opened 2019-10-01
Solve migration script issues


boolafish:
These are some remaining tasks from #298 that needs further discussion to decide how to do.

1. Where to output? Currently it outputs to console directly, however, it would clog the screen + every `truffle test` would trigger the migration script too. Better to write it to file or output to somewhere. https://github.com/omisego/plasma-contracts/pull/298#discussion_r328931562

1. Replace usage of `global`. The migration script should be designed to be append-able. However, our usage of global makes those var only stays in memory. We need to figure our better way to share variable across scripts. https://github.com/omisego/plasma-contracts/pull/298#issuecomment-535812805

boolafish:
Something might also be worth consider is to run our E2E test with our migration script. That is, we just get our deployed contracts and start to run those instead of `new`-ing in the e2e test.

boolafish:
tag @jbunce and @kendricktan for inputs. I think we need help from your side to see where to output said from console.

kendricktan:
We should also take into account how to run the migrations via `vault-ethereum`

boolafish:
After discussion with @kendricktan , we are progressing this with:
1. No change with the output issue. Currently it is still bearable to from the console output.
1. Remove the `global` by just getting the address from the `accounts` each time. That is the only place we are using `global`. In the future, we are planning to move to vault service + multi sig with `maintainer address`. However, how that works together is still unknown and it is not possible to abstract things out now. we would just leave it for future change.

PR #363, opened 2019-10-16
Removed globals and fixed truffle-config provider

kendricktan:
Issue: https://github.com/omisego/plasma-contracts/issues/323

Fixes:
1. Fixed `truffle-config.js` to reflect 3 account addresses
2. Removed global variables, and always read from `accounts` (injected by truffle)

This fixes the following problem (defined in the issue linked)
> Replace usage of global. The migration script should be designed to be append-able. However, our usage of global makes those var only stays in memory. We need to figure our better way to share variable across scripts. #298 (comment)


plasma_framework/migrations/2_deploy_plasma_framework.js
```diff
@@ -2,16 +2,21 @@ const PlasmaFramework = artifacts.require('PlasmaFramework');
 
 const config = require('./config.js');
 
-module.exports = async (deployer) => {
+module.exports = async (
+    deployer,
+    _,
+    // eslint-disable-next-line no-unused-vars
```
boolafish: can we use `_` if the variable is not used to get pass through the linter instead?

boolafish:
shall we also sneak the change of making exit period 10 min for develop env?
https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/migrations/config.js#L7

plasma_framework/migrations/2_deploy_plasma_framework.js
kendricktan: Unfortunately the linter doesn't support multiple `_` :|

PR #370, opened 2019-10-18
fix: output of contracts

InoMurko:
Branched out of `7b8a2643568556c1d126749724666bc37edc8141` because that's what we use in elixir-omg now. Will delete later when we bump them.


InoMurko:
we've moved on to head of master!

PR #396, opened 2019-10-28
376 rlp improvements

thec00n:
- add documentation 
- stricter bounds checking 
- More negative test cases


plasma_framework/contracts/mocks/utils/RLPMock.sol
```diff
@@ -17,14 +17,14 @@ contract RLPMock {
...
-    function decodeInt(bytes memory _data) public pure returns (int) {
-        return int(_data.toRlpItem().toUint());
-    }
-
     function decodeUint(bytes memory _data) public pure returns (uint) {
         return _data.toRlpItem().toUint();
     }
 
+    function decodeInt(bytes memory _data) public pure returns (int) {
+         return int(_data.toRlpItem().toUint());
```
kevsul:
```suggestion
        return int(_data.toRlpItem().toUint());
```

thec00n:
@kevsul thanks for looking into the PR. Am still working on it, should finish soon.

boolafish:
Is this ready for a second review? @thec00n

thec00n:
@boolafish yes please

plasma_framework/contracts/mocks/utils/RLPMock.sol
```diff
@@ -17,29 +19,96 @@ contract RLPMock {
...
         for (uint i = 0; i < items.length; i++) {
-            result[i] = items[i].toRlpBytes();
+            result[i] = toRlpBytes(items[i]);
         }
         return result;
     }
 
-    function decodeString(bytes memory _data) public pure returns (string memory) {
-        return string(_data.toRlpItem().toBytes());
+    function toBytes(RLPReader.RLPItem memory item) internal pure returns (bytes memory) {
```
boolafish: I am reading this as moving codes we don't need yet to mock contract. Is this correct? If that is the case, can you add comments on why we are having such complex functions in our mock.

plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -1,7 +1,11 @@
...
+ * @author Hamdi Allam hamdi.allam97@gmail.com
+ * @notice RLP decoding library forked from https://github.com/hamdiallam/Solidity-RLP
+ * @dev Added more test cases from https://github.com/ethereum/tests/tree/master/RLPTests
+ *      Created more custom invalid test cases
+ *      Added more checks to ensure the decoder reads within bounds of the input length
 */
+
 pragma solidity ^0.5.0;
 
 library RLPReader {
```
boolafish: Since we are enhancing this for security purpose. Shall we be strict on our `uint`? which is to use `uint256` instead.

thec00n: Which `uint` are u referring to @boolafish ?

plasma_framework/contracts/mocks/utils/RLPMock.sol
thec00n: I wanted to make it clear on what parts of the code are used in production and which ones are not. Therefore I moved everything that is not strictly necessary to the mock contract. 

We have them so after the RLP has been decoded we can compare them back to strings and bytes. Where should I add a comment in the mock class?

plasma_framework/contracts/src/utils/RLPReader.sol
boolafish: More like all `uint`, aside from the one in for loop `(uint i=0)` ones that probably don't lose any clarity on the size.
I know that current solidity version uint == uint256. But when moving versions this from experience breaks sometimes. Also, just for readability a bit.

eg. when I was trying to use the code from the stack overflow answer, which is of version 0.4: https://ethereum.stackexchange.com/questions/13483/how-to-pass-arbitrary-bytes-to-a-function-in-remix

I have to specify `uint8` in casting instead of original `uint`: https://gist.github.com/kasima/e5fd6f05bb2d6adf49167b64bb0ad2d1#file-exitgame-sol-L80

https://ethereum.stackexchange.com/questions/43241/why-write-uint256-instead-of-uint-if-theyre-the-same-thing

boolafish: anyway, this is a minor one

plasma_framework/contracts/mocks/utils/RLPMock.sol
boolafish: Probably just on top of the contract and explain a bit. I am asking this as most our mock are used for:
1. Purely proxy the call to implementation
1. Add some simple helper function to inject states

So this is a bit out. I feel we are putting here to conform the solidity spec tests (?) and just for conveniency if we ever need reference code, right?

thec00n: Yes solidity spec tests is one reason, the other is that the code is unsafe in some of this functions. This can be distracting for the audit and in general for running automated security analysis tools that would report issues.

Issue #389, opened 2019-10-25
Current test coverage is not 100%


thec00n:
## Issue Type 

```
[ x ] bug report
[ ] feature request
```

## Current Behavior
The following contracts do not have yet a 100% test coverage:

- [x] RLPReader: currently tracked in #376: 
- [x] PriorityQueue: https://coveralls.io/builds/26522392/source?filename=contracts/src/framework/utils/PriorityQueue.sol
- [x] ExitGameController: I think we should replace the require with asserts as in both cases this should never occur. Security tools can then try to trigger the assert. 
(https://coveralls.io/builds/26522392/source?filename=contracts/src/framework/ExitGameController.sol#L54 and https://coveralls.io/builds/26522392/source?filename=contracts/src/framework/ExitGameController.sol#L73)
- [x] PaymentChallengeIFENotCanonical.sol: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
- [x] PaymentChallengeIFEInputSpent: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/controllers/PaymentChallengeIFEInputSpent.sol
- [x] PaymentChallengeStandardExit: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/controllers/PaymentChallengeStandardExit.sol
- [x] PaymentStartInFlightExit: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
- [x] PaymentExitGame: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/PaymentExitGame.sol
- [x] PaymentInFlightExitModelUtils: https://coveralls.io/builds/26522392/source?filename=contracts/src/exits/payment/PaymentInFlightExitModelUtils.sol

## Suggested Fix

- In a lot of cases the gaps in coverage are due to not triggering the negative branch of a require statement. Review the test suite and the existing code base and check if the require conditions are extensive enough and that the tests trigger the all code branches as expected. 
- Consider adding more negative test cases

pgebal:
I think we can split that into a couple of PRs, so I put a checklist into issue description. I'm on it, currently working on ExitGameController and PaymentChallengeIFENotCanonical. The rest is up for grabs.

boolafish:
I will take the Priority Queue test

pgebal:
[...]

PR #401, opened 2019-10-28
389 improve test coverage in exit game controller

pgebal:
- Full test coverage for `PaymentChallengeIFENotCanonical`.
- Use `assert` instead of `require` in `ExitGameController` when verifying conditions that are expected to never fail.
 
References #389

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -51,7 +51,7 @@ contract ExitGameController is ExitGameRegistry {
         require(!mutex, "Reentrant call");
         mutex = true;
         _;
-        require(mutex, "Not locked");
+        assert(mutex);
         mutex = false;
     }
 
@@ -70,7 +70,7 @@ contract ExitGameController is ExitGameRegistry {
      * @dev Accessible only from non quarantined exit games, uses a mutex
      */
     function deactivateNonReentrant() external onlyFromNonQuarantinedExitGame() {
-        require(mutex, "Not locked");
+        assert(mutex);
         mutex = false;
     }
 
--- a/plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
+++ b/plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
@@ -29,7 +29,9 @@ const {
 } = require('../../../helpers/constants.js');
 const { buildOutputGuard } = require('../../../helpers/utils.js');
 const { buildUtxoPos, UtxoPos } = require('../../../helpers/positions.js');
-const { PaymentTransactionOutput, PaymentTransaction } = require('../../../helpers/transaction.js');
+const {
+    PaymentTransactionOutput, PaymentTransaction, PlasmaDepositTransaction,
+} = require('../../../helpers/transaction.js');
 const { createInclusionProof } = require('../../../helpers/ife.js');
 
 contract('PaymentInFlightExitRouter', ([_, ifeOwner, inputOwner, outputOwner, competitorOwner, challenger]) => {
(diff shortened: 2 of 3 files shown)


plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
```diff
@@ -80,10 +89,26 @@ contract('PaymentInFlightExitRouter', ([_, ifeOwner, inputOwner, outputOwner, co
...
     };
 
-    const buildValidNoncanonicalChallengeArgs = (decodedIfeTx, outputType) => {
-        const { inputTx } = createInputTransaction(outputType);
-
-        const { competingTx, decodedCompetingTx, competingTxPos } = createCompetitorTransaction(outputType);
+    const buildValidNoncanonicalChallengeArgs = (
+        decodedIfeTx,
+        outputType,
+        nonDepositInputTx = true,
```
boolafish: Minor: I would use `isDepositInputTx = false` seems a bit easier to read as a var.

plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
```diff
@@ -60,17 +63,23 @@ contract('PaymentInFlightExitRouter', ([_, ifeOwner, inputOwner, outputOwner, co
...
-        return {
-            inputTx: web3.utils.bytesToHex(inputTx.rlpEncoded()),
-            decodedInputTx: inputTx,
-        };
+        return web3.utils.bytesToHex(inputTx.rlpEncoded());
+    };
+
+    const createDepositInputTransaction = (outputType) => {
+        const output = new PaymentTransactionOutput(
+            outputType, TEST_COMPETING_TX_OUTPUT_AMOUNT, buildOutputGuard(competitorOwner), ETH,
```
boolafish: probably should be `TEST_IFE_INPUT_AMOUNT` instead of `TEST_COMPETING_TX_OUTPUT_AMOUNT`.
also should be `buildOutputGuard(inputOwner)`

plasma_framework/test/src/exits/payment/PaymentChallengeIFENotCanonical.test.js
```diff
@@ -115,7 +140,7 @@ contract('PaymentInFlightExitRouter', ([_, ifeOwner, inputOwner, outputOwner, co
         };
     };
 
-    const buildInFlightExitData = async (exitIdHelper, outputIdHelper, outputType) => {
+    const buildInFlightExitData = async (exitIdHelper, outputIdHelper, outputType, nonDepositInputTx = true) => {
```
boolafish: same minor comment on `isDepositInputTx` instead

boolafish:
btw, @thec00n how does the tool that detects `assert` should never be triggered works? Does it work on cross contract calls as well?

thec00n:
`btw, @thec00n how does the tool that detects assert should never be triggered works?`
Find a tx such that an assert violation can be reached. In MythX there are currently two types of tools that can find such an issue. 

Generally analyzing our deployed contracts might not be optimal because they are huge and as you mentioned are dynamically linked which is not easy for tools to work with. I want to spend more time on isolating functionality and using the Mock contracts to analyze interesting properties.

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -70,7 +70,7 @@ contract ExitGameController is ExitGameRegistry {
      * @dev Accessible only from non quarantined exit games, uses a mutex
      */
     function deactivateNonReentrant() external onlyFromNonQuarantinedExitGame() {
-        require(mutex, "Not locked");
+        assert(mutex);
```
boolafish: @thec00n I put up the discussion of `assert` because it is quite easy to come up with a test that can trigger this assert. We just need an invalid ExitGame that violates our protocol, our only protection against it is the quarantined period.

thec00n: `I put up the discussion of assert because it is quite easy to come up with a test that can trigger this assert. We just need an invalid ExitGame that violates our protocol, our only protection against it is the quarantined period.`

If that is the only case where a tool can trigger an assert violation then that's fine. I suspect that they will not find this case because it's too complex create such a state for an automated tool. We are interested in assert violations apart from this scenario.

PR #402, opened 2019-10-28
IFE with deposit tx has wrong priority

pik694:
The PR provides required fixes to original `RootChain` contract related to wrong priority of IFE with a deposit.

Detailed info can be found in the following issue: #216.


pik694:
Merged even though the tests are not passing, since in this old branch we have no circle ci config (we used to use Travis). The tests are passing locally.

PR #407, opened 2019-10-29
fix: wrong comment on time unit

boolafish:


kevsul:
Ha ha, I just saw this too!

PR #413, opened 2019-10-30
Update RLPReader.sol

cduangprom:
Formatting edits in file


boolafish:
Hi @cduangprom thanks for the wording improvement!
Just FYI that we usually have tests on the error message of whatever behinds `require(..., "error message")`. So the change there would break our test and the CI would not pass.

If you are okay with this, do you mind to search the file to find the test with old error message and update it to the new one? Or it might be preferable to focus on the code document/comment wordings for now.

PR #415, opened 2019-10-30
refactor: move some ExitId tests to PaymentExitGame test suite to expose tested functions through public API

pgebal:
- Bumps test coverage of `PaymentExitGame` to 100% by moving `ExitId` tests to `PaymentExitGame` test suite (`PaymentExitGame` exposes tested function through public API).

References #389


boolafish:
well...that is the dilemma when things are coupled. I personally don't even think chasing this coverage here has strong benefit. But I would suggest to just replace the e2e test code using the `exitId` function from exit game to implicitly cover the code as there is no logic there anyway....

It is currently using the `ExitIdWrapper`. We can use the function form exit game directly instead.

eg. https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/src/exits/payment/PaymentExitGame.test.js#L244

pgebal:
> It is currently using the `ExitIdWrapper`. We can use the function from exit game directly instead.
> 
> eg. https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/test/src/exits/payment/PaymentExitGame.test.js#L244

Good idea! I'll do it that way.

PR #421, opened 2019-10-31
feat: convert truffle config from infura to remote

unnawut:
Currently the migration code on elixir-omg uses infura flag as a workaround to deploy contracts to local geth with specific private keys:

```yaml
  plasma-deployer:
    command:
      - /bin/sh
      - -c
      - |
          apk add --update curl
          cd /home/node/plasma-contracts/plasma_framework
          npx truffle migrate --network infura
          # ...
    environment:
      - INFURA_URL=http://geth:8545
      - INFURA_API_KEY=
      # ...
```

This PR generalizes the `infura` config into a `remote` config where the private keys are expected as inputs for contract deployment.

## Usage

```shell
$ REMOTE_URL=http://remote_provider_url \
  DEPLOYER_PRIVATEKEY=0xaaaa... \
  MAINTAINER_PRIVATEKEY=0xbbb... \
  AUTHORITY_PRIVATEKEY=0xccc... \
  npx truffle migrate --network remote
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,15 +63,14 @@ Deploying the contracts requires three accounts:
 2. `AUTHORITY` The account that used by the Child chain to submit blocks. It must not have made any transaction prior to calling the scripts i.e. its nonce must be 0.
 3. `MAINTAINER` The account that can register new vaults, exit games, etc.
 
-
-Normally you will deploy the contracts using an Ethereum client that you run yourself, such as Geth or Parity. However, you can also use a provider such as Infura. In this case you'll need to know the private keys for the `DEPLOYER`, `AUTHORITY` and `MAINTAINER` accounts. See `truffle-config.js` for an example.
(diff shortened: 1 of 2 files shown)


unnawut:
I believe it's because `local` is expecting the client to be unlocked and hence no private keys need to be provided, and this unlocked setup is still being used? And so @kendricktan used `infura` as a workaround. 

So this PR simply turns `infura` into `remote` since the existing infura logic wasn't really specific to infura except for the url building (which can be done outside anyway), and setting the private keys was more of a remote client setup.

Happy to move it to local but I'm not sure we want to remove the unlocked account approach yet...

boolafish:
oh I see, I think I misunderstood that the docker compose is only used in local development.

boolafish:
Can we document how to use this `remote` network and what env var should be set in README too? https://github.com/omisego/plasma-contracts#deploying

kendricktan:
If you update the README and fix the linting issues then LGTM :)

unnawut:
@boolafish README updated

unnawut:
Also if you could press the merge button... I have no write access to this repo 😅

PR #424, opened 2019-11-04
test: improve test coverage of starting in-flight exit

pgebal:
Bumps test coverage for starting in-flight exit.
Removes repeated verification of output guard data for input transactions.
Repeated `require` check of output guard handler address for input transactions not being zero is changed into `assert`.

References #389


plasma_framework/test/src/exits/payment/PaymentStartInFlightExit.test.js
```diff
@@ -324,28 +338,212 @@ contract('PaymentInFlightExitRouter', ([_, alice, richFather, carol]) => {
...
-                    DEPOSIT_BLOCK_NUMBER,
-                );
-                await this.framework.setBlock(BLOCK_NUMBER, inputTxsBlockRoot1, 0);
-                await this.framework.setBlock(DEPOSIT_BLOCK_NUMBER, inputTxsBlockRoot2, 0);
-
-                await expectRevert(
-                    this.exitGame.startInFlightExit(args, { from: alice, value: this.startIFEBondSize.toString() }),
-                    'Spending condition contract not found',
-                );
+            describe('and output guard handlers for all output types are registered', () => {
```
boolafish: minor: I would start with `given` instead of `and` here but that might be just my preference

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -213,10 +213,6 @@ library PaymentStartInFlightExit {
             exitData.inputTxs.length == exitData.inFlightTx.inputs.length,
             "Number of input transactions does not match number of in-flight transaction inputs"
         );
-        require(
```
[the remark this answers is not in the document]
boolafish: ah...okay just realize the same check is done in `getOutputId` so it is hard to do it here.

Might consider to just break this `verifyNumberOfInputsMatchesNumberOfInFlightTransactionInputs` and embed it into `createStartExitData` so that each line immediately verifies the length as that function is sort of not really possible of doing its original purpose.

pgebal:
> Minor: Potentially we can have another array for `outputGuardHanlder` in `StartExitData` so that we would verify it once and only once. For all helper functions it just uses the one from the `StartExitData`.
> 
> But just an refactor idea. I am fine with current code.

I had the same idea, but decided not to do it. As there are two of us now, I'll code that :+)

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -352,8 +384,7 @@ library PaymentStartInFlightExit {
                 output.outputGuard,
                 exitData.outputGuardPreimagesForInputs[i]
             );
-            IOutputGuardHandler handler = exitData.controller.outputGuardHandlerRegistry.outputGuardHandlers(output.outputType);
-            assert(address(handler) != address(0));
+            IOutputGuardHandler handler = exitData.outputGuardHandlersForInputTxs[i];
```
pgebal: @kevsul  I would be very grateful if you could take a quick look on what's going on here.
I have a problem that happy path unit tests (contract as in this file) fail here. All checks in `verifyStart` are commented out, except `verifyOutputGuardHandlersForInputTxsRegistered`. 
I can't find a bug. Possibly it's the way array of output guard handlers it's being initialized or used.
I don't think it's worth spending much time on that, I can get rid off `IOutputGuardHandler` array.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -182,19 +185,40 @@ library PaymentStartInFlightExit {
         return outputIds;
     }
 
+    function getOutputGuardHandlersForInputTxs(
```
boolafish: you forgot the return the array here, that might be the issue

pgebal: You are right @boolafish. I spotted that, when I uploaded it here, hence I removed the comment.
I was surprised the compiler didn't raised an error.

PR #432, opened 2019-11-06
feat: verify output type for deposit transactions

### LINKED ISSUE #430: The DepositVerifiers should check that the outputType is correct
## Issue Type 

```
[ ] bug report
[x ] feature request
```

## Current Behavior

Currently the DepositVerifiers don't check the outputType in the deposit transaction. This could lead to an unspendable utxo

## Expected Behavior

The DepositVerifiers should check that the outputType is payment (1)

## Steps to Reproduce

1.
2. 
3. 

- Full output of error:
- Command that caused error:
- Code that caused error:

## System Specs

- python version:
- pyetherem version:
- environment (output of running `pip freeze`):
- operating system:

pgebal:
Closes #430

### DIFF
--- a/plasma_framework/contracts/src/vaults/verifiers/Erc20DepositVerifier.sol
+++ b/plasma_framework/contracts/src/vaults/verifiers/Erc20DepositVerifier.sol
@@ -12,8 +12,13 @@ import {PaymentOutputModel as DepositOutputModel} from "../../transactions/outpu
 contract Erc20DepositVerifier is IErc20DepositVerifier {
     using DepositOutputModel for DepositOutputModel.Output;
 
+    uint256 public supportedOutputType;
     // Hardcoded transaction type for payment transaction
     uint8 constant private DEPOSIT_TX_TYPE = 1;
+
+    constructor(uint256 outputType) public {
+        supportedOutputType = outputType;
+    }
 
     /**
      * @notice Overrides the function of IErc20DepositVerifier and implements the verification logic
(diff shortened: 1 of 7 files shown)


plasma_framework/contracts/src/vaults/verifiers/Erc20DepositVerifier.sol
```diff
@@ -12,9 +12,14 @@ import {PaymentOutputModel as DepositOutputModel} from "../../transactions/outpu
 contract Erc20DepositVerifier is IErc20DepositVerifier {
     using DepositOutputModel for DepositOutputModel.Output;
 
+    uint256 public supportedOutputType;
     // Hardcoded transaction type for payment transaction
     uint8 constant private DEPOSIT_TX_TYPE = 1;
 
+    constructor(uint256 outputType) public {
```
boolafish: If we are doing this, might consider to just inject `txType` too.
I am fine with hardcode both or inject both, but better to be consistent

plasma_framework/contracts/src/vaults/verifiers/EthDepositVerifier.sol
```diff
@@ -10,9 +10,14 @@ import {PaymentOutputModel as DepositOutputModel} from "../../transactions/outpu
 contract EthDepositVerifier is IEthDepositVerifier {
     using DepositOutputModel for DepositOutputModel.Output;
 
+    uint256 public supportedOutputType;
     // Hardcoded transaction type for payment transaction
     uint8 constant internal DEPOSIT_TX_TYPE = 1;
 
+    constructor(uint256 outputType) public {
```
boolafish: Same as above, inject or hardcode both txType and outputType

plasma_framework/contracts/src/vaults/verifiers/Erc20DepositVerifier.sol
pgebal: done

plasma_framework/contracts/src/vaults/verifiers/EthDepositVerifier.sol
pgebal: done

PR #433, opened 2019-11-06
150 python tests se challenges

pik694:
On step further in direction of retrieving python tests - this time we reclaimed tests of SE challenges.


plasma_framework/python_tests/plasma_core/transaction.py
```diff
@@ -123,8 +123,19 @@ def serialize(cls, obj):
         return tx_sedes.serialize(tx_elems)
 
     def sign(self, index, account, verifyingContract=None):
```
boolafish: Do you want to also rename `verifyingContract` here?

plasma_framework/python_tests/plasma_core/transaction.py
```diff
@@ -123,8 +123,19 @@ def serialize(cls, obj):
...
             msg_hash).to_canonical_address() if sig != NULL_SIGNATURE else NULL_ADDRESS
+
+
+def _amend_signature(sig):
+    """ We are making it in order to make signatures produced by eth_keys library
+        compatible with openzellelin ECDSA public key recovery.
+
+        Please note:
+        https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c3f2ed81683bd0095673f525f7ee9639370a2432/contracts/cryptography/ECDSA.sol#L53
+    """
```
boolafish: oh....ouch :( Looks super painful to realize this.....

plasma_framework/python_tests/plasma_core/utils/eip712_struct_hash.py
```diff
@@ -1,63 +1,81 @@
...
+    empty_input = Input()
+
+    eip712_inputs = [Input(blknum=i.blknum, txindex=i.txindex, oindex=i.oindex) for i in inputs]
+    return eip712_inputs + [empty_input] * (4 - len(inputs))  # pad with empty inputs
+
+
+def _map_outputs(outputs):
+    empty_output = Output()
+
+    eip712_outputs = [
```
boolafish: minor: might consider to just write in `for` style as it is using multiple lines anyway. We can even just pad the empty output within the loop too.

plasma_framework/python_tests/tests/tests_utils/plasma_framework.py
```diff
@@ -149,8 +169,19 @@ def startStandardExit(self, utxo_pos, output_tx, output_tx_inclusion_proof, **kw
         args = (utxo_pos, output_tx, b'', output_tx_inclusion_proof)
         return self.payment_exit_game.startStandardExit(args, **kwargs)
 
-    def challengeStandardExit(self, standard_exit_id, challenge_tx, input_index, challenge_tx_sig):
-        raise NotImplementedError
+    def challengeStandardExit(self, standard_exit_id, challenge_tx, input_index, challenge_tx_sig, exiting_tx,
+                              **kwargs):
+        """NOTICE: takes more arguments than the original one"""
```
boolafish: It took me a while to understand this `NOTICE`. Might be better to be clear it's ALD vs Rootchain.sol or probably comment that extra args for potential future feature which is non-used now.

plasma_framework/python_tests/plasma_core/transaction.py
pik694: Actually it wasn't as painful as one might expect ;)

PR #442, opened 2019-11-11
chore: run gas report on e2e test only

boolafish:
### What
Unit tests relies on mock heavily, which might end up giving noice when reviewing gas usage.

### Some notes
After looking at the gas report, finally realize sth that had confused me for a while. Our report is always missing some functions. Seems like to be of this issue: https://github.com/cgewecke/eth-gas-reporter/issues/186

So basically all method using struct as input (almost all PaymentExitGame functions) would not have gas report due to that.


boolafish:
probably not that important to separate that clean now, closing this first

PR #444, opened 2019-11-12
fix: rlp catch invalid leading zeros

kevsul:


kevsul:
Closed in favour of #476

Issue #425, opened 2019-11-04
Merkle - The implementation does not enforce inclusion of leaf nodes.


cleanunicorn:
## Issue Type 

- [x] bug report

## Current Behavior
A observation with the current Merkle tree implementation is that not necessarily leaf nodes can be validated. One can call `checkMembership()` and instead of providing a leaf, it can provide a hash within the Merkle tree.

https://github.com/omisego/plasma-contracts/blob/ca378932458e2e9156aa00f76298088a12b16fe9/plasma_framework/contracts/src/utils/Merkle.sol#L9-L42

The current implementation will validate the provided "leaf" and return `true`. This is a known problem of Merkle trees https://en.wikipedia.org/wiki/Merkle_tree#Second_preimage_attack. 

## Expected Behavior
It should only validate leaf nodes as being included in the Merkle tree.

## Steps to Reproduce
Provide a hash from within the Merkle tree as the `leaf` argument, the index has to match the index of that node in regards to its current level in the tree. 
The `rootHash` has to be the correct Merkle tree `rootHash`.
The proof has to skip the necessary number of levels because the nodes "underneath" the provided "leaf" will not be processed.

## Suggested Fix
A remediation needs a fixed Merkle tree size which you have, but it also needs a prepended byte to each node in the tree. Another way would be to create a structure for the Merkle node and mark it as `leaf` or `no leaf`.

## Motivation for Change
If this change is made, the method will only validate provided leafs.

pgebal:
I think that's not an issue in our use case. We use MT lib to prove that transaction is included in a block stored in plasma.`leaf` argument is always calculated in our plasma contracts as a hash of a transaction bytes provided by a user. To make false claims about transaction being included in a block, user would need to prepare a transaction that is a preimage of a node. That looks to me as breaking the hash function. @boolafish @cleanunicorn

paulperegud:
There might be some vectors that would allow attacked to exploit the library. Attacker would need to provide a 32 bytes long value that is at the same time a hash of node in the tree AND a valid transaction (both valid RLP encoded tx and satisfies additional checks). Alternatively attacker would need to find an execution path where we don't unpack the transaction. Are we sure that there are no such paths?

We can patch the contract and fix tree depth. Alternatively, to still have the ability to have a tree of variable depth, we can do:
1) hash tx body with salt (any value) 
2) pass the Tx body into the library itself, instead of doing the hashing/salting outside

Appropriate changes in client software are also needed.

[...]

PR #446, opened 2019-11-13
Prevent second preimage attack on Merkle tree

pgebal:
Prevents second preimage attack on Merkle by prepending `0` byte to each leaf hash data, and `1` when computing internal node hashes. Solution was proposed in issue description.

That's the first step of solving the issue. The second is updating elixir libs for Merkle.

References #425


plasma_framework/python_tests/plasma_core/utils/merkle/fixed_merkle.py
```diff
@@ -41,31 +44,31 @@ def __create_tree(self, leaves):
...
             else:
-                computed_hash = sha3(proof_segment + computed_hash)
+                computed_hash = sha3(NODE_SALT + proof_segment + computed_hash)
             computed_index = computed_index // 2
 
         return computed_hash == self.root
 
     def create_membership_proof(self, leaf):
-        hashed_leaf = sha3(leaf)
+        hashed_leaf = sha3(LEAF_NODE + leaf)
```
pdobacz: `LEAF_NODE`  leftover? seems undefined

PR #452, opened 2019-11-15
Prevent second preimage attack on merkle tree

pgebal:
References #425
That's part one of solving this issue. Second part are upcoming changes to elixir-omg .


plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -5,38 +5,75 @@ const { expect } = require('chai');
...
+    before('setup merkle contract', async () => {
         this.merkleContract = await Merkle.new();
-        this.merkleTree = new MerkleTree(leaves);
     });
 
     describe('checkMembership', () => {
+        const leaves = [...Array(maxSize).keys()].map(index => web3.utils.bytesToHex(`leaf ${index}`));
+
+        before('setup merkle contract and tree value', async () => {
+            this.merkleContract = await Merkle.new();
```
boolafish: why a new Merkle contract?

plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -5,38 +5,75 @@ const { expect } = require('chai');
...
+            const result = await this.merkleContract.checkMembership(
+                web3.utils.sha3(leaves[leafIndex]),
+                leafIndex,
+                fakeRootHash,
+                proof,
+            );
+            expect(result).to.be.false;
+        });
+
+        it('should return false node preimage is provided as a leaf', async () => {
```
boolafish:
```suggestion
        it('should return false when a node preimage is provided as a leaf', async () => {
```

plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -5,38 +5,75 @@ const { expect } = require('chai');
...
         this.merkleContract = await Merkle.new();
-        this.merkleTree = new MerkleTree(leaves);
     });
 
     describe('checkMembership', () => {
+        const leaves = [...Array(maxSize).keys()].map(index => web3.utils.bytesToHex(`leaf ${index}`));
+
+        before('setup merkle contract and tree value', async () => {
+            this.merkleContract = await Merkle.new();
+            this.merkleTree = new MerkleTree(leaves);
```
boolafish: probably passing in `MERKLE_TREE_DEPTH` in here as well

plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -48,9 +85,31 @@ contract('Merkle', () => {
...
+                    leafIndex,
+                    rootHash,
+                    wrongSizeProof,
+                ),
                 'Length of Merkle proof must be a multiple of 32',
             );
         });
     });
+
+    it('check membership for tree with empty leaves', async () => {
```
boolafish: this is a bit misleading though, it does have a leave of the deposit tx.

plasma_framework/test/src/utils/Merkle.test.js
```diff
@@ -5,38 +5,75 @@ const { expect } = require('chai');
...
+                        leafIndex,
+                        rootHash,
+                        proof,
+                    );
                     expect(result).to.be.true;
                 });
             });
         });
 
         it('should return false when not able to prove included', async () => {
```
boolafish: It was asked by @thec00n a while ago, but if don't mind to put a small effort of adding a test, can you as well add one more in this file?

Something like the following:
```
it('should return false when leaf index mismatches the proof')
```

Basically we are missing tests when the index passed is wrong.

PR #459, opened 2019-11-19
Some minor fixes

### LINKED ISSUE #453: No existence checks on framework block and timestamp reads
## Issue Type 

```
[X] bug report
[ ] feature request
```

## Current Behavior

The exit game libraries make several queries to the main `PlasmaFramework` contract where plasma block hashes and timestamps are stored. In multiple locations, the return values of these queries are not checked for existence.

1. `PaymentStartStandardExit.setupStartStandardExitData`:

https://github.com/omisego/plasma-contracts/blob/1c7cd867b1353ae72c825a6d15b2d382689f30ad/plasma_framework/contracts/src/exits/payment/controllers/PaymentStartStandardExit.sol#L124

2. `PaymentChallengeIFENotCanonical.respond`:

https://github.com/omisego/plasma-contracts/blob/1c7cd867b1353ae72c825a6d15b2d382689f30ad/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol#L176

3. `PaymentPiggybackInFlightExit.enqueue`:

https://github.com/omisego/plasma-contracts/blob/1c7cd867b1353ae72c825a6d15b2d382689f30ad/plasma_framework/contracts/src/exits/payment/controllers/PaymentPiggybackInFlightExit.sol#L167

4. `TxFinalizationVerifier.checkInclusionProof`:

https://github.com/omisego/plasma-contracts/blob/1c7cd867b1353ae72c825a6d15b2d382689f30ad/plasma_framework/contracts/src/exits/utils/TxFinalizationVerifier.sol#L54

## Suggested Fix

Each query to `PlasmaFramework.blocks` should be followed with a check that the returned value is nonzero.

## Motivation for Change

Although none of these examples seem exploitable, adding existence checks makes it easier to reason about the code.

### LINKED ISSUE #454: ECDSA error value is not handled
## Issue Type 

```
[X] bug report
[ ] feature request
```

## Current Behavior

The OpenZeppelin `ECDSA` library returns `address(0x00)` for many cases with malformed signatures:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/49042f2b1ae76eb9befa12000b98211981a139ec/contracts/cryptography/ECDSA.sol#L57-L63

The `PaymentOutputToPaymentTxCondition` contract does not explicitly handle this case:

https://github.com/omisego/plasma-contracts/blob/1c7cd867b1353ae72c825a6d15b2d382689f30ad/plasma_framework/contracts/src/exits/payment/spendingConditions/PaymentOutputToPaymentTxCondition.sol#L65-L68

## Suggested Fix

Adding a check to handle this case will make it easier to reason about the code.

boolafish:
### Note
1. fix: use exitId instead of top priority in processExits tests 
1. fix: checks block data is empty or not when getting from framework. closes #453 
1. fix: revert when not able to recover the signer from signature. closes #454

### DIFF
--- a/plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
+++ b/plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
@@ -240,14 +240,14 @@ contract('PaymentInFlightExitRouter', ([_, alice, inputOwner, nonInputOwner, out
             );
         });
 
-        it('should fail when failed to get the block timestamp info of the exit position', async () => {
+        it('should fail when there is no block for the exit position to enqueue', async () => {
             const data = await buildPiggybackInputData();
             await this.exitGame.setInFlightExit(data.exitId, data.inFlightExitData);
             await expectRevert(
                 this.exitGame.piggybackInFlightExitOnInput(
                     data.argsInputOne, { from: inputOwner, value: this.piggybackBondSize.toString() },
                 ),
-                'Failed to get the block timestamp information of the exit position to enqueue',
+                'There is no block for the exit position to enqueue',
             );
         });
 
--- a/plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
+++ b/plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
@@ -218,11 +218,9 @@ contract('PaymentInFlightExitRouter', ([_, alice, inputOwner, nonInputOwner, out
         it('should fail when the same input has been piggybacked', async () => {
             const data = await buildPiggybackInputData();
             await this.exitGame.setInFlightExit(data.exitId, data.inFlightExitData);
-            await this.exitGame.piggybackInFlightExitOnInput(
-                data.argsInputOne, { from: inputOwner, value: this.piggybackBondSize.toString() },
-            );
 
-            // second attmept should fail
+            await this.exitGame.setInFlightExitInputPiggybacked(data.exitId, data.argsInputOne.inputIndex);
+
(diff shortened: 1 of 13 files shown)


plasma_framework/test/src/exits/payment/PaymentPiggybackInFlightExitOnInput.test.js
```diff
@@ -242,6 +240,17 @@ contract('PaymentInFlightExitRouter', ([_, alice, inputOwner, nonInputOwner, out
             );
         });
 
+        it('should fail when failed to get the block timestamp info of the exit position', async () => {
```
pgebal: Let's name this test and the one in `PaymentPiggybackInFlightExitOnOutput.test.js` the same.
I think we should use different error message, how about:
There is not block for the utxo`? That's basically what is going on here.

plasma_framework/test/src/exits/payment/PaymentStartStandardExit.test.js
```diff
@@ -147,10 +148,21 @@ contract('PaymentStandardExitRouter', ([_, outputOwner, nonOutputOwner]) => {
...
         });
 
+        it('should fail when the block of the position does not exists in the Plasma Framework', async () => {
+            const { args } = buildTestData(this.dummyAmount, outputOwner, this.dummyBlockNum);
+            // test by not stubbing the block data accordingly
+            await expectRevert(
+                this.exitGame.startStandardExit(
+                    args, { from: outputOwner, value: this.startStandardExitBondSize },
+                ),
+                'Failed to get the block timestamp of the UTXO position',
```
pgebal: Again, how about `There is not block for the utxo`?

Issue #128, opened 2019-06-14
Expend some getters in contract to batch getter to test out parity high mem issue


boolafish:
See discussion here:
https://github.com/omisego/devops/issues/87#issuecomment-501452334

The followings are using default getter from public variable. (all using `mapping`)
```
- exits
- inflightExits
- blocks
```

We could try out some experiment by write some simple batch getter by taking an array of ids/kets and return the value.

boolafish:
We have two options to return list, return as different list per field of struct or to use `ABIEncoderV2` to return a list of struct directly.

Some really badly written POC code with ABIEncoderV2

```
contract A {
    struct B {
        uint256 num;
    }
    
    mapping(uint256 => B) bb;
    
    function getBList(uint256[] memory keys) public view returns (B[] memory) {
        B[] memory result = new B[](keys.length);
        for (uint i = 0; i < keys.length; i++) {
            uint256 k = keys[i];
            result[i] = bb[k];
        }
        return result;
    }
}
```

boolafish:
https://github.com/omisego/devops/issues/87#issuecomment-514561683
seems like the need of this has de-prioritized. Close this for now to not stack the open issues.
If we are to revisit this, let's see what action do we want to take again.

PR #468, opened 2019-11-20
feature: retrieve array of data in one go

InoMurko:


InoMurko:
Hi @kevsul @thec00n 
question... should I just remove `standardExit(uint160 exitId)` and we always use `standardExits(uint160[] memory exitIds)`? And change existing tests to pass in an array of length 1?

kevsul:
In general I think it's better to keep the contracts as simple as possible and let clients do more complex stuff. I realise that this loop isn't especially complex, but it's still more complex than a one liner.

So yeah, this way is slightly more convenient for the client at the cost of slightly more complex contracts, but what's the use case? How often you need to get multiple standard exits at once?

InoMurko:
It completely depends on the amount of standard exits. The problem we're having is that if you have a large amount of SEs they result in _an RPC call per SE_. Which is
1. it's hard to predict the response times because it goes over the wire
2. related costs (Infura)

And having a large amount, say 10, 20, 50,... (let alone couple of thousand) if one of the calls fails, it makes processing SEs fail completely and needs to be restarted. This largely comes from the initial elixir-omg design...

kevsul:
Ok, got it, I can see how it's useful here.

> should I just remove standardExit(uint160 exitId) and we always use standardExits(uint160[] memory exitIds)? And change existing tests to pass in an array of length 1?

Yep, I'm in favour of this.

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
```diff
@@ -105,10 +105,22 @@ contract PaymentStandardExitRouter is
...
+    function standardExit(uint160 exitId) public view returns (PaymentExitDataModel.StandardExit memory) {
         return standardExitMap.exits[exitId];
     }
 
+    /**
+     * @notice Getter retrieves standard exit data of the PaymentExitGame
+     * @param exitIds Exit IDs of the standard exits
+     */
+    function standardExits(uint160[] memory exitIds) public view returns (PaymentExitDataModel.StandardExit[] memory) {
+        PaymentExitDataModel.StandardExit[] memory ids;
```
pgebal: `exits` would be a better name than `ids`

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
```diff
@@ -105,10 +105,22 @@ contract PaymentStandardExitRouter is
...
-    function standardExits(uint160 exitId) public view returns (PaymentExitDataModel.StandardExit memory) {
+    function standardExit(uint160 exitId) public view returns (PaymentExitDataModel.StandardExit memory) {
         return standardExitMap.exits[exitId];
     }
 
+    /**
+     * @notice Getter retrieves standard exit data of the PaymentExitGame
+     * @param exitIds Exit IDs of the standard exits
+     */
+    function standardExits(uint160[] memory exitIds) public view returns (PaymentExitDataModel.StandardExit[] memory) {
```
pgebal: I think we can make it `external`.

InoMurko: done!

boolafish:
> It completely depends on the amount of standard exits. The problem we're having is that if you have a large amount of SEs they result in an RPC call per ...

Just want to follow the reasoning a bit more, is the reason of this still solving this: https://github.com/omisego/plasma-contracts/issues/128
Or it is a different issue?

InoMurko:
@boolafish different but related!

plasma_framework/docs/contracts/PaymentStandardExitRouter.md
```diff
@@ -66,20 +66,21 @@ function (PlasmaFramework plasmaFramework, uint256 ethVaultId, uint256 erc20Vaul
 | txFinalizationVerifier | ITxFinalizationVerifier |  | 
 | safeGasStipend | uint256 |  | 
 
-### standardExits
+### standardexits
```
boolafish: Ah...I guess you did this manually?
Can you run `npn run docgen` instead (under `plasma_framework/` folder)?

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
```diff
@@ -103,10 +103,15 @@ contract PaymentStandardExitRouter is
 
     /**
      * @notice Getter retrieves standard exit data of the PaymentExitGame
-     * @param exitId Exit ID of the standard exit
+     * @param exitIds Exit IDs of the standard exits
      */
-    function standardExits(uint160 exitId) public view returns (PaymentExitDataModel.StandardExit memory) {
-        return standardExitMap.exits[exitId];
+    function standardExits(uint160[] calldata exitIds) external view returns (PaymentExitDataModel.StandardExit[] memory) {
```
thec00n: Unbound loops are not recommended in smart contracts. Let's add another parameter and require the caller to provide how many items he wants to have returned.

plasma_framework/test/endToEndTests/PaymentExitGame.e2e.test.js
```diff
@@ -283,7 +284,7 @@ contract('PaymentExitGame - End to End Tests', ([_deployer, _maintainer, authori
                 });
```
thec00n: Can we add a test cases that retrieves multiple `PaymentExitDataModel.StandardExit`s. Wonder if there is any upper bound to return arrays of structs. Can we test with 10,100,1000?

plasma_framework/contracts/src/exits/payment/routers/PaymentStandardExitRouter.sol
boolafish: This function is passing in the exact keys to retrieve data, the size is exactly the size of the array. I don't really see more benefit by passing in the size of array as another parameter here though.

thec00n: If this is just a getter that a clients uses it's fine. If at some point we decided to use this from another contract then the call could run out of gas.

boolafish: > If at some point we decided to use this from another contract then the call could run out of gas.

ehh yes. But my point is that adding another var does not add benefit. The client should just pass in less exitIds in that case.

InoMurko: also I presume that because we tagged the function as `external` you can't really use it from another contract, right?

boolafish: oh no lol, tag it `external` make it accessible from outside of that contract. (So not accessible within the contract, weird I know lol)

`public` make it accessible too everything. Just `external` would be slightly gas efficient compare to public

thec00n: It's fine for client, the problem would arise if the call is from another contract. I am ok with @dev comment saying that you need to to be careful if you call this function from another contract,

InoMurko:
tnx @boolafish @kevsul. I'll be adding @thec00n request for test!

boolafish:
I think this is ready to merge?

PR #476, opened 2019-11-22
Rlp meltdown

kevsul:
Extra restrictions to RLP and tx decoding:

- refactored RLPReader.sol yo remove duplicate code
- RLPReader.toUint() treats leading zeros as invalid
- RLPReader.toAddress() requires the item data to be exactly 20 bytes.
- RLPReader.toBytes32() requires the item data to be exactly 32 bytes.
- Transaction inputs must be bytes32
- Transaction inputs cannot be empty (all zeros)
- Transaction outputs cannot be empty (outputGuard == 0 AND amount == 0)
- Transaction metadata must be bytes32


plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -135,102 +178,85 @@ library RLPReader {
...
+                dataLen := div(mload(memPtr), exp(256, sub(WORD_SIZE, lengthLen)))
+                decodedLength := add(dataLen, add(lengthLen, 1))
             }
-            // Check valid long string i.e. value of length > MAX_SHORT_LEN with no leading zeros
+            // Check that the length has no leading zeros
             require(byte1 != 0, "Invalid RLP encoding");
+            // Check that the value of length > MAX_SHORT_LEN
             require(dataLen > MAX_SHORT_LEN, "Invalid RLP encoding");
+            // Calculate the offset
+            offset = (byte0 - (STRING_LONG_START - 1)) + 1;
```
boolafish: minor: it would easier to read if this is sth like how it is calculated previously in `add(legthLen, 1)`.

but probably just me as I am checking this function agains the RLP spec python example: https://github.com/ethereum/wiki/wiki/RLP

plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -135,102 +178,85 @@ library RLPReader {
...
+                dataLen := div(mload(memPtr), exp(256, sub(WORD_SIZE, lengthLen)))
+                decodedLength := add(dataLen, add(lengthLen, 1))
             }
-            // Check valid long list i.e. value of length > MAX_SHORT_LEN with no leading zeros
+            // Check that the length has no leading zeros
             require(byte1 != 0, "Invalid RLP encoding");
+            // Check that the value of length > MAX_SHORT_LEN
             require(dataLen > MAX_SHORT_LEN, "Invalid RLP encoding");
+            // Calculate the offset
+            offset = (byte0 - (LIST_LONG_START - 1)) + 1;
```
boolafish: minor: same as previous comment it would easier to read if this is sth like how it is calculated previously in `add(lengthLen, 1)`.

plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -135,102 +178,85 @@ library RLPReader {
...
-     * @notice Decodes the RLPItems length from a bytes array.
+     * @notice Decodes the RLPItem's length and offset.
      * @dev This function is dangerous !!! Ensure that the returned length is within the bounds of where memPtr points to.
      * @param memPtr Pointer to the dynamic bytes array in memory
-     * @return The encoded RLPItem length
+     * @return The length of the RLPItem (including the length field) and the offset of the payload
      */
-    function decodeItemLengthUnsafe(uint256 memPtr) internal pure returns (uint256) {
-        uint256 decodedItemLengthUnsafe;
+    function decodeLengthAndOffset(uint256 memPtr) internal pure returns (uint256, uint256) {
```
boolafish: Can we just pass in the `RLPItem` and also do the length check done in the RLP spec python example?

https://github.com/ethereum/wiki/wiki/RLP

boolafish:
lol, it is so much "fun" to read the assemblies, ewwww

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -38,18 +38,26 @@ library PaymentTransactionModel {
 
         bytes32[] memory inputs = new bytes32[](rlpInputs.length);
         for (uint i = 0; i < rlpInputs.length; i++) {
-            bytes32 input = bytes32(rlpInputs[i].toUint());
+            bytes32 input = rlpInputs[i].toBytes32();
+            // Disallow empty inputs
+            require(uint256(input) != 0, "Empty input not allowed");
```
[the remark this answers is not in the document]
kevsul: Agreed, 'empty' is confusing. How about "Null input not allowed"?

plasma_framework/contracts/src/utils/RLPReader.sol
kevsul: I don't think it easy to do it like that. In the python example they use string manipulation with substr and concatenation, but this isn't so easy in Solidity. Memory pointers seem to be a better fit.

boolafish: oh, I am suggesting about the memory pointer part. But if just compare the two functions, the python example also has conditions like `length > prefix - 0x80` in the branchings. If we pass in the RLPItem, I think we can use the `item.len` to match the logic in python.

plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -135,102 +178,85 @@ library RLPReader {
     }
 
     /**
-     * @notice Decodes the RLPItems length from a bytes array.
+     * @notice Decodes the RLPItem's length and offset.
      * @dev This function is dangerous !!! Ensure that the returned length is within the bounds of where memPtr points to.
      * @param memPtr Pointer to the dynamic bytes array in memory
-     * @return The encoded RLPItem length
+     * @return The length of the RLPItem (including the length field) and the offset of the payload
```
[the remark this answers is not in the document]
kevsul: I mean the total length  - length of the length field and the data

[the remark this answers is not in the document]
boolafish: ah...just see `offset` is in the later part of the sentence

plasma_framework/contracts/src/utils/RLPReader.sol
kevsul: They're different. In python `length = len(input)`, but in Solidity you only have the length of the overall buffer. You don't get the length of each item (until you decode it)

kevsul: I mean, in python `length` is the length of the remainder of the buffer, which is not the same as the `item.len`.

boolafish: okay, let me try to understand the assemblies again 😅

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -24,8 +24,24 @@ library PaymentTransactionModel {
...
+     *      - `outputs` is an list of 0 to 4 elements
+     *      - `output.outputType` must be an integer value with no leading zeros
+     *      - `output.outputGuard` is a 20 byte long array
+     *      - `output.token` is a 20 byte long array
+     *      - `output.amount` must be an integer value with no leading zeros
+     *      - An `output` may not be null; A null output is one whose amount is zero
+     * @param tx An RLP-encoded transaction
+     * @return A decoded PaymentTransaction struct
+     */
+    function decode(bytes memory tx) internal pure returns (PaymentTransactionModel.Transaction memory) {
```
boolafish: my IDE argues about the `tx` variable

> This declaration shadows a builtin symbol.

kevsul: Nice catch, I wonder why sol linter is picking these up?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -35,20 +51,24 @@ library PaymentTransactionModel {
...
+            bytes32 input = rlpInputs[i].toBytes32();
+            // Disallow null inputs
+            require(uint256(input) != 0, "Null input not allowed");
             inputs[i] = input;
         }
 
         PaymentOutputModel.Output[] memory outputs = new PaymentOutputModel.Output[](rlpOutputs.length);
         for (uint i = 0; i < rlpOutputs.length; i++) {
             PaymentOutputModel.Output memory output = PaymentOutputModel.decode(rlpOutputs[i]);
+            require(output.outputType > 0, "Output type must not be 0");
```
thec00n: 1.) Why are the checks related to `PaymentOutputModel` not in `decode`?
2.) I think we concluded that the amount has be greater than 0?

thec00n:
- Can we add tests that cover all loc/branches in https://coveralls.io/builds/27286020/source?filename=contracts/src/utils/RLPReader.sol
- Is the plan to remove https://github.com/omisego/plasma-contracts/blob/11a731267140cb8f54084f456c08f97ada8d0206/plasma_framework/contracts/src/transactions/WireTransaction.sol? If not than we need to add checks there as well.

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
kevsul: > 1.) Why are the checks related to PaymentOutputModel not in decode?

kevsul: > 2.) I think we concluded that the amount has be greater than 0?

kevsul:
> * Can we add tests that cover all loc/branches in https://coveralls.io/builds/27286020/source?filename=contracts/src/utils/RLPReader.sol

Yep

> * Is the plan to remove https://github.com/omisego/plasma-contracts/blob/11a731267140cb8f54084f456c08f97ada8d0206/plasma_framework/contracts/src/transactions/WireTransaction.sol? If not than we need to add checks there as well.

Will be refactored.

Issue #473, opened 2019-11-22
Start Exit should check that it is the tx type the exit game supporting


boolafish:
## Issue Type 

```
[x] bug report
[ ] feature request
```
Realized this after seeing the issue #472 
> Crucially, this association is never used. It is implied heavily that transactions with some txType must use a certain registered exit game contract. In fact, this is not true. When using PaymentExitGame, its routers, and their associated controllers, the txType is invariably inferred from the encoded transaction, not from the mappings in ExitGameRegistry

## Current Behavior
startSE and startIFE does not check the tx type is of what that exit game contract is in responsible or not.

## Expected Behavior
start SE and start IFE should verify that the exiting tx is really the tx type that the exit game contract is dealing with.

## Suggested Fix
startIFE, check with the `supportedTxType`.
startSE, pass the `supportedTxType` to controller constructor and check

PR #478, opened 2019-11-25
Check tx type of exit game

boolafish:
### Note
- Add tx type check on startSE and startIFE. This makes sure that exit game only start exit with the correct tx type. So we cannot start exit for Payment tx v2 on PaymentExitGame V1 contract. closes: #473 
- Use `Args` in Routers (SE, IFE). Stack too deep again when trying to pass the `supportTxType` to the SERouter. Thus passing in args struct instead.


plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -233,6 +234,10 @@ library PaymentStartInFlightExit {
         require(exit.exitStartTimestamp == 0, "There is an active in-flight exit from this transaction");
     }
 
+    function verifyInFlightTxType(StartExitData memory exitData) private pure {
+        require(exitData.inFlightTx.txType == exitData.controller.supportedTxType, "Not the supported tx type of the exit game");
```
pgebal: Why not just `Unsupported transaction type` or 'Exit game does not support this transaction type' ? Same for standard exit.

boolafish: done

PR #479, opened 2019-11-25
Rlp meltdown gerhard

kevsul:


plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -114,23 +114,23 @@ library RLPReader {
...
-        require(itemLen <= item.len, "Decoded length is greater than input data");
+        require(itemLen == item.len, "Decoded item length must be equal to the input data length");
 
-        uint256 len = itemLen - offset;
+        uint256 dataLen = itemLen - offset;
 
         uint result;
-        uint byte0;
-        uint memPtr = item.memPtr + offset;
+        uint byte0Data;
```
kevsul: If we change these variable names here, we should change them everywhere else as well for consistency

plasma_framework/contracts/src/utils/RLPReader.sol
```diff
@@ -197,8 +196,9 @@ library RLPReader {
             // Item is a single byte
             decodedLength = 1;
             offset = 0;
-        } else if (byte0 < STRING_LONG_START) {
-            // Item is a short string (length <= 55 bytes)
+        } else if (STRING_SHORT_START <= byte0 && byte0 < STRING_LONG_START) {
```
kevsul: Nitpick: the added `STRING_SHORT_START <= byte0` (and similar in the other if statements) is unnecessary and costs extra gas.
I guess you put it in to make the code cleaner, I'm in favour of that and am happy to leave it there. Just wanted to point out the extra (probably minimal) gas cost

plasma_framework/test/src/utils/RLP.test.js
```diff
@@ -36,8 +47,11 @@ contract('RLP', () => {
         await testNumberDecoded(this.test.decodeUint, 100);
     });
 
-    it('should decode int 0', async () => {
-        await testNumberDecoded(this.test.decodeInt, 0);
+    it('should decode 0x00', async () => {
```
kevsul: It turns out that toUint() should _not_ decode `0x00`.
I'll change this test when I fix that

plasma_framework/contracts/src/utils/RLPReader.sol
thec00n: Agree on the extra gas cost, it just drove me mad to jump back and forth 3 times until you are sure again what range byte0 has. I am ok if you want to remove them again

plasma_framework/contracts/src/utils/RLPReader.sol
thec00n: Agree. Should I look through the code and make the changes?

PR #480, opened 2019-11-26
Improve readibility of payment game impl doc and tx types doc

thec00n:
Fix grammar and typos for the most part while reading the docs and trying to understand output and tx type:

**A question**:

Paragraph here: https://github.com/omisego/plasma-contracts/blob/cbe34d117a74772f6bf1d8798d47515c105a3e47/plasma_framework/docs/design/payment-game-implementation-v1.md#L34

* Am not sure about this statement -> `Also, output type is bound to the input type`. Could we give more information here?
* Totally don't know what that mean? `Currently the only output type that requires the spending tx's input to be using `utxoPos`. Potentially there can be output type deciding to be pointed by `outputId` instead in the future.`

**Generally**
It's hard to grasp all the scenarios of how tx types out put types are allowed to be used within the system. It be helpful to give some concrete examples with the specific flows (deposit, transact, exit) on where output and tx types are used to execute custom logic.

Also it be good to describe on what properties an output and tx type define in the system.

 CC @boolafish

### DIFF
--- a/plasma_framework/tenderly.yaml
+++ b/plasma_framework/tenderly.yaml
@@ -1,2 +1,0 @@
-account_id: 1171f09b-c532-452c-b473-e58742e28ce6
-project_slug: audit-1c7cd86-rinkeby
--- a/plasma_framework/tenderly.yaml
+++ b/plasma_framework/tenderly.yaml
@@ -1,0 +1,2 @@
+account_id: 1171f09b-c532-452c-b473-e58742e28ce6
+project_slug: audit-1c7cd86-rinkeby
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ coverage.json
 
 # pytest-xprocess
 .xprocess/
+
+#tenderly config 
+tenderly.yaml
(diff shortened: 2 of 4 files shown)


plasma_framework/tenderly.yaml
```diff
@@ -0,0 +1,2 @@
+account_id: 1171f09b-c532-452c-b473-e58742e28ce6
```
boolafish: lol, let's add to gitignore

thec00n: lol again ... done

boolafish:
> Am not sure about this statement -> Also, output type is bound to the input type. Could we give more information here?

> Currently the only output type that requires the spending tx's input to be using utxoPos. Potentially there can be output type deciding to be pointed by outputId instead in the future.

It is about 1:1 relationship of output type and input pointer type ish stuff. hmm, so basically we are discovering how to do the change from using UtxoPos -> OutputId. And the conclusion was we have a 1:1 relationship of the input schema with the output type that it is pointing to. So for instance, all inputs pointing to output type A should be using UtxoPos while output type B might need the pointers to be using OutputId schema.

https://github.com/omisego/research/issues/93#issuecomment-517734509

boolafish:
Merging as I am adding the recent changes to the doc, (delete api and the new logic on processIFE). Would like to avoid conflicts.

Issue #470, opened 2019-11-21
Revisit the output flagging (on exit) protocol, implementation & docs


pdobacz:
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior

consider the following scenario:
1. Alice and Bob sign `tx1`, spending outputs `oA` and `oB` respectively, both "old" outputs
2. Bob signs `tx2`, spending `oB`
3. `tx1` is submitted and withheld
4. `tx2` is submitted, included and withheld
4. Bob IFEs with `tx2` (first!)
5. Alice IFEs with `tx1`
6. Both IFEs get challenged, but `tx2` IFE gets a response and is canonical, `tx1` is non canonical
7. Bob's IFE finalizes 1st and flags Bob's input, Bob gets oB out
8. Due to the flag rule, Alice's IFE will not finalize, because oB is flagged, leaving her without `oA`

https://github.com/omisego/plasma-contracts/blob/master/plasma_framework/docs/design/payment-game-implementation-v1.md was used as source, which is a summary of current impl.

## Expected Behavior

Alice should be able to exit `oA`

## Suggested Fix

Something was already proposed by @paulperegud, pls do the honors

As part of the effort I suggest, that precise specs of this part of the protocol is produced somewhere in a suitable spot. Then pls combine this with fixing https://github.com/omisego/elixir-omg/blob/master/docs/standard_vs_in_flight_exits_interaction.md, per this comment: https://github.com/omisego/elixir-omg/pull/1118#issuecomment-557132890

> docs/specs should hold the current design of the mechanics and (if necessary) explanation why it works, in its entirety. Just state what and when is flagged and when flags are checked and that's it.
>
> All the links to prior discussions/PRs should only be used to provide historical background. But one should not need to follow those links to understand how and why things work.

## Motivation for Change

ask Alice :)

boolafish:
Note from slack

by @paulperegud 
```
1) flag only when finalizing the exit
2) for canonical mark all inputs spend and exited outputs spend
3) for non-canonical mark only exiting inputs
```

However, above would not solve the exact scenario as stated in the issue. Thus second patch:
```
Looks like the problem is skipping IFE when input flagged? Probably should treat it as non canonical instead?
```

boolafish:
seems like the proposal of treat input flagged as non-canonical would also solve the original issue: https://github.com/omisego/plasma-contracts/issues/102#issuecomment-495809967

[...]

PR #483, opened 2019-11-27
fix: treat input spent as non-canonical

boolafish:
### Note
Changes the behavior when input is flagged as spent. Previously, it would be omitted while now it
would be consider non-canonical instead. Omitting would lead to edge case that some input would be locked to the system while it should be able to exit.

Also, this makes changes to how we flag inputs/outputs:
1) flag only when finalizing the exit
2) for canonical mark all inputs spend and exited outputs spend
3) for non-canonical mark only exiting inputs

ref: https://github.com/omisego/plasma-contracts/issues/470

### Side change and refactor
- Add `isInputEmpty`, `isOutputEmpty` functions to `PaymentInFlightExitModelUtils`
- Use `isXXXEmpty` check instead of `amount > 0`
- apply `isXXXEmpty` to piggyback functions to make them fail more explicitly
- fix `PaymentInFlightExitModelUtils` tests that have the for loop written wrong....
- move tests for controller into `controllers/` folder. Too many PaymentXXX tests now.

### TODO
- [x] run docgen

### Test
```
      When any input is spent, given the challenge game result is canonical
        ✓ should be treat as non canonical (140ms)
      When the exit is non canonical, and some inputs/outputs are piggybacked
        ✓ should flag and only flag piggybacked inputs with the same token as spent (136ms)
        ✓ should NOT flag output as spent (145ms)
      When the exit is canonical, and some inputs/outputs are piggybacked
        ✓ should flag ALL inputs with the same token as spent (166ms)
        ✓ should flag and only flag piggybacked output with the same token as spent (169ms)
```

### DIFF
--- a/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
+++ b/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
@@ -13,6 +13,43 @@ contract PaymentInFlightExitModelUtilsMock {
         ife.exitStartTimestamp = exitStartTimestamp;
     }
 
+    /** Helper functions */
+
+    function setWithdrawData(string memory target, uint16 index, ExitModel.WithdrawData memory data)
+        public
+    {
+        if (stringEquals(target, "inputs")) {
+            ife.inputs[index] = data;
+        } else if (stringEquals(target, "outputs")) {
+            ife.outputs[index] = data;
+        } else {
+            revert("target should be either inputs or outputs");
+        }
+    }
+
+    function stringEquals(string memory a, string memory b) private pure returns (bool) {
+        return keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b));
+    }
+
+
+    /** Wrapper functions */
+
+    function isInputEmpty(uint16 index)
+        external
+        view
+        returns (bool)
+    {
+        return PaymentInFlightExitModelUtils.isInputEmpty(ife, index);
+    }
+
+    function isOutputEmpty(uint16 index)
+        external
+        view
+        returns (bool)
+    {
+        return PaymentInFlightExitModelUtils.isOutputEmpty(ife, index);
+    }
+
     function isInputPiggybacked(uint16 index)
         external
         view
--- a/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
+++ b/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
@@ -14,8 +14,10 @@ contract PaymentInFlightExitModelUtilsMock {
     }
 
     /** Helper functions */
-
-    function setWithdrawData(string memory target, uint16 index, ExitModel.WithdrawData memory data)
+    function setWithdrawData(
+        string memory target,
+        uint16 index, ExitModel.WithdrawData memory data
+    )
         public
     {
         if (stringEquals(target, "inputs")) {
@@ -31,9 +33,7 @@ contract PaymentInFlightExitModelUtilsMock {
         return keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b));
     }
 
-
     /** Wrapper functions */
-
     function isInputEmpty(uint16 index)
         external
         view
--- a/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
+++ b/plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
@@ -16,7 +16,8 @@ contract PaymentInFlightExitModelUtilsMock {
     /** Helper functions */
     function setWithdrawData(
         string memory target,
-        uint16 index, ExitModel.WithdrawData memory data
+        uint16 index,
+        ExitModel.WithdrawData memory data
     )
         public
     {
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
@@ -137,15 +137,15 @@ library PaymentProcessInFlightExit {
         returns (bool)
     {
         uint256 inputNumOfTheToken;
-        for (uint i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
+        for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
+            if (exit.inputs[i].token == token && !exit.isInputEmpty(i)) {
                 inputNumOfTheToken++;
             }
         }
         bytes32[] memory outputIdsOfInputs = new bytes32[](inputNumOfTheToken);
         uint sameTokenIndex = 0;
-        for (uint i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
+        for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
+            if (exit.inputs[i].token == token && !exit.isInputEmpty(i)) {
                 outputIdsOfInputs[sameTokenIndex] = exit.inputs[i].outputId;
                 sameTokenIndex++;
             }
@@ -232,7 +232,7 @@ library PaymentProcessInFlightExit {
     {
         uint256 inputNumOfTheToken;
         for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
+            if (exit.inputs[i].token == token && !exit.isInputEmpty(i)) {
                 inputNumOfTheToken++;
             }
         }
@@ -247,7 +247,7 @@ library PaymentProcessInFlightExit {
         bytes32[] memory outputIdsToFlag = new bytes32[](inputNumOfTheToken + piggybackedOutputNumOfTheToken);
         uint indexForOutputIds = 0;
         for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
+            if (exit.inputs[i].token == token && !exit.isInputEmpty(i)) {
                 outputIdsToFlag[indexForOutputIds] = exit.inputs[i].outputId;
                 indexForOutputIds++;
             }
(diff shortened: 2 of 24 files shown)


plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -193,25 +197,45 @@ library PaymentProcessInFlightExit {
...
     )
         private
     {
-        // We flag _all_ inputs regardless of whether it is piggybacked
-        // If exiting from output, all inputs are considered spent and can only exit from output in future.
-        // If exiting from input, to keep things simple, all users must piggyback the input at the same time,
-        // instead of re-starting the IFE and then re-piggyback their input.
         uint256 inputNumOfTheToken;
         for (uint16 i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
```
boolafish: does not seem necessary to check whether amount exists or not? We set data for all inputs and outputs in start IFE no matter the amount value.

boolafish: ah....okay. So we use fixed array in IFE exit data so it is actually needed.

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -429,17 +404,46 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
...
 
-        describe('When the exit is non canonical, and inputs are piggybacked', () => {
+        describe('When any input is spent, given the challenge game result is canonical', () => {
+            beforeEach(async () => {
+                const exit = await buildInFlightExitData();
+                this.dummyExitId = 666;
+                exit.isCanonical = true;
+                await this.exitGame.setInFlightExit(this.dummyExitId, exit);
+
+                // flags the first input and piggyback the two ETH inputs (first and second)
```
pgebal: piggyback -> piggybacks

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -429,17 +404,46 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
...
                 this.dummyExitId = 666;
                 this.exit = await buildInFlightExitData();
                 this.exit.isCanonical = false;
 
-                // piggybacks all three inputs
-                this.exit.exitMap = (2 ** 0) | (2 ** 1) | (2 ** 2);
-
                 await this.exitGame.setInFlightExit(this.dummyExitId, this.exit);
 
+                // piggybackes the first input for ETH and third for ERC20
```
pgebal: Duplicated line.

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -488,6 +482,24 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
                 expect(postBalance).to.be.bignumber.equal(expectedBalance);
             });
 
+            it('should flag and only flag piggybacked inputs with the same token as spent', async () => {
```
pgebal: flag and only flag -> flag only

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -429,17 +404,46 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
...
+                exit.isCanonical = true;
+                await this.exitGame.setInFlightExit(this.dummyExitId, exit);
+
+                // flags the first input and piggyback the two ETH inputs (first and second)
+                await this.exitGame.proxyFlagOutputSpent(TEST_OUTPUT_ID_FOR_INPUT_1);
+                await this.exitGame.setInFlightExitInputPiggybacked(this.dummyExitId, 0);
+                await this.exitGame.setInFlightExitInputPiggybacked(this.dummyExitId, 1);
+            });
+
+            it('should be treat as non canonical', async () => {
```
pgebal: treat -> treated

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -429,17 +404,46 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
             });
         });
 
-        describe('When the exit is non canonical, and inputs are piggybacked', () => {
+        describe('When any input is spent, given the challenge game result is canonical', () => {
```
pgebal: I understand that's the scenario described in the issue. Right?

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -429,17 +404,46 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
...
+                // flags the first input and piggyback the two ETH inputs (first and second)
+                await this.exitGame.proxyFlagOutputSpent(TEST_OUTPUT_ID_FOR_INPUT_1);
+                await this.exitGame.setInFlightExitInputPiggybacked(this.dummyExitId, 0);
+                await this.exitGame.setInFlightExitInputPiggybacked(this.dummyExitId, 1);
+            });
+
+            it('should be treat as non canonical', async () => {
+                const { logs } = await this.exitGame.processExit(this.dummyExitId, VAULT_ID.ETH, ETH);
+                const inputIndexForInput2 = 1;
+                await expectEvent.inLogs(
```
pgebal: It would be good, if we could check that there was no event for inputIndex 0.

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
```diff
@@ -354,14 +339,18 @@ contract('PaymentInFlightExitRouter', ([_, ifeBondOwner, inputOwner1, inputOwner
 
                 this.dummyExitId = 666;
                 this.exit = await buildInFlightExitData();
+                await this.exitGame.setInFlightExit(this.dummyExitId, this.exit);
```
pgebal: Everywhere in this file we use dummyExitId with value 666. How about, we put it into a constant?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -133,14 +138,14 @@ library PaymentProcessInFlightExit {
...
-            if (exit.inputs[i].token == token) {
+            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
                 inputNumOfTheToken++;
             }
         }
         bytes32[] memory outputIdsOfInputs = new bytes32[](inputNumOfTheToken);
         uint sameTokenIndex = 0;
         for (uint i = 0; i < MAX_INPUT_NUM; i++) {
-            if (exit.inputs[i].token == token) {
+            if (exit.inputs[i].token == token && exit.inputs[i].amount > 0) {
```
pgebal: Why was that added check necessary?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -158,9 +163,8 @@ library PaymentProcessInFlightExit {
         pure
         returns (bool)
     {
-        // The check to determine whether input is spent, performed in 'isAnyInputSpent'
-        // For this reason, no need to check again here.
         return withdrawal.token == token &&
+                exit.isInputPiggybacked(index) &&
```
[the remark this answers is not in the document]
boolafish: aww, I need to add check on the `isOutputSpent` (I think) as we are not omitting the process anymore but I copied the wrong line @@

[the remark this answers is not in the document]
boolafish: also, test added for the case that if already flagged, would not withdraw

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
boolafish: yup. Also some other tests as put in the description.

plasma_framework/test/src/exits/payment/PaymentProcessInFlightExit.test.js
boolafish: added the test under non-canonical test cases (describe) instead of here.

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
boolafish: I want to be consistent. We flag only when the input is not empty however we check inputs even if it is empty. It only works because empty input --> outputId == 0 --> the flag of that would always be false. I think it is better to not check the flag for that.

boolafish: Also, I abstract the check of `amount > 0` to be more explicit checking whether it is an empty withdraw data and have a function in library for that.

plasma_framework/contracts/mocks/exits/payment/PaymentInFlightExitModelUtilsMock.sol
```diff
@@ -13,6 +13,43 @@ contract PaymentInFlightExitModelUtilsMock {
...
+        } else {
+            revert("target should be either inputs or outputs");
+        }
+    }
+
+    function stringEquals(string memory a, string memory b) private pure returns (bool) {
+        return keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b));
+    }
+
+
```
pgebal: We usually use one empty line for spacing. Here we have 2.

plasma_framework/test/src/exits/payment/PaymentInFlightExitModelUtils.test.js
```diff
@@ -3,22 +3,226 @@
...
+            for (let i = 0; i < MAX_INPUTS; i++) {
+                const actual = await modelUtils.isInputEmpty(i);
+                expect(actual).to.be.true;
+            }
+        });
+
+        it('returns false when the outputId is not empty', async () => {
+            const modelUtils = await PaymentInFlightExitModelUtils.new(0, 0);
+
+            const dataWithNonEmptyOutputId = {
```
pgebal: I would make a constant `emptyWithdrawData` on a higher scope and in each test I would just update one value in it. That would save some lines of code.

Issue #474, opened 2019-11-22
Add support for Tenderly through automation


jbunce:
```
[ ] bug report
[x] feature request
```

## Current Behavior
We have to push deployed contracts to Tenderly manually for monitoring. This means if it's forgotten and we lose the build artifacts we don't have contract level monitoring for that environment.

## Suggested Fix
This is intended to be achieved through a module/migration within Truffle.

1) Introduce a flag that disables Tenderly (assume that a Tenderly deployment is desired by default). Suggestion: `export TENDERLY=false`
2) Add support for obtaining the Tenderly authentication token. Suggestion: `export TENDERLY_TOKEN=foo`
3) Add support for creating the Tenderly project: `export TENDERLY_PROJECT=bar`
4) Add support for creating the Tenderly project:
```
tenderly init --create-project --project $TENDERLY_PROJECT
```
5) After the contracts have been deployed issue:
```
tenderly login --authentication-method=token --token=$TENDERLY_TOKEN
```
6) Push the contracts to Tenderly:
```
tenderly push
```

This would fit with the work in https://github.com/omisego/devops/issues/242

thec00n:
@jbunce It's not clear to be how the tenderly integration steps fit in the current repository and how this is then executed during deployment. Is it that we define the steps in package.json and then you call them sequentially?

```
1.) install tenderly
2.) set env variables
3.) create the project 
4.) truffle migrate
5.) tenderly login 
6.) tenderly push 
```

jbunce:
Hey @thec00n. Without actually going through the process myself while implementing this I'm not sure. If you want to take a stab at this let me know and I'll see if I can help.

jbunce:
From a conversation @thec00n and I had in person:

- We've agreed to facilitate this for development environments
- Manual process for production environments
- Implementation in JS with exec/subprocess
- `plasma-contracts` change required so we don't pass authority key material into Truffle

Feel free to add if I've missed anything :)

thec00n:
Closing this issue as it addressed with #493. Splitting up dev/prod deployments is tracked here https://github.com/omisego/plasma-contracts/issues/499

PR #493, opened 2019-11-29
tenderly installation and setup

thec00n:
Addresses issue #474 

- Increased gas limit for local deployment
- Added installation and setup script for tenderly
- The tenderly flow is deactivated by default so it does not run in CI

Installation requires `tar` and `curl` to be installed.


plasma_framework/truffle-config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
             host: process.env.ETH_CLIENT_HOST || '127.0.0.1',
             port: process.env.ETH_CLIENT_PORT || 8545,
             from: process.env.DEPLOYER_ADDRESS,
-            gas: 6000000,
+            gas: 10000000,
```
boolafish: why do we need to increase the gas?

thec00n: I did get out of gas error when I tried to deploy locally.

boolafish: ah...okay, just check current deployment gas and indeed is slightly exceeding 6m. Can you just remove that line instead. It would be using default truffle gas cost of ~6.7m which is ganache default. I would not want to increase to 10m as that is too high and I don't think any chain is using such high gas limit at the moment.

plasma_framework/migrations/7_tenderly_setup.js
```diff
@@ -0,0 +1,43 @@
...
+    // eslint-disable-next-line no-unused-vars
+    deployer,
+    _,
+    // eslint-disable-next-line no-unused-vars
+    [deployerAddress, maintainerAddress, authorityAddress],
+) => {
+    const tenderly = process.env.TENDERLY || false;
+
+    if (tenderly) {
+        const installationStatus = execSync('curl https://raw.githubusercontent.com/Tenderly/tenderly-cli/master/scripts/install-linux.sh | sh', { encoding: 'utf-8' }); // the default is 'buffer'
```
jbunce: What happens if Tenderly is already installed?

thec00n: It would print a message that it is already installed.

boolafish:
@thec00n what would happen if we deploy new contract after `7_tenderly_setup.js`, for example, `8_deploy_payment_v2_exit_game.js`. Do we need another `tenderly push` after `8` or what should we do?

PR #494, opened 2019-11-29
Fix python tests make instructions in README

### LINKED ISSUE #485: Readme instructs to use non-existent make command
The readme file says to run

    make dev

However, there is no make target called `dev` in the makefile.

Btw. I think you should change the other make lines too bye prepending `sudo` as it currently is not evident that those commands need to be run as root.

pik694:
Fix running python tests `make` instructions in README file. 

Closes #485

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Run truffle, passing in the network e.g.
 
 
 ### Building and running the python tests
+All the code is in the `plasma_framework/python_tests` directory, so go there first.
 
 Installing dependencies needed for compilation:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,8 @@ Run truffle, passing in the network e.g.
 
 
 ### Building and running the python tests
+We suggest running the following commands with an active python virtual environment ex. `virtualenv`.
```
boolafish: I think the standard way in python3 has moved to use `venv` instead

https://docs.python.org/3/library/venv.html#venv-def

pik694: Thanks, changed it :)

Issue #194, opened 2019-07-30
Document limitation of the system (ALD)


boolafish:
Using this as a place holder first and also a tracker of the task.
We want this for both us and the auditor and our user to be easier to track our system.

boolafish:
[Note]
- Global assumption of the design
  - outputId should be unique
  - tx bytes should be unique aside from deposit tx

PR #495, opened 2019-11-29
docs: add new tx type check list

boolafish:
### Note
This adds a doc for the check list when adding or designing new tx types.

### About Reviewing
- I am putting many reviewers for this to make sure nothing is missed
- Was collecting from: 
    - https://github.com/omisego/docs/issues/27
    - https://github.com/omisego/plasma-contracts/issues/194

### DIFF
--- a/plasma_framework/docs/design/add-new-tx-type-check-lists.md
+++ b/plasma_framework/docs/design/add-new-tx-type-check-lists.md
@@ -1,0 +1,16 @@
+# Check List For Adding New Transaction Types
+
+This document provides some points that should be taken into consideration whenever designing a new transaction type. Since there are some known restrictions of the Plasma Framework design, please make sure that the newly created tx type and exit game does not break the global assumptions and limitations.
+
+1. Aside from deposit transactions, all transactions should be unique. 
+   - To be unique it means that the encoded transaction should not collide with another transaction.
+   - Usually, a tx with valid inputs would be unique. An input would hold the information of an output identifier that can point to an output. A valid output identifier should be unique as a result making the transaction unique as well.
+   - Deposit transaction is a known exception of this as it does not have inputs and the outputs can be holding with same `amount` and `token`. See: [previous related issue of deposit tx not being unique](https://github.com/omisego/plasma-contracts/issues/80)
+2. `OutputId` should be unique for all outputs. This means that for each output, the current `outputId` schema should be unique and not collide with each other. We use `outputId` to flag the output in `PlasmaFramework` thus require it to be unique. The schema of `OutputId` is:
+   - Deposit Output: `hash(txBytes, outputIndex, utxoPos)`
+   - Other Output: `hash(txBytes, outputIndex)`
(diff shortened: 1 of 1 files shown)


plasma_framework/docs/design/add-new-tx-type-check-lists.md
```diff
@@ -0,0 +1,9 @@
+# Check List For Adding New Transaction Types
+
+This document provides some check lists that whenever we are designing a new transaction type. Since there are some known restriction of the Plasma Framework design, please make sure that new tx type and exit game does not break the global assumptions and limitations.
+
+1. Aside from deposit transactions, all transactions should be unique. This promises the `OutputId` to be unique as `OutputId` uniqueness is a global assumption placed to `PlasmaFramework`.
```
pik694: What does `uniqueness` of a transaction mean? Do you mean the type or the two txs of the same type to look the same? I would also write sth about how to design a new `output id`.

plasma_framework/docs/design/add-new-tx-type-check-lists.md
```diff
@@ -0,0 +1,9 @@
...
+1. output index is max to `TX_OFFSET`  (10000) of the `UtxoPos` schema. As a result, no more than such amounts of inputs or outputs in one transaction.
```
pik694: Start the first word with an upper case.

pik694: actually I think we can have more outputs than `TX_OFFSET`, only number of outputs are strictly limited.

plasma_framework/docs/design/add-new-tx-type-check-lists.md
```diff
@@ -0,0 +1,9 @@
...
+1. `ExitId` for the new exit game contract must promise to be not collide with all existing `ExitId` schemas used in another exit game contracts. (Each exit game contract can potentially introduce new schema if needed. Not recommended though)
```
pik694: Rather than using parentheses I would add sth like `NOTICE:`.

plasma_framework/docs/design/add-new-tx-type-check-lists.md
```diff
@@ -0,0 +1,9 @@
...
+1. Please always define new transaction type and new output type. No reusing on the output type. For example, even an extension of Payment V1 to Payment V2, the output should be using different output type. (see previous discussion: [here](https://github.com/omisego/plasma-contracts/issues/214#issuecomment-526466192))
```
pik694: When defining a new tx type, a new output type should be introduced along. Do not reuse an already existing output type.

plasma_framework/docs/design/add-new-tx-type-check-lists.md
boolafish: > actually I think we can have more outputs than TX_OFFSET

Do you mean inputs? hmmm...I think you're right on that

pik694: You are right - _inputs_. Corrected it.

PR #496, opened 2019-12-02
Remove output guard handler

boolafish:
### Note
As we are planning to scope down the `PaymentExitGame` contract, we are removing some DEX supporting feature. This PR removes the output guard related logic, so now the PaymentExitGame only reads the `outputGuard` field as `owner` directly. 

- This also removed the extendibility e2e test on extend to DEX as that is not supported by current feature set of PaymentExitGame anymore.
- Next step is to swap `tx finalization` with a simplified version and remove all confirm sig related optional args.
- Reading with commit order is recommended. Will squash and merge for this PR.
- The reducing scope plan: https://omisego.atlassian.net/wiki/spaces/RES/pages/129105932/Reduce+the+scope+of+Payment+ExitGame+V1

### Warning
Would recommend clients to not integration with the result of this PR and wait for the simplification of `txVerifier` and then do the change at once to reduce some effort. However, it is still okay to migration to this one, just remove `outputGuard` related args.


boolafish:
checked, it is not missing any new branches. Both master and this PR is missing 3 branches. Just the total amount of branches decrease as a result lowering the coverage.

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeStandardExit.sol
```diff
@@ -107,34 +101,18 @@ library PaymentChallengeStandardExit {
...
-            guard: output.outputGuard,
-            preimage: data.args.outputGuardPreimage
-        });
-        require(outputGuardHandler.isValid(outputGuardData),
-                "Output guard information is invalid");
-
         uint256 challengeTxType = WireTransaction.getTransactionType(data.args.challengeTx);
         uint8 protocol = data.controller.framework.protocols(challengeTxType);
+
+        //TODO: simplify tx finalizeion to be MoreVP only
```
kevsul:
```suggestion
        //TODO: simplify tx finalization to be MoreVP only
```

plasma_framework/test/endToEndTests/Extendiblility.e2e.test.js
```diff
@@ -1,312 +0,0 @@
-const EthVault = artifacts.require('EthVault');
```
kevsul: Do we really want to delete this entire test? Now we have no e2e test for extendibility 😞

boolafish: I am thinking to compensate that with the fee transaction e2e coming soon. So it would be payment v1 -> payment v2 + fee.

boolafish: hmmmm......so asked the fee team and seem like we are more on the side of not needing payment v2 for fee. sigh....any cool idea for extendibility test?

kevsul: @boolafish Let's just merge this and add a new extensibility test later. I don't want to hold up this PR any longer 😄

boolafish: Thanks :)

boolafish:
ref #501

Issue #501, opened 2019-12-04
Placeholder for refactoring proposal on reducing scope of Payment ExitGame


boolafish:
This doc was originally in confluence: [here](https://omisego.atlassian.net/wiki/spaces/RES/pages/129105932/Reduce+the+scope+of+Payment+ExitGame+V1). Put an issue for it for public availability and easier reference with PR and issues. 

### Background

We start the audit of Payment ExitGame with the ability to support DEX feature in the future without the need to re-write the main code. As the time of this doc is written, the latest commit of plasma-contract is 493ddeadc821fae7293cfc3088939eddf89cb559. This is still with the ability to seamlessly support DEX feature extension. Here is a demo test of extension to DEX with existing Payment ExitGame implementation: https://github.com/omisego/plasma-contracts/blob/493ddeadc821fae7293cfc3088939eddf89cb559/plasma_framework/test/endToEndTests/Extendiblility.e2e.test.js#L35

During the start of the project, DEX was the next highest priority in the schedule. As a result, it makes sense to take DEX into consideration at that time. However, as priority switches as things changes. Also, the spec of the DEX is changed from Restricted Custody to Velocity which the detail of it is not fine tuned yet.

Current implementation takes quite a lot of complexity into consideration. One is that sadly the tx type dependency is not trivial to solve if we want the code base to be a more generalized solution which is ignorant of tx types and the protocol (MVP, MoreVP) that they are using. Also, some abstraction we introduced have apis or arguments only for MVP protocol, which is unclear when we would ever start to use that.

As auditors have pointed out, current code base have too many extra abstraction/complexity that does not seem straight forward to understand as there is no concrete example of using them as the first version of our launching scope would be without DEX feature. This is some extra unnecessary risks to bear for our first launch. Also, those feature would not be well tested as other feature would be now.

Following are some issue related:
https://github.com/omisego/plasma-contracts/issues/456
https://github.com/omisego/plasma-contracts/issues/457
https://github.com/omisego/plasma-contracts/issues/458
https://github.com/omisego/plasma-contracts/issues/471
https://github.com/omisego/plasma-contracts/issues/472
 
As a result, to be able to run security on the launch, and to not be burned by those not fully tested feature, we are considering to reduce the scope of Payment ExitGame V1 contract by removing the ability of seamlessly supporting DEX.

### Issue 457: Recommendation: Remove optional arguments and clean unused code
https://github.com/omisego/plasma-contracts/issues/457

[...]

PR #502, opened 2019-12-04
WireTransaction refactor

kevsul:
The intention is that WireTransaction is a generic transaction, that doesn't make much assumptions about the actual format. It assumes that a transaction is like this
```
    struct Transaction {
        uint256 txType;
        RLPReader.RLPItem[] inputs;
        RLPReader.RLPItem[] outputs;
        RLPReader.RLPItem txData;
    }
```
In RLP an RLP item can be an array, so the `txData` field can be extended to whatever a transaction needs.

It also defines an Output struct for convenience.
```
    struct Output {
        uint256 outputType;
        bytes20 outputGuard;
        address token;
        uint256 amount;
    }
```

Most tx types will use this format (maybe?), but it can be extended as well

I'll rename WireTransaction to GenericTransaction, but it's easier to see the diff if I don't change it yet.

I haven't really done much optimisation around multiple decodings of the same transaction yet - there are probably some places that we do this that can be avoided.

ref #501


boolafish:
oh....I wish we'd come up with this idea of using RLP as struct field earlier! This is a smart and great way of abstract the data dependency out.

boolafish:
> I'll rename WireTransaction to GenericTransaction, but it's easier to see the diff if I don't change it yet.

Just a reminder, documents need to be updated as well. Probably the doc change can be in different PR.

plasma_framework/contracts/src/vaults/verifiers/EthDepositVerifier.sol
```diff
@@ -34,7 +31,8 @@ contract EthDepositVerifier is IEthDepositVerifier {
         require(decodedTx.outputs[0].token == address(0), "Output requires correct currency (ETH)");
         require(decodedTx.outputs[0].outputType == supportedOutputType, "Invalid output type");
 
-        address depositorsAddress = decodedTx.outputs[0].owner();
+        // address depositorsAddress = decodedTx.outputs[0].owner();
```
boolafish: seems like a leftover

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -41,36 +39,47 @@ library PaymentTransactionModel {
      * @return A decoded PaymentTransaction struct
      */
     function decode(bytes memory _tx) internal pure returns (PaymentTransactionModel.Transaction memory) {
```
boolafish: Some really minor comments on the document. 
1. Can we change the beginning from `/*` to `/**`? This makes the IDE happy to change color :)
1. `tx` -> `_tx`

plasma_framework/test/src/transactions/PaymentTransactionModel.test.js
```diff
@@ -136,6 +134,38 @@ contract('PaymentTransactionModel', () => {
         const encoded = web3.utils.bytesToHex(transaction.rlpEncoded());
         await expectRevert(this.test.decode(encoded), 'Output amount must not be 0');
     });
+
+    describe('decodeOutput', () => {
```
boolafish: Minor: probably should move the two tests above inside here too:
- should fail when an output amount is zero
- should fail when an output type is zero

plasma_framework/test/src/transactions/GenericTransaction.test.js
```diff
@@ -4,17 +4,17 @@ const { BN, constants, expectRevert } = require('openzeppelin-test-helpers');
 
 const { PaymentTransaction, PaymentTransactionOutput } = require('../../helpers/transaction.js');
 
-const WireTransaction = artifacts.require('WireTransactionWrapper.sol');
+const GenericTransaction = artifacts.require('GenericTransactionWrapper.sol');
 
-contract('WireTransaction', () => {
+contract('GenericTransaction', () => {
```
boolafish: Some comments on this test file:
1. Remove the test of `getTransactionType`. It is no longer a function of the library.
1. Add tests for `decode` and `decodeOutput`. I think some of them are currently covered in the `PaymentTransactionModel` tests. Please consider move those tests related to `GenericTransaction` here or duplicate the tests.

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -41,36 +39,47 @@ library PaymentTransactionModel {
...
-        bytes32 metaData = rlpTx[3].toBytes32();
+        bytes32 metaData = btx.txData.toBytes32();
 
-        return Transaction({txType: txType, inputs: inputs, outputs: outputs, metaData: metaData});
+        return Transaction({txType: btx.txType, inputs: inputs, outputs: outputs, metaData: metaData});
+    }
+
+    function decodeOutput(RLPReader.RLPItem memory output) internal pure returns (GenericTransaction.Output memory) {
+        RLPReader.RLPItem[] memory outputRlpList = output.toList();
+        require(outputRlpList.length == 4, "Output must have 4 items");
```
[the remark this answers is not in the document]
kevsul: The difference is that a PaymentTransaction.Output must have exactly 4 items, whereas a GenericTransaction.Ouput may have more than 4 (if the concrete output format extends the struct and adds another field).

Maybe a better way to do this would be to follow the same model as the Transaction struct, and have an `outputData` field that can be extended as necessary?
However, if we do add this, it can't be optional - PaymentTransactions would have to always include a value (e.g. `0`) for this field.

What do you think?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -41,36 +39,47 @@ library PaymentTransactionModel {
...
-
-        RLPReader.RLPItem[] memory rlpOutputs = rlpTx[2].toList();
-        require(rlpOutputs.length <= MAX_OUTPUT_NUM, "Transaction outputs num exceeds limit");
+        GenericTransaction.Transaction memory btx = GenericTransaction.decode(_tx);
+        return fromGeneric(btx);
+    }
 
-        uint txType = rlpTx[0].toUint();
-        require(txType > 0, "Transaction type must not be 0");
+    function fromGeneric(GenericTransaction.Transaction memory btx) internal pure returns (PaymentTransactionModel.Transaction memory) {
```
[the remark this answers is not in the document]
kevsul: Yeah, the idea is that if we already have a GenericTransaction and we need to convert it to a PaymentTransaction, we can use this method.

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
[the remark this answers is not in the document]
boolafish: Ah...I see the difference now.

Btw, Do you mean the following structure?

```
struct Output {
        uint256 outputType;
        bytes20 outputGuard;
        address token;
        uint256 amount;
        RLPReader.RLPItem outputData;
}
```
Yeah....I am a bit not sure whether it justify the non-optional part. Though ugly, currently it is possible to use `GenericTransaction.decodeOutput` to get the first four fields first and then decode further if the output holds more field.

However, after seeing your trick of using RlpItem in transaction, If it is even possible I would be reeeeeally happy if the following is possible without bringing too much mess:

```
struct Output {
        uint256 outputType;
        RLPReader.RLPItem outputData;
}
```
And then introduce sth like `FungibleTokenOutputModel` to cover the `outputData`:
```
function getOutput(GenericTransaction.Transaction memory transaction, uint16 outputIndex) internal pure returns (FungibleOutput memory);

function decodeOutput(RLPItem memory outputRlp) internal pure returns (FungibleOutput memory);
```

Then our assumptions in implementation would make a lot more sense that `PaymentTx` only deal with fungible token outputs. This would make supporting ERC721 a lot easier with this structure, just need to introduce another `NonFungibleTokenOutputModel`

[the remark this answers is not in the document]
kevsul: Yes, I like this idea too 😄 
I don't think it's a huge change, and now is definitely the time to do it. Let me try it out and see what it looks like

plasma_framework/contracts/src/transactions/GenericTransaction.sol
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+    * @dev Decodes an RLP encoded transaction into the generic format.
+    */
+    function decode(bytes memory transaction) internal pure returns (Transaction memory) {
+        RLPReader.RLPItem[] memory rlpTx = transaction.toRlpItem().toList();
+        require(rlpTx.length == TX_NUM_ITEMS, "Invalid encoding of transaction");
+        uint256 txType = rlpTx[0].toUint();
+        require(txType > 0, "Transaction type must not be 0");
```
thec00n: Really minor issue: We are mixing `!=0` and `>0` when we check for fields in transactions. Could we stick to either the value must be greater than zero or the value can not be zero?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -37,40 +35,58 @@ library PaymentTransactionModel {
...
-        uint txType = rlpTx[0].toUint();
-        require(txType > 0, "Transaction type must not be 0");
+    /**
+     * @notice Converts a GenericTransaction to a PaymentTransaction
+     * @param genericTx A GenericTransaction.Transaction struct
+     * @return A PaymentTransaction.Transaction struct
+     */
+    function fromGeneric(GenericTransaction.Transaction memory genericTx) internal pure returns (PaymentTransactionModel.Transaction memory) {
+        require(genericTx.inputs.length <= MAX_INPUT_NUM, "Transaction inputs num exceeds limit");
+        require(genericTx.outputs.length <= MAX_OUTPUT_NUM, "Transaction outputs num exceeds limit");
```
thec00n: Is there an exit flow where zero outputs is valid?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
```diff
@@ -37,40 +35,58 @@ library PaymentTransactionModel {
...
+    /**
+     * @notice Decodes a transaction output
+     * @param output The RLP encoded output
+     * @return The decoded GenericTransaction.Output struct
+     */
+    function decodeOutput(RLPReader.RLPItem memory output) internal pure returns (GenericTransaction.Output memory) {
+        RLPReader.RLPItem[] memory outputRlpList = output.toList();
+        require(outputRlpList.length == 4, "Output must have 4 items");
+        GenericTransaction.Output memory decodedOutput = GenericTransaction.decodeOutput(outputRlpList);
+        require(decodedOutput.amount != 0, "Output amount must not be 0");
```
thec00n: We check that the amount can not be zero here and that that output type can not be zero in `GenericTransaction.decodeOutput`. Is this intentional and does that mean that we plan to have specific transactions that can have zero amount?

thec00n:
I like the changes introduced in this PR especially the two decoding functions that either return a specific tx or a generic tx. This makes the transaction abstraction much clearer and explicit.

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
kevsul: It's intentional. GenericTransaction should make as few assumptions as possible about the transaction.

kevsul:
Following on from the above suggestions, I've made GenericTransaction.Output truly generic, and added the FungibleTokenOutputModel to handle outputs of `[outputGuard, token, amount]`.
I think it greatly increases flexibility for adding new transaction types going forward.

The main change on the client side is the rlp encoding; output data is nested in a list inside the output, e.g.
```
output: [outputType, [outputGuard, token, amount]]
```
instead of 
```
output: [outputType, outputGuard, token, amount]
```

@pdobacz I'd love to get your feedback on this change 🙏  
(sorry for the huge PR, the interesting changes are in `GenericTransaction.sol`, `FungibleTokenOutputModel` and `PaymentTransactionModel`)

plasma_framework/test/src/transactions/FungibleTokenOutput.test.js
```diff
@@ -0,0 +1,116 @@
...
+            ]);
+
+            await expectRevert(
+                this.test.decodeOutput(invalidOutput),
+                'Output data must have 3 items',
+            );
+        });
+
+        it('should fail when amount is 0', async () => {
+            // return [this.outputType, [this.outputGuard, this.token, this.amount]];
```
boolafish: leftover?

thec00n:
I have reservations of adding new code at this point that is not strictly needed for this release. What's the rationale of introducing `FungibleTokenOutputModel` as part of this PR?

plasma_framework/contracts/src/transactions/PaymentTransactionModel.sol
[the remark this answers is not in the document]
boolafish: @thec00n 

> What's the rationale of introducing FungibleTokenOutputModel as part of this PR?

Here is the discussion of the model design👆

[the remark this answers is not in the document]
boolafish: Previous design of `WireTransaction` has the issue of not being really generalized enough as all outputs need to conform to same data structure. We've already seen that it would be not easy to support multi ERC721 <> ERC20 atomic swap tx.

We did not realize this RLPItem field trick until now. By leaving the responsibility to each output model itself, it leave us the ability to add whatever new output model we need in the future.

plasma_framework/test/src/transactions/GenericTransaction.test.js
```diff
@@ -0,0 +1,135 @@
+const GenericTransaction = artifacts.require('GenericTransactionWrapper.sol');
+const rlp = require('rlp');
+const { expect } = require('chai');
+const { BN, expectRevert } = require('openzeppelin-test-helpers');
+
+contract('GenericTransaction', () => {
+    const EMPTY_BYTES32 = `0x${Array(64).fill(0).join('')}`;
```
boolafish: minor: consider use the `EMPTY_BYTES_32` from `constants.js`

kevsul:
> I have reservations of adding new code at this point that is not strictly needed for this release. I can see how the change to `GenericTransaction.Output` makes sense but what's the rationale of introducing `FungibleTokenOutputModel` as part of this PR?

I understand and share your reservations about adding new code at this point, but I think in this case it's worth it. And really it's a refactor of `WireTransaction`, `PaymentTransactionOutput` and `PaymentTransaction` to `GenericTransaction`, `FungibleTokenOutput` and `PaymentTransaction`.

The rationale for the change is:

- The old WireTransaction was intended to be generic, but it wasn't really. It made assumptions about the format of a transaction output. We were never happy about these assumptions because they would force us to either hard upgrade or invent some ugly workaround when we add NFT (ERC-721, etc.) support.

- The new GenericTransaction makes no assumptions other than the basic format of a transaction, including txType and outputType. Just knowing these types will allow us to look up a registry or convert the generic tx to a concrete tx type.

- `FungibleTokenOutput` provides a common output format for fungible token transactions. It's not strictly necessary, but most transaction types will deal with fungible tokens (payment V1, V2, Dex deposit, Dex settlement), so having it here will avoid a lot of duplicate code in the future.

pdobacz:
> @pdobacz I'd love to get your feedback on this change pray

I'm a liitle bit out of brain now, but this specifically looks good and the rationale here in general makes sense to me too.

I'm assuming that `txData` will today be used as `bytes32 metadata` - is this correct? A thought that I have is to consider an alternative, where metadata is split away further from anything that our code interprets. Basically `metadata` should be "some bytes that our code never ever looks into (except hashing, encoding, decoding). So separating it as verbosely as possible, from anything that **is** interpreted somehow by our code could be considered.

boolafish:
> So separating it as verbosely as possible, from anything that is interpreted somehow by our code could be considered.

I am actually thinking to make use of that field for fee tx. Current design of fee tx has a weird output for nonce/blockNum. I think it makes more sense to put that in that extra field as an extension of tx. However, I am not really sure whether we need to verify the correctness of nonce/blockNum or not....

see https://github.com/omisego/feat-fee/pull/47#discussion_r355914884

kevsul:
> I'm assuming that `txData` will today be used as `bytes32 metadata` - is this correct? A thought that I have is to consider an alternative, where metadata is split away further from anything that our code interprets.

Yeah, I was thinking that each transaction type can define if it supports metadata or not, and if so then it will define where it goes in the txdata field.

But you're suggesting to keep it at the generic level, so that all tx types support metadata in the same way, right? i.e.
```
[txtype, inputs, outputs, txdata, metadata]
```

I can see arguments for both, but consistency should probably win out, so I'm happy to go with this way.

kevsul:
> Generally LGTM. Only concern is probably auditors would ask why a always empty field for `txData`🤦‍♂️ and then recommend to remove it just like other recommendations.....

You're going to use it in the feeTx, right? That's a good example of why we need it

plasma_framework/contracts/src/transactions/GenericTransaction.sol
```diff
@@ -0,0 +1,96 @@
...
+    uint8 constant private TX_NUM_ITEMS = 5;
+
+    using RLPReader for bytes;
+    using RLPReader for RLPReader.RLPItem;
+
+    struct Transaction {
+        uint256 txType;
+        RLPReader.RLPItem[] inputs;
+        Output[] outputs;
+        RLPReader.RLPItem txData;
```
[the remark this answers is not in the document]
kevsul: Yeah, elixir-omg will have to be updated to support the new format.

Issue #471, opened 2019-11-22
Recommendation: Remove `TxFinalizationModel` and `TxFinalizationVerifier`. Implement stronger checks in `Merkle`


wadeAlexC:
## Issue Type 

```
[ ] bug report
[X] feature request
```

#### Description

`TxFinalizationVerifier` is an abstraction around the block inclusion check needed for many of the features of plasma exit games. It uses a struct defined in `TxFinalizationModel` as inputs to its two functions: `isStandardFinalized` and `isProtocolFinalized`. 

`isStandardFinalized` returns the result of an inclusion proof. Although there are several branches, only the first is used:

https://github.com/omisego/plasma-contracts/blob/493ddeadc821fae7293cfc3088939eddf89cb559/plasma_framework/contracts/src/exits/utils/TxFinalizationVerifier.sol#L19-L32

`isProtocolFinalized` is unused:

https://github.com/omisego/plasma-contracts/blob/493ddeadc821fae7293cfc3088939eddf89cb559/plasma_framework/contracts/src/exits/utils/TxFinalizationVerifier.sol#L34-L47

The abstraction used introduces branching logic and requires several files to be visited to fully understand the function of each line of code: `ITxFinalizationVerifier`, `TxFinalizationModel`, `TxPosLib`, `Protocol`, `BlockController`, and `Merkle`. Additionally, the abstraction obfuscates the underlying inclusion proof primitive when used in the exit game contracts. `isStandardFinalized` is not clearly an inclusion proof, and `isProtocolFinalized` simply adds confusion.

Finally, the abstraction may have ramifications on the safety of `Merkle.sol`. As it stands now, `Merkle.checkMembership` should never be called directly by the exit game controllers, as it lacks an important check made in `TxFinalizationVerifier.checkInclusionProof`:

https://github.com/omisego/plasma-contracts/blob/493ddeadc821fae7293cfc3088939eddf89cb559/plasma_framework/contracts/src/exits/utils/TxFinalizationVerifier.sol#L49-L59

By introducing the abstraction of `TxFinalizationVerifier`, the input validation performed by `Merkle` is split across multiple files, and the reasonable-seeming decision of calling `Merkle.checkMembership` directly becomes unsafe. In fact, this occurs in one location in the contracts:

[...]

PR #503, opened 2019-12-05
refactor: simplify tx finalization

boolafish:
### Note
- Create new lib: `MoreVpFinalization` that only checks the finalization of MoreVP protocol.
- Migrate existing codes to MoreVpFinalization.
- It is pretty much not possible to test `isProtocolFinalized` in each controllers. This is because in MoreVP, it is basically by default true (as long as tx exists). When given empty value in test, it would always fail at decoding first. As a result, instead of `require(isProtocolFinalized)`, I changed it to `assert(isProtocolFinalized)` to emphasize that we are pretty sure that it should be protocol finalized.
- Remove the optional args in spending condition interface.
- Remove confirm sig, optional args from each exit game function args.

ref #501, #471


boolafish:
Coverage drop are from the introduction of `assert` on `isProtocolFinalized`

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -239,20 +221,13 @@ library PaymentStartInFlightExit {
...
-                protocol: protocol,
-                txBytes: exitData.inputTxs[i],
-                txPos: exitData.inputUtxosPos[i].txPos(),
-                inclusionProof: exitData.inputTxsInclusionProofs[i],
-                confirmSig: bytes(""),
-                confirmSigAddress: address(0)
-            });
-            require(exitData.controller.txFinalizationVerifier.isStandardFinalized(finalizationData),
-                    "Input transaction is not standard finalized");
+            bool isStandardFinalize = MoreVpFinalization.isStandardFinalized(
```
kevsul:
```suggestion
            bool isStandardFinalized = MoreVpFinalization.isStandardFinalized(
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartInFlightExit.sol
```diff
@@ -239,20 +221,13 @@ library PaymentStartInFlightExit {
...
+                exitData.controller.framework,
+                exitData.inputTxs[i],
+                exitData.inputUtxosPos[i].txPos(),
+                exitData.inputTxsInclusionProofs[i]
+            );
+            require(isStandardFinalize, "Input transaction is not standard finalized");
```
kevsul:
```suggestion
            require(isStandardFinalized, "Input transaction is not standard finalized");
```

Issue #150, opened 2019-07-04
Run python tests on plasma framework


pik694:
In #116, we decided to  rewrite all current tests into `js`, but in order to ensure that the rewrite is secure and works as intended we want the old python tests to run on the new contracts. 

Once we have rewritten all the tests we will eventually delete python tests, but for now we should keep them, but the tests should now test `plasma framework` instead of `RootChain`.

bg1156:
*but the tests should now test plasma framework instead of RootChain.*

*just curious*, so is the root chain no longer needed in the framework?
tldr would be nice, note i am not a dev~

bg1156:
*just curious*, so is the root chain no longer needed in the framework?
tldr would be nice, note i am not a dev~

pik694:
We are in progress of refactoring the old contract which was called `RootChain`. 
Instead of having one enormous contract (and file),  we divide it into maintainable, readable pieces, which altogether are called `plasma framework`. 

When I wrote :
>  instead of RootChain
I made a mental shortcut and also it is how we internally differentiate those two versions. 

What I should have written was that we will switch python tests from testing the old version of the contract into the new one.

bg1156:
thank you, just trying to keep up

pik694:
🎉

boolafish:
🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳🥳

PR #506, opened 2019-12-05
Python tests - start IFE 

pik694:
The following changes enable starting IFE tests.

Refers #150.

### DIFF
--- a/plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
+++ b/plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
@@ -218,6 +218,7 @@ def test_start_in_flight_exit_invalid_outputs_should_fail(testlang):
         testlang.start_in_flight_exit(spend_id)
 
 
+@pytest.mark.skip("TODO: remove test when such scenarion is covered by a test that processes exits")
 @pytest.mark.parametrize("num_inputs", [1, 2, 3, 4])
 def test_start_in_flight_exit_cancelling_standard_exits_from_inputs(testlang, num_inputs):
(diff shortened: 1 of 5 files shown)


plasma_framework/python_tests/tests/tests_utils/plasma_framework.py
```diff
@@ -202,8 +202,31 @@ def challengeStandardExit(self,
...
+                          in_flight_tx,
+                          input_txs,
+                          input_txs_inclusion_proofs,
+                          in_flight_tx_sigs,
+                          input_utxos_pos,
+                          **kwargs):
+
+        in_flight_tx_witnesses = in_flight_tx_sigs
+
+        # The following arguments are to be used in the future
```
boolafish: We are removing these args actually. I think should be seen while rebase to master.

plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
```diff
@@ -95,19 +92,19 @@ def test_start_in_flight_exit_with_erc20_token_and_eth_should_succeed(testlang,
 
     # Inputs are correctly set
     input_info = in_flight_exit.get_input(0)
-    assert input_info.owner == to_canonical_address(owner.address)
-    assert input_info.token == NULL_ADDRESS
+    assert input_info.exit_target == owner.address
+    assert input_info.token == NULL_ADDRESS_HEX
```
boolafish: Might consider to have a const `ETH` or `ETH_TOKEN` with the value of `NULL_ADDRESS_HEX` to make this more readable.

PR #508, opened 2019-12-05
Migrate mock 712

InoMurko:


boolafish:
hmm...does this work for you?

My personal understanding is that migration script should be done under here: https://github.com/omisego/plasma-contracts/tree/master/plasma_framework/migrations

Not sure if truffle is able to have multiple migration target......

InoMurko:
yeah, no it doesn't. I was just testing :D

PR #515, opened 2019-12-10
refactor: delete TxPosLib

### LINKED ISSUE #254: `TxPosLib` library HAS TO be removed !!!

### LINKED ISSUE #465: Recommendation: Merge `TxPosLib` into `UtxoPosLib` and implement a `decode` function with range checks
## Issue Type 

```
[ ] bug report
[X] feature request
```

#### Description

`TxPosLib` and `UtxoPosLib` serve very similar functions. They both provide utility functions to access the block number and tx index of a packed utxo position variable. `UtxoPosLib`, additionally, provides a function to retrieve the output index of a packed utxo position variable.

What they both lack, though, is sanity checks on the values packed inside a utxo position variable. By implementing a function `UtxoPosLib.decode(uint _utxoPos) returns (UtxoPos)`, each exit controller contract can ensure that the values it is using make logical sense. The `decode` function should check that:

* `txIndex` is between 0 and the maximum number of transactions per block
* `outputIndex` is between 0 and 3

Currently, neither of these restrictions is explicitly enforced. As for `blockNum`, the best check is that it exists in the `PlasmaFramework` contract with a nonzero root. Since `UtxoPosLib` is a pure library, that check is better performed elsewhere (See #465).

Once implemented, all contracts should avoid casting values directly to the `UtxoPos` struct, in favor of using the `decode` function. Merging the two files will help with this.

pgebal:
Closes #254 
Closes #465

Changes:
 * removes `TxPosLib`
 * changes API of `ISpendingCondition`, single `utxoPos` argument is provided instead of `outputIndex` and `txPos`
 * changes definition of tx position, now it's utxo position of zero index output (like it was in prev. version of contracts)
* changes `UtxoPosLib` API and renames it to `PosLib`


boolafish:
I propose that rename the `UtxoPosLib` to `PositionLib` or `PosLib`. Have a `struct` instead of `uint256` holding position value.
```
struct Position {
    uint256 blockNum,
    uint256 txIndex,
    uint16 outputIndex,
}

(PS. probably blockNum and txIndex can be smaller?)
```

and have the following functions:
```
encode(Position pos) returns uint256; // equivalent with current `utxoPos.value`
encodePackedTxPos(Position pos) returns uin256; // equivalent with current `txPos.value`
decode(uint256 encodedPosition) returns (Position); // using the formula of utxoPos
```

I don't think we need `decodeFromPackedTxPos` so I did not put it there.

There is a `build` function in UtxoPosLib now, probably we can remove that function from lib and just write it on the spending condition that needs that.

pgebal:
Good idea @boolafish . I'll refactor it your way.

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
```diff
@@ -3,7 +3,7 @@ pragma experimental ABIEncoderV2;
 
 import "../interfaces/IStateTransitionVerifier.sol";
 import "../payment/PaymentExitDataModel.sol";
-import "../../utils/UtxoPosLib.sol";
+import "../../utils/PosLib.sol";
```
boolafish: probably an unused import

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -0,0 +1,78 @@
...
+    function encode(Position memory pos) internal pure returns (uint256) {
+        return pos.blockNum * BLOCK_OFFSET + pos.txIndex * TX_OFFSET + pos.outputIndex;
+    }
+
+    /**
+     * @notice Encodes a transaction position
+     * @param pos Transaction position
+     * @return Position encoded as an integer
+     */
+    function encodePackedTxPos(Position memory pos) internal pure returns (uint256) {
```
boolafish: ah...I was actually proposing using this function `encodePackedTxPos` to replace the `getTxPostionForExitPriority` function, so we don't feel like providing function so specifically to exit priority. As it would better be the lib for exit priority to know how it want the encoded bytes IMO.

Also, I was thinking to use `Position txPos` for all internal communications such as `spending condition`, `MoreVpFinalization` so that we don't ever need to specifically encode to a txPos with this schema... however, do realize that there is a gap from our original data is `utxoPos` and need to drop the `outputIndex` field.

Personally I would probably have a helper function to drop the `outputIndex` out, probably sth like:
 `function toStrictTxPos(PosLib.Position utxoPos) returns PosLib.Position`.

I will leave this to you. However, if we are keeping current style, I think this function can be named `encodeTxPos` directly?

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -0,0 +1,78 @@
...
+        return encode(pos) / TX_OFFSET;
+    }
+
+    /**
+     * @notice Encodes a position
+     * @param pos Position
+     * @return Position encoded as an integer
+     */
+    function encode(Position memory pos) internal pure returns (uint256) {
+        return pos.blockNum * BLOCK_OFFSET + pos.txIndex * TX_OFFSET + pos.outputIndex;
```
boolafish: auditors might want a range check on this function. so make sure the blockNum, txIndex, outputIndex does not exceed the max amount of it's possibility. Same applies to the `encodePackedTxPos` if we are keeping the function.

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -0,0 +1,78 @@
...
+    uint256 constant internal BLOCK_OFFSET = 1000000000;
+    uint256 constant internal TX_OFFSET = 10000;
+
+    /**
+     * @notice Returns the UTXO position for a given transaction position and output index
+     * @param txPos Transaction position - utxo position of zero index output
+     * @param outputIndex Index of the output
+     * @return utxo position
+     */
+    function buildPositionFromTxPosAndOutputIndex(uint256 txPos, uint16 outputIndex)
```
boolafish: I am actually thinking to change the interface of spending condition a little bit. Instead of 
```
function verify(
        bytes calldata inputTxBytes,
        uint16 outputIndex,
        uint256 inputTxPos,
        bytes calldata spendingTxBytes,
        uint16 inputIndex,
        bytes calldata signature
    )
```

merges the `outputIndex` and `inputTxPos` to `utxoPos` or `outputPos`.
so sth like:
```
function verify(
        bytes calldata inputTxBytes,
        PosLib.Position calldata utxoPos,
        bytes calldata spendingTxBytes,
        uint16 inputIndex,
        bytes calldata signature
    )
```

It was originally separated to make it clearer that the `txPos` might potentially be 0 if we are doing MoreVP 2.0, so where you can chain in-flight txs. However, I think that is far away and it is still possible to support that by giving the `utxoPos` with outputIndex only.

If doing so, it removes the need of this `buildPositionFromTxPosAndOutputIndex` function. Also, it removes one use case of `encodePackedTxPos` function.

plasma_framework/contracts/src/utils/PosLib.sol
pgebal: done

plasma_framework/contracts/src/utils/PosLib.sol
pgebal: done

plasma_framework/contracts/src/exits/payment/PaymentTransactionStateTransitionVerifier.sol
pgebal: done

plasma_framework/contracts/src/utils/PosLib.sol
pgebal: Done, but because of unspecified compilation error (compilation freeze) I couldn't use `PosLib.Position` struct in the interface API, I needed to use `uint256` for `utxoPos`.

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -0,0 +1,67 @@
...
+        return encode(pos) / TX_OFFSET;
+    }
+
+    /**
+     * @notice Encodes a position
+     * @param pos Position
+     * @return Position encoded as an integer
+     */
+    function encode(Position memory pos) internal pure returns (uint256) {
+        return pos.blockNum.mul(BLOCK_OFFSET).add(pos.txIndex.mul(TX_OFFSET)).add(pos.outputIndex);
```
boolafish: using SafeMath does mitigate some issue, at least the one that blockNum * BLOCK_OFFSET would exceed max range of uint256.

But there are still other concerns, eg. `outputIndex < TX_OFFSET`, `txIndex < BLOCK_OFFSET / TX_OFFSET`.

plasma_framework/contracts/src/exits/utils/MoreVpFinalization.sol
```diff
@@ -21,26 +21,27 @@ library MoreVpFinalization {
...
         bytes memory txBytes,
-        TxPosLib.TxPos memory txPos,
+        PosLib.Position memory txPos,
         bytes memory inclusionProof
     )
         internal
         view
         returns (bool)
     {
+        require(txPos.outputIndex == 0, "Invalid transaction position");
```
thec00n: Why do require outputIndex to be zero?

boolafish: 🤔and no test on CI failed for this?

pgebal: @thec00n because it previously was a TxPos. I left it as it was, we still require it to be a tx pos (utxo pos of 0 index output)

Issue #466, opened 2019-11-19
Recommendation: remove `IsDeposit` and add a similar getter to `BlockController`


wadeAlexC:
## Issue Type 

```
[ ] bug report
[X] feature request
```

#### Description

The `IsDeposit` library is used to check whether a block number is a deposit or not. The logic is simple - if `blockNum % childBlockInterval` is nonzero, the block number is a deposit.

By including this check in `BlockController` instead, the contract can perform an existence check as well. The function in `BlockController` would return the same result as the `IsDeposit` library, but would additionally revert if the block in question does not exist:

```
function isDeposit(uint _blockNum) public view returns (bool) {
  require(blocks[_blockNum].timestamp != 0, "Block does not exist");
  return _blockNum % childBlockInterval != 0;
}
```

Note that this check is made at the cost of an external call. If the check needs to be made multiple times in a transaction, the result should be cached.

PR #516, opened 2019-12-11
refactor: move isDeposit into BlockController

kevsul:
I think this refactor simplifies the code as well as adding the extra check that the block exists: it removes the isDeposit library, reduces the size of the `Controller` structs and makes the code a bit easier to read.

I don't feel super strongly about it though - I'm happy to leave it as it is if we think the current way is better.


kevsul:
Closes #466

Issue #509, opened 2019-12-06
migrate Mock contracts with devel flag


InoMurko:
## Issue Type 

```
[ ] bug report
[ x] feature request
```

## Current Behavior
Currently `npx truffle migrate` only deploys essential (root chain) contracts.

## Expected Behavior
Because contracts are also used in development, we would like to have access to contracts in
`./plasma_framework/contracts/mocks/`, specifically `plasma-contracts/plasma_framework/contracts/mocks/transactions/eip712Libs/PaymentEip712LibMock.sol`.

## Suggested Fix
`npx truffle migrate --network remote command` should accept a `--devel` flag that deploys mock contract above as well, and expose them in outputs.json so that they're able to be served ... 

## Motivation for Change
Support development on elixir-omg side (this allows us to delete some of the helper code that used to do the deployment)

InoMurko:
The currently identified contracts we would need are `PaymentEip712LibMock` and `ERC20Mintable` (openzeppelin_solidity)

pdobacz:
slight suggestion to avoid using the word "mock" in this context (and rename the contracts that fall into this category). Would `proxy` work for the `PaymentEip712Lib...` thing? I think currently this is the only contract that does, but not sure.

InoMurko:
I don't have an opinion on this. As long as it's evident it's not what we're shipping to production...

As per discussion on Slack,  `git SHA1 of length 7` should be stored in the contracts so relevant parties are able to retrieve that (similarly to how we retrieve `MIN_EXIT_PERIOD`)

PR #526, opened 2019-12-17
feature: deploy mocks if MOCKS env var

InoMurko:
This tackles https://github.com/omisego/plasma-contracts/issues/509

Deploys two contracts that elixir-omg uses in tests and allows us to delete massive amounts of complex code... so hooraaay?!

Also, if this is tested somehow and you think I should... lemme know!


InoMurko:
To test this locally:

```
EX_DEV=true npx truffle migrate --network remote command
9_output_results.js
===================
{"authority_address":"0xc0f780dfc35075979b0def588d999225b7ecc56f","eth_vault":"0x0433420dee34412b5bf1e29fbf988ad037cc5db7","erc20_vault":"0x04badc20426bc146453c5b879417b25029fa6c73","payment_exit_game":"0x92ce4d7773c57d96210c46a07b89acf725057f21","plasma_framework_tx_hash":"0x375b3b5d36b504b8ccf25d89d0938fef2021f928ab2700df37b7909f242656e9","plasma_framework":"0xc673e4ffcb8464faff908a6804fe0e635af0ea2f","paymentEip712LibMock":"0x19925cc645720fbb61f76304ee15501e3197f3a9","erc20Mintable":"0x32063dba91cf95eb3d58fad9e391ee888878b61c"}
```

```
npx truffle migrate --network remote command
9_output_results.js
===================
{"authority_address":"0xc0f780dfc35075979b0def588d999225b7ecc56f","eth_vault":"0x0433420dee34412b5bf1e29fbf988ad037cc5db7","erc20_vault":"0x04badc20426bc146453c5b879417b25029fa6c73","payment_exit_game":"0x92ce4d7773c57d96210c46a07b89acf725057f21","plasma_framework_tx_hash":"0x375b3b5d36b504b8ccf25d89d0938fef2021f928ab2700df37b7909f242656e9","plasma_framework":"0xc673e4ffcb8464faff908a6804fe0e635af0ea2f"}
```

plasma_framework/Makefile
```diff
@@ -1,6 +1,10 @@
 PYTHON_TESTS_DIR = python_tests/
 RUN_PYTHON_MAKEFILE_COMMAND = $(MAKE) -C $(PYTHON_TESTS_DIR)  
 
+.PHONY: list
```
InoMurko: switch between projects I forget which receipts I have available so this little 
`make list` helps me so that I don't have to open Makefile... If you want it I can leave it otherwise I can delete it too

InoMurko:
I've heard `EX_DEV` is not cool. Any recommendations?

boolafish:
Curious, do we always wants to do this in non prod env?

There is a DEPLOYMENT_ENV already used, might be able to leverage that env

InoMurko:
> Curious, do we always wants to do this in non prod env?
> 
> There is a DEPLOYMENT_ENV already used, might be able to leverage that env

Hm... not really I think. We use `ERC20Mintable` from openzeppelin so that we're able to test anything related to ERC20.

and `PaymentEip712LibMock` for signature tests. For now that's where they live but we're trying hard to get rid of tests like that.

InoMurko:
I see that `DEPLOYMENT_ENV` is specifically used as a string to identify an environment.

boolafish:
Yes, it’s a string. Well...was more recommending that if we always wants the mock contracts on develop env.

We can still have another string for your need but then it might not be more simple then having another flag env.

Meanwhile, I remember @author showed case a way to mount the migration script with docker on demand on elixir side. Just wants to make sure do we still prefer to have support from this repo instead of the setup Author shown.

InoMurko:
What @achiurizo showed is cool but I don't think I can point to two migration directories and deploy both plasma contracts and later deploy `ERC20Mintable` and `PaymentEip712LibMock`.
I don't have enough experience with truffle to see how this could work.

achiurizo:
> Meanwhile, I remember @achiurizo showed case a way to mount the migration script with docker on demand on elixir side. Just wants to make sure do we still prefer to have support from this repo instead of the setup Author shown.

The use case I had was specifically so I didn't need to make changes here and move quickly(but yes, it was a neat hack :D)

> What @achiurizo showed is cool but I don't think I can point to two migration directories and deploy both plasma contracts and later deploy ERC20Mintable and PaymentEip712LibMock.
I don't have enough experience with truffle to see how this could work.

There's a way I mentioned it to @boolafish at some point but it kind of forces you to spin up "one-off" docker service instances just to load the set of contracts you want into the geth/ganache instance(not ideal if we continue to scale out more ephemeral  contracts like eip712libmock).

I think having this upstream(this repo) here to manage that makes the most sense and most available for future projects dependent on these contracts to load artifacts like the mocks.

plasma_framework/Makefile
kevsul: Nice, leave it in 😄

PR #528, opened 2019-12-18
150 challenge ife input output

pik694:
This batch of tests is focused on challenging inputs and outputs of IFEs.


pik694:
Refers #150

PR #537, opened 2019-12-30
fix: do not fail on processing exits from in-flight transaction that has multiple distinct output tokens

### LINKED ISSUE #536: Processing exits from in-flight transaction fails when the transaction has ETH and ERC20 token outputs
## Issue Type 

```
[x ] bug report
[ ] feature request
```

## Current Behavior
Processing exits from in-flight transaction fails when the transaction has ETH and ERC20 token outputs

## Expected Behavior
Exits are successfully processed.

## Steps to Reproduce
1. Start in-flight exit from a transaction with 2 outputs: ETH and ERC20.
2. Piggyback both.
3. Call processExits twice. The second call fails.

## Suggested Fix
Change how exit game controller handles exit processor for a given exit in `enqueue` and `processExits` functions.

pgebal:
Closes #536 

Changes:
 * fix failing on processing exits from in-flight transaction that has multiple distinct output tokens
 * enable python tests for processing exits

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -14,8 +14,8 @@ import "../utils/PosLib.sol";
  *         For details, see the Plasma MVP spec: https://ethresear.ch/t/minimal-viable-plasma/426
  */
 contract ExitGameController is ExitGameRegistry {
-    // exit priority => IExitProcessor
-    mapping (uint256 => IExitProcessor) public delegations;
+    // exit hashed (priority, token) => Delegation
+    mapping (bytes32 => IExitProcessor) public delegations;
     // hashed (vault id, token) => PriorityQueue
     mapping (bytes32 => PriorityQueue) public exitsQueues;
     // outputId => bool
@@ -131,7 +131,9 @@ contract ExitGameController is ExitGameRegistry {
         uint256 priority = ExitPriority.computePriority(exitableAt, txPos, exitId);
 
         queue.insert(priority);
-        delegations[priority] = exitProcessor;
+
+        bytes32 delegationKey = getDelegationKey(priority, token);
+        delegations[delegationKey] = exitProcessor;
 
         emit ExitQueued(exitId, priority);
         return priority;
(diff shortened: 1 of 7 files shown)


boolafish:
1. Might consider to use `hash(priority, exitQueueKey)` instead.
1. BTW, do we promise that we would not enqueue with same priority twice to same queue? hmm....

kevsul:
> 1. Might consider to use `hash(priority, exitQueueKey)` instead.

Agreed, because there might be 2 queues of the same token.

> 2. BTW, do we promise that we would not enqueue with same priority twice to same queue? hmm....

Not sure. All the outputs of an IFE appear to have the same priority?

pgebal:
> > 1. BTW, do we promise that we would not enqueue with same priority twice to same queue? hmm....
> 
> Not sure. All the outputs of an IFE appear to have the same priority?

We have a check that validates if we enqueue once per (exitId, token) pair.

pgebal:
> > 1. Might consider to use `hash(priority, exitQueueKey)` instead.
> 
> Agreed, because there might be 2 queues of the same token.
> 

Right. I'll fix that.
> > 1. BTW, do we promise that we would not enqueue with same priority twice to same queue? hmm....
> 
> Not sure. All the outputs of an IFE appear to have the same priority?
We have a check that validates if we enqueue once per (exitId, token) pair.

boolafish:
> We have a check that validates if we enqueue once per (exitId, token) pair.

Where is this?

I think in the case of SE<>IFE, it would crashes. eg, if you SE the youngest input of IFE, both of them should have same priority for the same queue.

pgebal:
> > We have a check that validates if we enqueue once per (exitId, token) pair.
> 
> Where is this?
> 
> I think in the case of SE<>IFE, it would crashes. eg, if you SE the youngest input of IFE, both of them should have same priority for the same queue.

Sorry it looks you are right about SE <> IFE.

pgebal:
> > We have a check that validates if we enqueue once per (exitId, token) pair.
> 
> Where is this?
> 
> I think in the case of SE<>IFE, it would crashes. eg, if you SE the youngest input of IFE, both of them should have same priority for the same queue.

@boolafish Shouldn't they have different exitIds? I'll check that, but looking at how we compute exit ids for se and ife I can't see straight away that they have the same ids.

boolafish:
Ah you’re right, they have different exitId

PR #538, opened 2019-12-30
tests: enable tests for start standard exit and start in-flight exit interactions

pgebal:
References #150
Changes:
 * remove test that checks if it's impossible to start a standard exit after in-flight exit from the same output was started earlier , as it is not longer an expected behavior
 * add test that checks if finalizing standard exit blocks finalizing of the same output exited via in-flight exit

### DIFF
--- a/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
+++ b/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
@@ -704,19 +704,25 @@ def test_in_flight_exit_is_cleaned_up_even_though_none_of_outputs_exited(testlan
     assert testlang.get_balance(owner) == pre_balance + testlang.root_chain.inFlightExitBond() * 2
 
 
-def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework):
-    # exit cross-spend test, case 7
+@pytest.mark.parametrize("num_outputs", [1, 2, 3, 4])
+def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework, num_outputs):
     owner_1, amount, amount_spent = testlang.accounts[0], 100, 1
     deposit_id = testlang.deposit(owner_1, amount)
-    spend_id = testlang.spend_utxo([deposit_id], [owner_1], [(owner_1.address, NULL_ADDRESS, amount_spent)])
+
+    outputs = []
+    for i in range(0, num_outputs):
+        outputs.append((owner_1.address, NULL_ADDRESS, amount_spent))
+    spend_id = testlang.spend_utxo([deposit_id], [owner_1], outputs)
+
+    output_index = num_outputs - 1
 
     testlang.start_in_flight_exit(spend_id)
 
     blknum, txindex, _ = decode_utxo_id(spend_id)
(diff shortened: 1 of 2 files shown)


pgebal:
@boolafish please take a look if it's ok to remove those two tests

boolafish:
agree that the two tests are no longer valid

plasma_framework/python_tests/tests/contracts/root_chain/test_start_standard_exit.py
```diff
@@ -130,60 +130,6 @@ def test_start_standard_exit_from_deposit_must_be_exitable_in_minimal_finalizati
     assert testlang.get_standard_exit(deposit_id) == [NULL_ADDRESS_HEX, 0, 0, False]
 
 
-@pytest.mark.skip("Skipped due to mix with IFE")
-@pytest.mark.parametrize("num_outputs", [1, 2, 3, 4])
-def test_start_standard_exit_on_piggyback_in_flight_exit_valid_output_owner_should_fail(testlang, num_outputs):
```
pik694: I would rather move this test than remove it.
I think we should have covered all interaction scenarios from an old doc: [https://github.com/omisego/elixir-omg/blob/v0.2/docs/standard_vs_in_flight_exits_interaction.md](https://github.com/omisego/elixir-omg/blob/v0.2/docs/standard_vs_in_flight_exits_interaction.md). 

Even if we have moved to ALD, the cross-spend scenarios still exist, even though they may now be handled differently - they are protocol-level.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -702,3 +702,26 @@ def test_in_flight_exit_is_cleaned_up_even_though_none_of_outputs_exited(testlan
 
     # assert IFE and piggyback bonds were sent to the owners
     assert testlang.get_balance(owner) == pre_balance + testlang.root_chain.inFlightExitBond() * 2
+
+
+def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework):
```
pik694: could we also parametrise the new test, as we did with the previous one

plasma_framework/python_tests/tests/contracts/root_chain/test_start_standard_exit.py
pgebal: This is no longer a valid scenario. The one thing we should test here is that finalizing standard exit from the same output as ife exit does not exit the output twice. It's covered in the test added to process exits tests

pik694: I think that the scenario 9 per se is still valid - `Mallory submits IFE -> Mallory piggybacks output -> Mallory does SE on output`. It should be now tested in some other way.

boolafish: feel like case 7 and 9 become same cases as there is no "blocking piggyback" part anymore?

pgebal: @pik694 scenario is valid except for the result, now it's ok to start an exit. It just won't get paid as ife is processed earlier.

PR #540, opened 2019-12-31
Split truffle migrations

### LINKED ISSUE #532: Split Migration 5 into smaller tasks
## Issue Type 

```
[ ] bug report
[X] feature request
```

## Current Behavior
Migration 5 contains a number of operations which Truffle does not handle well (see https://github.com/trufflesuite/truffle/issues/2257).

## Expected Behavior
Contracts deploy

## Steps to Reproduce
Deploy contracts to mainnet

## Suggested Fix
During a recent mainnet deployment we were able to proceed by splitting this migration into smaller steps which is a workaround for the faulty deployment tooling.

## Motivation for Change
:D

kevsul:
Closes #532

Workaround for https://github.com/trufflesuite/truffle/issues/2257

The main idea is to have less steps per migration script, so that if Truffle does fail for some reason, then the next run won't need to repeat as many deployments.

Also, renamed the files Basic programming style 🤘 so that it's easier to add in more steps in the future


boolafish:
I think I am not fully following the new file name style. Do you mind to give a description of what is your plan to add new files and what should be the convention?

jbunce:
I'm with @boolafish here. What's the standard? :)

kevsul:
The intention is to avoid having to rename files again when adding new contracts to the deployment.
The idea was:
- 1-98: Initial setup and main contracts like PlasmaFramework, SpendingConditionRegistry, etc.
- 99: Renounce the SpendingConditionRegistry because this has to happen before any ExitGames
- 100: First ExitGame contracts
- 200: Second ExitGame contracts
- ...
- 1000-: Cleanup, dev/test contracts and output

boolafish:
so 1~1000 is reserved for our first deployment. 
In the future when we develop new exit game (eg. DEX), we start to use 1000~2000 range?

kevsul:
> so 1-1000 is reserved for our first deployment.
> In the future when we develop new exit game (eg. DEX), we start to use 1000-2000 range?

No, DEX exit games would start at 300_
Next exit game at 400_, etc

boolafish:
hmm, I don't it can work like that as truffle migrate would start from the last index which would be 1000 in the case unless you're deploy with `reset` flag. However, we don't want to that as (at least in prod) we'd only want new exit games to be deployed and registered.

kevsul:
> > so 1-1000 is reserved for our first deployment.
> > In the future when we develop new exit game (eg. DEX), we start to use 1000-2000 range?

Oh, you mean after this is all deployed?
Yeah, I would start at 2000_

PR #541, opened 2019-12-31
tests: refactor/delete disabled python tests for starting ife

### LINKED ISSUE #150: Run python tests on plasma framework

pgebal:
Closes #150 
Changes:
 * refactor test for starting ife twice, move it to process exits suite
 * remove tests not aligned with ALD design: starting an ife no longer challenges standard exits and assumes that ife is canonical


plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
```diff
@@ -218,70 +200,6 @@ def test_start_in_flight_exit_invalid_outputs_should_fail(testlang):
         testlang.start_in_flight_exit(spend_id)
 
 
-@pytest.mark.skip("TODO: remove test when such scenarion is covered by a test that processes exits")
-@pytest.mark.parametrize("num_inputs", [1, 2, 3, 4])
-def test_start_in_flight_exit_cancelling_standard_exits_from_inputs(testlang, num_inputs):
```
pik694: before deleting this test, please make sure that the interaction of parallel SE and IFE is tested in processing tests. 
Now do not have such a test since the previous contract made the interaction checks at starting/piggybacking exits

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -731,3 +731,26 @@ def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework
...
+    start_piggyback_process(spend_id, owner, testlang)
+
+    # Second time should succeed but should not withdraw funds from the vault
+    pre_exit_balance = testlang.get_balance(plasma_framework.eth_vault)
+    start_piggyback_process(spend_id, owner, testlang)
+    post_exit_balance = testlang.get_balance(plasma_framework.eth_vault)
+    assert post_exit_balance == pre_exit_balance
+
+
+def start_piggyback_process(spend_id, owner, testlang):
```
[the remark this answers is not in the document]
pgebal: right, misleading name

plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
pgebal: Test for that interaction is named `test_output_exited_via_se_and_ife_exits_only_once`, it was merged earlier. I uploaded a merge commit, so please check it. I think this test is enough.

pik694: I would like to have two cases: SE from IFE input, SE from IFE output, and in each case check how it behaves when IFE exits first and then when SE exits first.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -717,6 +717,67 @@ def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework
...
+
+    spend_id = testlang.spend_utxo([deposit_id], [owner], outputs=[(owner.address, NULL_ADDRESS, amount_spent)])
+
+    testlang.start_standard_exit(deposit_id, account=owner)
+
+    start_ife_piggyback_and_process(spend_id, owner, testlang)
+
+    post_exit_balance = testlang.get_balance(plasma_framework.eth_vault)
+    # deposit exit not challenged and withdrawn via se - broken Plasma
+    assert post_exit_balance == 0
```
pik694: I would rather try to ensure that the money wasn't transferred to the exit owner.

Maybe, also, let's check events that the certain exit was omitted?

plasma_framework/python_tests/tests/contracts/root_chain/test_start_in_flight_exit.py
pgebal: in-flight transaction input, standard exit interaction is tricky, if SE is not challenged -> plasma broken, so that's why named tests with `does_not_break_exit_processing` suffix

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
pgebal: I wouldn't go that far, funds are stealed anyway

pik694: What I mean is that if the balance of sth vault is 0 before the processing, you will never get balance below zero (at least in eth, this may not be true to ERC20 tokens 🤔 ).

Lets make non-zero eth vault balance, before processing and then check whether sth has changed. 

And I still am for checking the event.

pik694: right, but after line 755 it is again 0

pik694: sure, but when processing exits we have two exits in the queue - we process the first one which transfers all the money from the contract, and then when processing the second exit the SuT may try to transfer money for the second time, but after such a transfer you will still end up with 0 amount - kind of a false positive test.

pgebal: As plasma framework does not guarantee what happens with funds after invalid exits are not challenged (se  <> ife input, both processed) I think we should not test those scenarios.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -715,3 +716,88 @@ def test_output_exited_via_se_and_ife_exits_only_once(testlang, plasma_framework
...
+    start_ife_piggyback_and_process(spend_id, owner, testlang)
+
+    # Second time should succeed but should not withdraw funds from the vault
+    pre_exit_balance = testlang.get_balance(plasma_framework.eth_vault)
+    start_ife_piggyback_and_process(spend_id, owner, testlang)
+    post_exit_balance = testlang.get_balance(plasma_framework.eth_vault)
+    assert post_exit_balance == pre_exit_balance
+
+
+def start_ife_piggyback_and_process(spend_id, owner, testlang):
```
pik694: Please move this helper either to the end of the file or to the beginning.

Issue #546, opened 2020-01-05
Merkle.checkMembership allows existence proofs for the same leaf in multiple locations in the tree


wadeAlexC:
## Issue Type 

```
[X] bug report
[ ] feature request
```

#### Description

`checkMembership` is used by several contracts to prove that transactions exist in the child chain. The function uses a `leaf`, an `index`, and a `proof` to construct a hypothetical root hash. This constructed hash is compared to the passed in `rootHash` parameter. If the two are equivalent, the proof is considered valid.

The proof is performed iteratively, and uses a pseudo-index (`j`) to determine whether the next proof element represents a "left branch" or "right branch":

https://github.com/omisego/plasma-contracts/blob/eed3a6ec12aa2ab6676b3e775136cdbf143910f6/plasma_framework/contracts/src/utils/Merkle.sol#L28-L41

If `j` is even, the computed hash is placed before the next proof element. If `j` is odd, the computed hash is placed after the next proof element. After each iteration, j is decremented by `j = j / 2`.

Because `checkMembership` makes no requirements on the height of the tree or the size of the proof relative to the provided `index`, it is possible to pass in invalid values for `index` that prove a leaf's existence in multiple locations in the tree.

#### Examples

By modifying existing tests, we showed that for a tree with 3 leaves, leaf 2 can be proven to exist at indices 2, 6, and 10 using the same proof each time. The modified test can be found here: https://gist.github.com/wadeAlexC/01b60099282a026f8dc1ac85d83489fd#file-merkle-test-js-L40-L67

```
it('should accidentally allow different indices to use the same proof', async () => {
  const rootHash = this.merkleTree.root;
  const proof = this.merkleTree.getInclusionProof(leaves[2]);

  const result = await this.merkleContract.checkMembership(
    leaves[2],
    2,
    rootHash,
    proof,
  );
  expect(result).to.be.true;

  const nextResult = await this.merkleContract.checkMembership(
    leaves[2],
    6,
    rootHash,
    proof,
  );
  expect(nextResult).to.be.true;

  const nextNextResult = await this.merkleContract.checkMembership(
    leaves[2],
    10,
    rootHash,
    proof,
  );
  expect(nextNextResult).to.be.true;
});
```

#### Recommendation

Use the length of the proof to determine the maximum allowed index. The passed-in index should satisfy the following criterion: `index < 2**(proof.length/32)`

paulperegud:
That's an amazing find!

PR #547, opened 2020-01-06
fix: mekle contract checks the index size

boolafish:
### Note
Merkle contract did not check whether index exceeds the max amount with the proof size. This can lead to issue where given different indexes that exceeds the size, it can still pass the check due to the implementation of for loop.

closes: #546

### DIFF
--- a/plasma_framework/contracts/src/utils/Merkle.sol
+++ b/plasma_framework/contracts/src/utils/Merkle.sol
@@ -25,7 +25,7 @@ library Merkle {
         require(proof.length % 32 == 0, "Length of Merkle proof must be a multiple of 32");
 
         // see https://github.com/omisego/plasma-contracts/issues/546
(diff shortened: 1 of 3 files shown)


plasma_framework/contracts/src/utils/Merkle.sol
```diff
@@ -24,6 +24,9 @@ library Merkle {
         require(proof.length != 0, "Merkle proof must not be empty");
         require(proof.length % 32 == 0, "Length of Merkle proof must be a multiple of 32");
 
+        // see https://github.com/omisego/plasma-contracts/issues/546
+        require(index < 2**(proof.length/32), "Index exceeds the possible size that matches with the proof");
```
pgebal: Maybe just "Invalid index" or "Index does not match the length of the proof"

PR #550, opened 2020-01-07
docs: add section elaborating on signed transactions for chch/watcher

pdobacz:
Following a request from @achiurizo, this adds a section elaborating handling of signed transactions in https://github.com/omisego/elixir-omg/ to the intregration docs transaction section.

I tried to follow conventions and blend in with the rest of the doc, but not sure if I got this right. Same goes for level of detail.

Also, there's an alternative location for this documentation: it could be put wherever the handling of such encoded signed transactions is done (`elixir-omg` currently). So if this doesn't fit in here, let's discuss.

### DIFF
--- a/plasma_framework/docs/integration-docs/integration-doc.md
+++ b/plasma_framework/docs/integration-docs/integration-doc.md
@@ -97,6 +97,30 @@ The current implementation supports only the `Payment` transaction type.
 
 Support for additional transaction types, such as ERC721, is reserved for future development.
 
+### Signed transaction format (child chain and watcher only)
+A note on encoding of signed transactions (transactions including witnesses) when handled in the [child chain and watcher implementations](github.com/omisego/elixir-omg).
+
+Signed transactions are:
+
+```
+signedTransaction ::= [signature] rawTransaction
+```
+
+Where
+```
+signatures ::= bytes
+rawTransaction ::= transaction
+```
+
+and are transferred and stored RLP-encoded.
+
+Child chain and watcher both expect, for a signed transaction to be valid, that:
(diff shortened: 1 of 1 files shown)


plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -90,13 +90,36 @@ outputType ::= uint256
...
+Where
+```
+signatures ::= bytes
+rawTransaction ::= transaction
+```
+
+and are transferred and stored RLP-encoded.
+
+Child chain and watcher both expect, for a signed transaction to be valid, that:
+ - every input corresponds to a signature by the respective output's owner (`outputGuard`)
```
kevsul: I think this makes more sense (but feel free to keep the original)
```suggestion
 - every input has a corresponding signature by the respective output's owner (`outputGuard`)
```

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -90,13 +90,36 @@ outputType ::= uint256
...
+
+Signed transactions are:
+
+```
+signedTransaction ::= [signature] rawTransaction
+```
+
+Where
+```
+signatures ::= bytes
```
```suggestion
signature ::= bytes
```

pdobacz:
hm, why would the Python tests fail here?

pgebal:
> hm, why would the Python tests fail here?

I have the same issue, these are gas cost tests. I don't know how to solve that yet.

boolafish:
> hm, why would the Python tests fail here?

Sigh, the latest release of `ganache-cli v6.8.1` breaks this: https://github.com/trufflesuite/ganache-cli/releases/tag/v6.8.1

PR #551, opened 2020-01-07
docs: add fees section to integration docs

### LINKED ISSUE #535: Integration docs: transaction fees

pgebal:
Closes #535

### DIFF
--- a/plasma_framework/docs/integration-docs/integration-doc.md
+++ b/plasma_framework/docs/integration-docs/integration-doc.md
@@ -165,7 +165,9 @@ Note that:
 12. `metadata` must be padded to 32 bytes long
 
 ### Payment transaction fees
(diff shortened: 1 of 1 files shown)


plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -164,6 +164,39 @@ Note that:
 11. `txData` is unused and must be set to `0`
 12. `metadata` must be padded to 32 bytes long
 
+### Payment transaction fees
+Operator collects fees for including a transaction in a block. Transaction fee is implicit. It is the difference between transaction input utxos and output utxos, grouped by the token. For example if transaction inputs are the following utxos:
```
```suggestion
Operator collects fees for including a transaction in a block. The transaction fee is implicit i.e. it is the difference between the transaction inputs and outputs, grouped by the token. For example if the transaction inputs are as follows:
```

pgebal:
> Can you add a section about how only certain tokens can be used to pay fees, and the errors returned if a tx is submitted with insufficient or no fees?
I think it's done. It would be great if we could refer to specification for errors returned by the child chain API but it's not ready yet. That's why I put the error code here in plasma-contracts docs.

PR #533, opened 2019-12-24
add stricter range check for audit feedback

boolafish:
### Note
- Move the check of Merkle proof non empty to `checkMembership` function under `Merkle.sol`
- Add stricter range check in `PosLib`:
   - `txIndex` should be max to 2^16 - 1. Also, use `uint16` for the field.
   - Since there is a limit on ExitPriority that only 54 bits are reserved for `blockNum` + `txIndex`. Set the `MAX_BLOCK_NUM = ((2 ** 54 - 1) - MAX_TX_INDEX) / (BLOCK_OFFSET / TX_OFFSET);` and add check with that range.
- Add tests for edge values

PR #560, opened 2020-01-13
docs: explain that in-flight transaction inputs/outputs are not available to spent on child chain

pgebal:
Closes #533
Changes:
 * rename `ExitGameController.isOutputSpent` to `ExitGameController.isOutputFinalized` to avoid confusion
 * add section in integration docs explaining when in-flight transaction inputs/ouputs are not available to spend on child chain

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -187,7 +187,7 @@ contract ExitGameController is ExitGameRegistry {
      * @notice Checks whether any of the output with the given outputIds is already spent
      * @param _outputIds Output IDs to check
      */
-    function isAnyOutputsFinalized(bytes32[] calldata _outputIds) external view returns (bool) {
+    function isAnyOutputFinalized(bytes32[] calldata _outputIds) external view returns (bool) {
         for (uint i = 0; i < _outputIds.length; i++) {
             if (isOutputFinalized[_outputIds[i]] == true) {
                 return true;
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -19,7 +19,7 @@ contract ExitGameController is ExitGameRegistry {
     // hashed (vault id, token) => PriorityQueue
     mapping (bytes32 => PriorityQueue) public exitsQueues;
     // outputId => bool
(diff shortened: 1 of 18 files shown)


plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -187,9 +187,9 @@ contract ExitGameController is ExitGameRegistry {
      * @notice Checks whether any of the output with the given outputIds is already spent
      * @param _outputIds Output IDs to check
      */
-    function isAnyOutputsSpent(bytes32[] calldata _outputIds) external view returns (bool) {
+    function isAnyOutputsFinalized(bytes32[] calldata _outputIds) external view returns (bool) {
```
kevsul: Strictly speaking this should be `isAnyOutputFinalized`.
But it's not a big deal, I'll leave it up to you...

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -629,6 +629,10 @@ PaymentExitGame.startInFlightExit([
 
 ### Piggybacking on an in-flight exit input / output
 
+To exit input or output of an in-flight transaction user has to piggyback on it.
```
```suggestion
To exit an input or output of an in-flight transaction the user has to piggyback on it.
```

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -629,6 +629,10 @@ PaymentExitGame.startInFlightExit([
...
+
+When in-flight transaction exits, child chain removes the inputs of the in-flight transaction from the spendable set and it no longer can be spent. Calls to child chain to spend the inputs or outputs of the in-flight transaction result in `submit:utxo_not_found` error. The effect of input or output not being available to spend is not visible in contract's state until exit is processed. After the in-flight exit is processed `PlasmaFramework.isOutputFinalized(piggybackedInputOrOutputPosition)` is set to false.
```
```suggestion
When in-flight transaction exits, the child chain removes the inputs of the in-flight transaction from the spendable set and they can no longer be spent. Calls to the child chain to spend the inputs or outputs of an in-flight transaction result in `submit:utxo_not_found` error. The effect of an input or output not being available to spend is not visible in contract's state until the exit is processed. After the in-flight exit is processed `PlasmaFramework.isOutputFinalized(piggybackedInputOrOutputPosition)` is set to true.
```

pgebal: What a mistake I've made with this `is set to false`!

Issue #559, opened 2020-01-13
Add check on enqueue that blocks enqueue with same priority


boolafish:
## Issue Type 

```
[ ] bug report
[x] feature request
```

## Why
Continues with previous issue #536 and its fixing PR #537 , but add explicit check on the `PlasmaFramework` side that we don't override the `delegations` map if already exists. We revert instead. This protects the Plasma Framework from the bug of exit game and block all priority queue.

https://github.com/omisego/plasma-contracts/blob/6f757eae1660bcd880f29346c67d27b9e6f2dfba/plasma_framework/contracts/src/framework/ExitGameController.sol#L136

PR #563, opened 2020-01-15
feat: explicit check on enqueue with same priority

boolafish:
### Note
It is invalid to do so as it would break the framework. Add explicit check instead of rely on
the correct implementation of ExitGame contracts.

closes: https://github.com/omisego/plasma-contracts/issues/559

### DIFF
--- a/plasma_framework/contracts/src/framework/ExitGameController.sol
+++ b/plasma_framework/contracts/src/framework/ExitGameController.sol
@@ -104,6 +104,7 @@ contract ExitGameController is ExitGameRegistry {
      *         priority queue to enforce the priority of exit during 'processExits'
      * @dev emits ExitQueued event, which can be used to back trace the priority inside the queue
      * @dev Caller of this function should add "pragma experimental ABIEncoderV2;" on top of file
+     * @dev Priority (exitableAt, txPos, exitId) must be unique per queue. Be aware to not enqueue with same priority twice in same queue before it is deleted from the queue.
      * @param vaultId Vault ID of the vault that stores exiting funds
      * @param token Token for the exit
      * @param exitableAt The earliest time a specified exit can be processed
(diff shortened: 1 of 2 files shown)


plasma_framework/test/src/framework/ExitGameController.test.js
```diff
@@ -134,7 +134,7 @@ contract('ExitGameController', () => {
             );
         });
 
-        it('can enqueue with the same exitable timestamp (priority) multiple times', async () => {
+        it('rejects when the the exact same priority has already been in the priority queue', async () => {
```
pgebal: delete one 'the'

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -104,6 +104,7 @@ contract ExitGameController is ExitGameRegistry {
      *         priority queue to enforce the priority of exit during 'processExits'
      * @dev emits ExitQueued event, which can be used to back trace the priority inside the queue
      * @dev Caller of this function should add "pragma experimental ABIEncoderV2;" on top of file
+     * @dev Priority (exitableAt, txPos, exitId) must be unique per queue. Be aware to not enqueue with same priority twice in same queue before it is deleted from the queue.
```
pgebal:
```suggestion
     * @dev Priority (exitableAt, txPos, exitId) must be unique per queue. Do not enqueue when the same priority is already in the queue.
```

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -133,6 +134,7 @@ contract ExitGameController is ExitGameRegistry {
         queue.insert(priority);
 
         bytes32 delegationKey = getDelegationKey(priority, vaultId, token);
+        require(address(delegations[delegationKey]) == address(0), "The exact same priority has already been in the priority queue");
```
pgebal:
```suggestion
        require(address(delegations[delegationKey]) == address(0), "The same priority is already enqueued");
```

Issue #327, opened 2019-10-02
Integrate Mythx analysis into CI


thec00n:
## Issue Type 

```
[ ] bug report
[ x ] feature request
```

## Expected Behavior
MythX supports quick mode and full mode. Quick mode should run on every branch. Full mode can take several hours we start with master only and see how it goes. 

## Implementation Steps 
1.) Setup Mythx and truffle-security and optimize test configuration 
2.) Triange security issues 
3.) Set a threshold for new security issues that break a build

PR #564, opened 2020-01-17
MythX setup

thec00n:
Closes https://github.com/omisego/plasma-contracts/issues/327

- Allows to run MythX in quick and full mode 
- Scans cover also Mock contracts that isolate security relevant functionality
- Added one property test for RLP decode/encode 1:1 relationship. Still need to have another pass over it
- It does not make sense to run it in CI just yet as it is not possible to ignore individual false positives


boolafish:
@thec00n I checkout the branch locally and tried to run `npm run mythx-quick`, but it got hanged like hours. How should we run the test?

Issue #568, opened 2020-01-23
service version


InoMurko:
## Issue Type 

```
[ ] bug report
[ x] feature request
```

## Current Behavior
manual git SHA 

## Expected Behavior
a deployment would contain it's own release name

## Suggested Fix
The service design we have should allows us to independently release any of our dependency services, as long as they pass the required testing. For the sake of introspection and traceability, a particular service should propagate it's current deployed version.

I understand that at the time of deployment, we can read the git SHA and release tag from the repository and bake it into the deploy... does ALD follow a truffle migration approach and would that allow us to update a particular public variable that dependency services can read? 

Could you @kevsul and @boolafish give me some hints how to implement this ... if you find it valuable.
The requirement comes from: https://github.com/omisego/elixir-omg/issues/937#issuecomment-577663312
https://github.com/omisego/devops/issues/312#issuecomment-576198237

(tldr dropping SHA from domain name of an environment would simplify infrastructure and datadog metrics, but the information is needed anyway! services that interact with contracts would be able to expose this information by reading it contracts)

InoMurko:
would this approach allow us to update the version from a setter?
for example when we add a new payment type (that doesn't mean we make a full contract redeploy) but just append a contract and register it.

So I guess the requirements are:
 - we follow https://semver.org/
 - set the public var in a constructor
 - provide a getter (so that external parties can read the value)
 - provide a setter so that only the contract operator can bump the version (when it adds a new payment type - ALD purpose)

Would you agree?

boolafish:
sure. if we are having setter anyway probably we don't even need to inject from constructor?

BTW, do we really want git hash instead of some version tag? It seems more consistent with general dependency control by versioning?

InoMurko:
Not having git hash works if what we're deploying is a release, but in a development environment, a master deploy happens more often.

InoMurko:
One thing I've noticed is that our tag 1.0.0 release, did not bump
https://github.com/omisego/plasma-contracts/blob/v1.0.0/plasma_framework/package.json#L3
was this intentional or forgotten or we're not using this?

If we were, I could pull the version from here:
```
plasma_framework  master $ npm version
{
  plasma_framework: '0.0.1',
  npm: '6.11.3',
  ares: '1.15.0',
  brotli: '1.0.7',
  cldr: '35.1',
  http_parser: '2.8.0',
  icu: '64.2',
  llhttp: '1.1.4',
  modules: '72',
  napi: '4',
[...]

PR #569, opened 2020-01-24
feature: track semver in contracts

InoMurko:
First version of https://github.com/omisego/plasma-contracts/issues/568

What was done?
- getters and setters on plasma framework
- setter can only be called by the maintainer
- migration calls the setter and pulls the plasma framework version from `package.json` and adds short git sha (7)

### DIFF
--- a/plasma_framework/migrations/2_deploy_plasma_framework.js
+++ b/plasma_framework/migrations/2_deploy_plasma_framework.js
@@ -1,6 +1,8 @@
 const PlasmaFramework = artifacts.require('PlasmaFramework');
 
+const childProcess = require('child_process');
 const config = require('../config.js');
+const pck = require('../package.json');
 
 module.exports = async (
     deployer,
@@ -19,4 +21,6 @@ module.exports = async (
 
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
+    const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
+    await plasmaFramework.setVersion(''.concat(pck.version, '+', sha), { from: maintainerAddress });
 };
--- a/plasma_framework/migrations/2_deploy_plasma_framework.js
+++ b/plasma_framework/migrations/2_deploy_plasma_framework.js
@@ -22,5 +22,5 @@ module.exports = async (
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
     const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
-    await plasmaFramework.setVersion('${pck.version}+${sha}'), { from: maintainerAddress });
+    await plasmaFramework.setVersion(`${pck.version}+${sha}`), { from: maintainerAddress });
 };
--- a/plasma_framework/migrations/2_deploy_plasma_framework.js
+++ b/plasma_framework/migrations/2_deploy_plasma_framework.js
@@ -22,5 +22,5 @@ module.exports = async (
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
     const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
-    await plasmaFramework.setVersion(`${pck.version}+${sha}`), { from: maintainerAddress });
+    await plasmaFramework.setVersion(`${pck.version}+${sha}`, { from: maintainerAddress });
 };
--- a/plasma_framework/migrations/2_deploy_plasma_framework.js
+++ b/plasma_framework/migrations/2_deploy_plasma_framework.js
@@ -22,5 +22,5 @@ module.exports = async (
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
     const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
-    await plasmaFramework.setVersion(`${pck.version}+${sha}`), { from: maintainerAddress });
+    await plasmaFramework.setVersion('${pck.version}+${sha}'), { from: maintainerAddress });
 };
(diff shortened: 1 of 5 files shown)


InoMurko:
In terms of ALD and adding new Vaults for example... would also need to bump the already deployed version of contracts. Not sure what the process is for that?

plasma_framework/migrations/2_deploy_plasma_framework.js
```diff
@@ -19,4 +21,6 @@ module.exports = async (
 
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
+    const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
+    await plasmaFramework.setVersion(''.concat(pck.version, '+', sha), { from: maintainerAddress });
```
InoMurko: @boolafish how would I check if this transaction succeeded? currently I think there's nothing checking if this actually worked (same for `activateChildChain`).

boolafish: hmmm, if it reverts, it would stop the migration script just like throwing errors without catching. 
If you want to assert probably we can use the node js library here to check with the result from your `getVersion` result.

https://nodejs.org/api/assert.html

InoMurko:
are we cool with this? @kevsul @boolafish @achiurizo ?

plasma_framework/migrations/2_deploy_plasma_framework.js
kevsul: Nitpick: javascript convention that we follow (more or less) is to use string interpolation instead of `.concat`when building strings
https://github.com/airbnb/javascript#es6-template-literals

```suggestion
    await plasmaFramework.setVersion(`${pck.version}+${sha}`), { from: maintainerAddress });
```

plasma_framework/test/src/framework/PlasmaFramework.test.js
```diff
@@ -27,5 +29,19 @@ contract('PlasmaFramework', ([authority, maintainer]) => {
         it('should set maintainer address', async () => {
             expect(await this.framework.getMaintainer()).to.equal(maintainer);
         });
+
+        it('should set semver string', async () => {
+            const sha = childProcess.execSync('git rev-parse HEAD').toString().trim()
+                .substring(0, 7);
+            await this.framework.setVersion(''.concat('1.0.1', '+', sha), { from: maintainer });
```
kevsul: Same picky nitpick as above 😄

paulperegud:
Just a word of warning. Semver is a poorly understood tool which goes against human ideas about versioning. 

Semver is:
* harder than git
* less understood than git

Yes, I'm fun at parties.

pnowosie:
Is this version somehow related to the version from eip-712 domain separator?
Obviously we’d like to match at most only major version, if at all.

https://github.com/omisego/plasma-contracts/blob/a1d16ccddfbfbca8330b91afd1327c82b2227b61/plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol#L42

InoMurko:
> Is this version somehow related to the version from eip-712 domain separator?
> Obviously we’d like to match at most only major version, if at all.
> 
> https://github.com/omisego/plasma-contracts/blob/a1d16ccddfbfbca8330b91afd1327c82b2227b61/plasma_framework/contracts/src/transactions/eip712Libs/PaymentEip712Lib.sol#L42

Interesting question. WDYT @kevsul @boolafish ?

boolafish:
It's a nice idea. However, how do we propose to do the version upgrade when adding new exit games? Now we have a setter that bumps the version to framework, but we probably cannot upgrade the version embedded into EIP712?

plasma_framework/migrations/2_deploy_plasma_framework.js
```diff
@@ -22,5 +22,5 @@ module.exports = async (
     const plasmaFramework = await PlasmaFramework.deployed();
     await plasmaFramework.activateChildChain({ from: authorityAddress });
     const sha = childProcess.execSync('git rev-parse HEAD').toString().trim().substring(0, 7);
-    await plasmaFramework.setVersion(`${pck.version}+${sha}`), { from: maintainerAddress });
+    await plasmaFramework.setVersion('${pck.version}+${sha}'), { from: maintainerAddress });
```
boolafish: lol, this commit is wrong. you need to use ` instead of ' to be able to put variable in js.

InoMurko: lol, I thought Kevin made a typo :D

plasma_framework/test/src/framework/PlasmaFramework.test.js
```diff
@@ -27,5 +29,19 @@ contract('PlasmaFramework', ([authority, maintainer]) => {
         it('should set maintainer address', async () => {
             expect(await this.framework.getMaintainer()).to.equal(maintainer);
         });
+
+        it('should set semver string', async () => {
+            const sha = childProcess.execSync('git rev-parse HEAD').toString().trim()
+                .substring(0, 7);
+            await this.framework.setVersion(`1.0.1+${sha}`, { from: maintainer });
+            expect(await this.framework.getVersion()).to.equal('1.0.1'.concat('+', sha));
```
kevsul: What happened to the law being the law? 😃

kevsul:
> Is this version somehow related to the version from eip-712 domain separator?
Obviously we’d like to match at most only major version, if at all.

Yeah, I think the EIP712 version is separate from the plasma framework contract version. In fact, the verifyingContract field should probably be the ExitGame contract address 🤔

pnowosie:
Let me highlight part of @boolafish comment and I'll follow up on later sentence.

> It's a nice idea. However, how do we propose to do the version upgrade when adding new exit games? 

Good Q - how it will work @InoMurko? Lets provide some scenarios. 
IMHO this versioning schema will not version the plasma_framework contract (we can change it but this means entirerly new network) - so the major number won't be changed(?)

Lets say we register new exit game - then we just bump the minor?
We register new Vault ->  just bump the minor?

> Now we have a setter that bumps the version to framework, but we probably cannot upgrade the version embedded into EIP712?

That's right! It'd mean the new network - because it disavows all prev signatures.

PR #574, opened 2020-02-04
feat: add simple anti front running fix

kevsul:
Simple front running protection.
Challenger has to provide data derived from sender, so that simply re-submitting the tx with a higher gas price doesn't work

https://github.com/omisego/security-issues/issues/11

There have been 3 options proposed:

1. Add the sender's address (or hash of it) as a parameter to the challenge functions
2. Add signature body (65bytes) into the tx and check if the msg.sender is the same as signer of the signature. 
3.  PoW concept - so:
`hash(txbytes || nonce || sender) = "0x00000000" <> anything`
then contract code would just compute additional hash and check required zeros

Option 1 is obviously the simplest to implement for both the contracts and the callers, and should defeat most generic front running scripts. Defeating front runners who are specifically targeting challenges will probably require more thought.

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
@@ -74,6 +74,8 @@ library PaymentChallengeIFENotCanonical {
     )
         public
     {
+        require(args.senderData == keccak256(abi.encodePacked(msg.sender)), "Incorrect senderData");
+
         uint160 exitId = ExitId.getInFlightExitId(args.inFlightTx);
         PaymentExitDataModel.InFlightExit storage ife = inFlightExitMap.exits[exitId];
         require(args.inFlightTxInputIndex < ife.inputs.length, "Input index out of bounds");
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
@@ -74,8 +74,6 @@ library PaymentChallengeIFENotCanonical {
     )
         public
     {
-        require(args.senderData == keccak256(abi.encodePacked(msg.sender)), "Incorrect senderData");
-
         uint160 exitId = ExitId.getInFlightExitId(args.inFlightTx);
         PaymentExitDataModel.InFlightExit storage ife = inFlightExitMap.exits[exitId];
         require(args.inFlightTxInputIndex < ife.inputs.length, "Input index out of bounds");
--- a/plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
+++ b/plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
@@ -59,6 +59,7 @@ library PaymentInFlightExitRouterArgs {
         uint256 competingTxPos;
         bytes competingTxInclusionProof;
         bytes competingTxWitness;
+        bytes32 senderData;
     }
 
     /**
@@ -79,6 +80,7 @@ library PaymentInFlightExitRouterArgs {
         bytes challengingTxWitness;
         bytes inputTx;
         uint256 inputUtxoPos;
+        bytes32 senderData;
     }
 
      /**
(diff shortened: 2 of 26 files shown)


boolafish:
A question, why sha the address instead of checking address directly?

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
```diff
@@ -74,6 +74,8 @@ library PaymentChallengeIFENotCanonical {
     )
         public
     {
+        require(args.senderData == keccak256(abi.encodePacked(msg.sender)), "Incorrect senderData");
```
pdobacz: doesn't the "respond to non-canonical challenge" (not sure of the exact name) require one as well?

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
```diff
@@ -59,6 +59,7 @@ library PaymentInFlightExitRouterArgs {
         uint256 competingTxPos;
         bytes competingTxInclusionProof;
         bytes competingTxWitness;
+        bytes32 senderData;
```
pdobacz: Do the integration docs and the `.md` files documenting the contracts reflect this change?

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
kevsul: Hmm, good point. But maybe neither "challengeInFlightExitNotCanonical" nor "respondToNonCanonicalChallenge" need this protection right now.

That's because the IFE bond doesn't get paid out on a canonical challenge or response - the _owner_ of the bond is changed here, but the bond doesn't get paid out until processExit is called.

So while a front runner could potentially benefit by front running a canonical challenge or response, they would have to be very familiar with how the exit game works.

Um. I'm inclined to add it in anyway though, for consistency with the other methods, and possible an easier path to adding more sophisticated front running protection later.

What do you think?

plasma_framework/contracts/src/exits/payment/routers/PaymentInFlightExitRouterArgs.sol
kevsul: Yep, will update them

plasma_framework/contracts/src/exits/payment/controllers/PaymentChallengeIFENotCanonical.sol
pdobacz: ah you're right! I'd definitely vote for consistent, to not surprise our future selves reading the code.

Just the question now is - what will be consistent:
 - `senderAddress` on anything resembling a challenge and related to any bond operation or
 - `senderAddress` on anything that actually can pay anything out.

On 2nd thought, the latter makes more sense, so it seems it should be dropped from `challengeIFENotCanon.`. The downside I see is that if sometime, someone decides that those call should pay some portion of the bond (for whatever reason), they'd have to remember to do `senderAddress` check.

Not 100% sure here.

kevsul: I'm going with option 2; only putting senderData on anything that actually pays out.
Trusting our future selves to do the right thing if or when we implement a more sophisticated front running solution

kevsul:
Wow, sorry @boolafish - I typed out an answer for this last week, but it must not have got sent correctly 😞 

> A question, why sha the address instead of checking address directly?

It's to (slightly) obfuscate the sender address, so that a script won't simply replace it with their own address

kevsul:
Dammit, I has a longer description as well, but it got lost too! 😠

I'll try to remember what I wrote originally...

plasma_framework/docs/integration-docs/integration-doc.md
```diff
@@ -547,7 +547,8 @@ PaymentExitGame.challengeStandardExit({
   exitingTx,
   challengeTx,
   inputIndex,
-  witness
+  witness,
+  spenderData
```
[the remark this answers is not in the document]
kevsul: Great catch!

pdobacz:
hm, sorry, I could swear I did an approve :heavy_check_mark: when placing that comment about `spender`, but it seems I only commented

kevsul:
> hm, sorry, I could swear I did an approve ✔️ when placing that comment about `spender`, but it seems I only commented

You had approved @pdobacz , but the later commits forced a re-review 😄

PR #580, opened 2020-02-13
chore: bump version number to 0.0.2

kevsul:


InoMurko:
Indeed 574 breaks the API's! @paulperegud 👍

pdobacz:
> This is major version upgrade.

SemverWatch :trollface:.

But I agree it is :(. Is there an option of pulling back to `v0.x.y` for more liberty? Or keeping the backwards compatible but deprecated call around? Sigh...

kevsul:
> This is major version upgrade.
> 
> #574
> 
> So, `2.0.0`

I was waiting for this! 😄 

TBH we shouldn't be at 1.0.0 yet - we're still in initial development and haven't released anything to the public yet.

I'm happy to pull it back to 0.0.2

Issue #584, opened 2020-02-14
IFE eating piggyback bond on the wrong side


boolafish:
## Issue Type 

```
[x] bug report
[ ] feature request
```
Kudo @pdobacz for realizing this as an real issue and not MoreVP design.
Slack discussion: [link](https://omisego.slack.com/archives/C7V603752/p1581666512116600)

## Current Behavior

Once piggybacked on the wrong side of canonicity, user would not get the piggyback bond back. The bond would be lock on the contract forever.

## Expected Behavior
Return bond even on the wrong side of canonicity.

## Steps to Reproduce
```
Alice creates tx1 spending utxo1 to Bob
Alice submits, tx1 withheld, Alice IFEs
Alice piggybacks on utxo1
Bob piggybacks on his output
Ife finalizes canonically
Alice doesn't get pb bond back
```

```
Alice and Carol create tx1 spending utxo1 and `utxo2` respectively to Bob
Alice submits, tx1 withheld, Alice IFEs
Alice piggybacks on utxo1 , Carol piggybacks on utxo2
Bob piggybacks on his output
Ife finalizes canonically
Alice & Carol don't get pb bond back, 
```

## Suggested Fix
Implement return piggyback bond not only on one side.

## Motivation for Change
protocol correctness

pdobacz:
I'd suggest adding those follow-up items to this effort:
  - generate more events to let consumers know in the cases, where there are only input piggybacks and the IFE finalizes canonically (and vice versa). Currently, there would be no events I think, and also `deleteIFE` facility would not work
  - update the `morevp.md` to be more explicit about the scenarios, where the pb bond must be returned. This could be the 2 scenarios posted in the description :point_up: 
  - when all of that is done, I think `elixir-omg` must also recognize the scenarios where there are piggybacks "on the other end", i.e. input piggybacks for canonical and output piggybacks for non-canonical. I'm afraid this is not catered for and would require the said events to fully work.

also :heart: the issue title :joy:, much jargon

PR #585, opened 2020-02-17
ife stop eating wrong side bond

boolafish:
### Note
- First fix for #584 , not closing the issue for other doc fixes and event improvement. Now all piggyback bond would be return while processIFE.
- Bumping version to 1.0.3 for this fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased - 2020-02-13
+## Unreleased - 2020-02-18
+
+## [1.0.3] - 2020-02-18
+- In-flight exit would return piggyback bond even if user piggyback the wrong canonicity. ([#585](https://github.com/omisego/plasma-contracts/pull/585))
 
 ## [1.0.2] - 2020-02-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased - 2020-02-18
 
 ## [1.0.3] - 2020-02-18
-- In-flight exit would return piggyback bond even if user piggyback the wrong canonicity. ([#585](https://github.com/omisego/plasma-contracts/pull/585))
+- In-flight exit returns all unchallenged piggyback bonds even if user piggybacks the wrong canonicity. ([#585](https://github.com/omisego/plasma-contracts/pull/585))
 
 ## [1.0.2] - 2020-02-13
 
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
@@ -76,14 +76,6 @@ library PaymentProcessInFlightExit {
 
                 if (shouldWithdrawInput(self, exit, withdrawal, token, i)) {
                     withdrawFromVault(self, withdrawal);
-                    bool success = SafeEthTransfer.transferReturnResult(
-                        withdrawal.exitTarget, withdrawal.piggybackBondSize, self.safeGasStipend
-                    );
-
-                    // we do not want to block a queue if bond return is unsuccessful
-                    if (!success) {
-                        emit InFlightBondReturnFailed(withdrawal.exitTarget, withdrawal.piggybackBondSize);
-                    }
                     emit InFlightExitInputWithdrawn(exitId, i);
                 }
             }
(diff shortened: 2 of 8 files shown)


plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -259,6 +247,50 @@ library PaymentProcessInFlightExit {
...
+    function returnInputPiggybackBonds(
+        Controller memory self,
+        PaymentExitDataModel.InFlightExit storage exit,
+        address token
+    )
+        private
+    {
+        for (uint16 i = 0; i < PaymentTransactionModel.MAX_INPUT_NUM(); i++) {
+            PaymentExitDataModel.WithdrawData memory withdrawal = exit.inputs[i];
+            if (token == withdrawal.token && exit.isInputPiggybacked(i)) {
```
pdobacz: this might be up to the wording (and possibly require a clarifying comment), but I'm assuming here that invalid and challenged piggybacks **won't** return the bond, right?

plasma_framework/test/src/exits/payment/controllers/PaymentProcessInFlightExit.test.js
```diff
@@ -464,6 +465,14 @@ contract('PaymentProcessInFlightExit', ([_, ifeBondOwner, inputOwner1, inputOwne
                 expect(postBalance).to.be.bignumber.equal(expectedBalance);
             });
 
+            it('should return piggyback bond to the output owner', async () => {
```
pdobacz: is there testing that a challenged invalid piggyback doesn't return the bond? Also are both canonicity-cases covered in here?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
boolafish: Thanks for pointing this out. I just double checked the challenge piggyback code. So if it challenged, it would return the bond immediately and delete the piggyback flag. So this implementation is fine.

boolafish: But I definitely have the wrong impression in my mind when making the code change....I thought challenge piggyback would end up changing the piggyback bond owner.

pdobacz: OK, maybe consider leaving a comment about challenged piggybacks - that this is taken care of b/c `isInputPiggybacked` returns `false`.

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -259,6 +247,54 @@ library PaymentProcessInFlightExit {
...
+        Controller memory self,
+        PaymentExitDataModel.InFlightExit storage exit,
+        address token
+    )
+        private
+    {
+        for (uint16 i = 0; i < PaymentTransactionModel.MAX_INPUT_NUM(); i++) {
+            PaymentExitDataModel.WithdrawData memory withdrawal = exit.inputs[i];
+
+            // challenged input would remove the piggyback flag
```
kevsul:
```suggestion
            // If the input has been challenged, isInputPiggybacked() will return false
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -259,6 +247,54 @@ library PaymentProcessInFlightExit {
...
+        Controller memory self,
+        PaymentExitDataModel.InFlightExit storage exit,
+        address token
+    )
+        private
+    {
+        for (uint16 i = 0; i < PaymentTransactionModel.MAX_OUTPUT_NUM(); i++) {
+            PaymentExitDataModel.WithdrawData memory withdrawal = exit.outputs[i];
+
+            // challenged output would remove the piggyback flag
```
kevsul:
```suggestion
            // If the output has been challenged, isOutputPiggybacked() will return false
```

boolafish:
@kevsul @thec00n to make things simple, added 1.0.3 version bumping within this PR as well.

CHANGELOG.md
```diff
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased - 2020-02-13
+## Unreleased - 2020-02-18
+
+## [1.0.3] - 2020-02-18
+- In-flight exit would return piggyback bond even if user piggyback the wrong canonicity. ([#585](https://github.com/omisego/plasma-contracts/pull/585))
```
kevsul:
```suggestion
- In-flight exit returns all unchallenged piggyback bonds even if user piggybacks the wrong canonicity. ([#585](https://github.com/omisego/plasma-contracts/pull/585))
```

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -259,6 +247,54 @@ library PaymentProcessInFlightExit {
...
+    }
+
+    function returnOutputPiggybackBonds(
+        Controller memory self,
+        PaymentExitDataModel.InFlightExit storage exit,
+        address token
+    )
+        private
+    {
+        for (uint16 i = 0; i < PaymentTransactionModel.MAX_OUTPUT_NUM(); i++) {
```
thec00n: Why looping over `PaymentTransactionModel.MAX_OUTPUT_NUM` and not the length of outputs?

paulperegud: Agree. With RLP decoding fixed - lets iterate through length of inputs.

boolafish: So `exit.inputs.length` and `exit.outputs.length` would be exactly same as `MAX_INPUT_NUM` and `MAX_OUTPUT_NUM` as that is how it is defined in the storage:
https://github.com/omisego/plasma-contracts/blob/bfbf7cc76cb2ba4564cc1e922dbc3a9bacb8c725/plasma_framework/contracts/src/exits/payment/PaymentExitDataModel.sol#L73-L74

However, agree it is more self explaining if using `in/outputs.length` instead.

PR #588, opened 2020-02-21
feat: convert to PaymentExitGame.getInFlightExitIds(bytes[])

unnawut:


unnawut:
Will do this in https://github.com/omisego/ex_plasma instead. Thanks @boolafish!

PR #591, opened 2020-02-27
fix: fix broken canonicity in IFE processing

pgebal:
Closes https://github.com/omisego/security-issues/issues/13

### DIFF
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
@@ -65,10 +65,25 @@ library PaymentProcessInFlightExit {
             return;
         }
 
-        // Check whether any input is already spent. Required to prevent operator stealing funds.
-        // See: https://github.com/omisego/plasma-contracts/issues/102#issuecomment-495809967
-        // Also, slightly different from the solution above
-        // see: https://github.com/omisego/security-issues/issues/13
+        /* To prevent a double spend, it is needed to know if an output can be exited.
+         * An output can not be exited if:
+         * - it is finalized by a standard exit
+         * - it is finalized by an in-flight exit as input of a non-canonical transaction
+         * - it is blocked from exiting, because it is an input of a canonical transaction
+         *   that exited from one of it's outputs
+         * - it is finalized by an in-flight exit as an output of a canonical transaction
+         * - it is an output of a transaction for which at least one of its inputs is already finalized
+         *
+         * Hence, Plasma Framework stores each output with an exit id that finalized it.
+         * When transaction is marked as canonical but any of it's input was finalized by
+         * other exit, it is not allowed to exit from the transaction's outputs.
+         * In that case exit from an unspent input is possible.
+         * When all inputs of a transaction that is marked as canonical are either not finalized or finalized
+         * by the same exit (which means they were marked as finalized when processing the same exit for a different token),
+         * only exit from outputs is possible.
+         *
+         * See: https://github.com/omisego/plasma-contracts/issues/102#issuecomment-495809967 for more details
+         */
         if (!exit.isCanonical || isAnyInputFinalizedByOtherExit(self.framework, exit, exitId)) {
             for (uint16 i = 0; i < exit.inputs.length; i++) {
                 PaymentExitDataModel.WithdrawData memory withdrawal = exit.inputs[i];
@@ -112,8 +127,6 @@ library PaymentProcessInFlightExit {
         }
     }
 
-    // input is unspent for a given exitId if it's not marked in finalized outputIds map
-    // or it's marked with the same exitId
     function isAnyInputFinalizedByOtherExit(
         PlasmaFramework framework,
         PaymentExitDataModel.InFlightExit memory exit,
--- a/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
+++ b/plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
@@ -123,18 +123,18 @@ library PaymentProcessInFlightExit {
         view
         returns (bool)
     {
-        uint256 inputNumOfTheToken;
+        uint256 nonEmptyInputIndex;
         for (uint16 i = 0; i < exit.inputs.length; i++) {
             if (!exit.isInputEmpty(i)) {
-                inputNumOfTheToken++;
+                nonEmptyInputIndex++;
             }
         }
-        bytes32[] memory outputIdsOfInputs = new bytes32[](inputNumOfTheToken);
-        inputNumOfTheToken = 0;
+        bytes32[] memory outputIdsOfInputs = new bytes32[](nonEmptyInputIndex);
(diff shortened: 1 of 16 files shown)


plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -80,18 +75,25 @@ library PaymentProcessInFlightExit {
...
+                        withdrawFromVault(self, withdrawal);
+                        emit InFlightExitOutputWithdrawn(exitId, i);
+                    }
                 }
-            }
 
-            flagOutputsWhenCanonical(self.framework, exit, token);
+                flagOutputsWhenCanonical(self.framework, exit, token, exitId);
+            }
+            //TODO: possibly emit an event that output exit was skipped
```
pgebal: I don't think if it's needed, buy other opinions are welcomed.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -817,3 +817,40 @@ def start_ife_piggyback_and_process(spend_id, owner, testlang):
...
+    testlang.process_exits(token.address, 0, 1)
+    testlang.process_exits(NULL_ADDRESS, 0, 1)
+
+    # caroline exits with the tokens (previously owned by alice)
+    caroline_token_balance = token.balanceOf(caroline.address)
+    assert caroline_token_balance == caroline_token_balance_before + amount_token
+
+    # but she can not exit with Eth
+    caroline_eth_balance = testlang.get_balance(caroline)
+    # TODO: piggyback bond is still returned, I am not sure if that's good or bad
```
pgebal: We still return piggyback bond in that case. That's not harmful. Someone have other opinions?

pgebal:
Looks like I made a mistake by disallowing to exit from inputs of 'canonical' transaction when some other input is spent. I'm bringing that back in next commit.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -817,3 +817,117 @@ def start_ife_piggyback_and_process(spend_id, owner, testlang):
...
```
pgebal: I do not think that's harmful. Any other ideas?

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -817,3 +817,117 @@ def start_ife_piggyback_and_process(spend_id, owner, testlang):
     testlang.piggyback_in_flight_exit_input(spend_id, 0, owner)
     testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
     testlang.process_exits(NULL_ADDRESS, 0, 10)
+
+
+def test_should_be_resistant_to_Thomalla_heist(testlang, plasma_framework, token):
```
pdobacz: the heist name is jargon, let's be explicit here about what the scenario is

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
pdobacz: if the piggybacks were valid (i.e. not attempting to double spend) per se, the bonds should be returned.

boolafish:
Not 100% sure, but since new exit can override the finalization exitId, it seems like to be possible to have bug?

tx1: should be canonical, ETH and OMG
tx2: should be non-canonical due to input flagged by tx1

Now ETH of tx1 is process first. Since canonical, all inputs + ETH outputs would be flagged with `exitId` of tx1. Now tx2 ETH get processed. Since non-canonical (due to input flagged), it would re-flag the inputs with `exitId` of tx2. Not when OMG of tx1 get processed, it would be considered non-canonical.....

pdobacz:
> but since new exit can override the finalization exitId

@boolafish not sure what you mean "can override"? But indeed this sounds dangerous. Is there an alicebobbing scenario that could illustrate the situation?

boolafish:
The implementation looks like the later exit can flag the output with new exitId

pgebal:
> Not 100% sure, but since new exit can override the finalization exitId, it seems like to be possible to have bug?
> 
> tx1: should be canonical, ETH and OMG
> tx2: should be non-canonical due to input flagged by tx1
> 
> Now ETH of tx1 is process first. Since canonical, all inputs + ETH outputs would be flagged with `exitId` of tx1. Now tx2 ETH get processed. Since non-canonical (due to input flagged), it would re-flag the inputs with `exitId` of tx2. Not when OMG of tx1 get processed, it would be considered non-canonical.....

right, exitId needs to be sticky, set only if prev. it was 0.

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -113,30 +112,32 @@ library PaymentProcessInFlightExit {
         }
     }
 
-    function isAnyInputSpent(
+    // input is unspent for a given exitId if it's not marked in finalized outputIds map
+    // or it's marked with the same exitId
```
paulperegud: First line is rather obvious - follows from the name of the function. Second line's message is very not obvious - question pops up immediately: "why would it ever be marked by the same exitId?"

This is "what" comment, not "why" comment.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
paulperegud: IMHO priorities should be in this order:
1. honest players never lose their bonds
2. prevent griefing
3. care about simplicity
4. punish wrongdoers

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -67,10 +67,9 @@ library PaymentProcessInFlightExit {
 
         // Check whether any input is already spent. Required to prevent operator stealing funds.
         // See: https://github.com/omisego/plasma-contracts/issues/102#issuecomment-495809967
-        // Also, slightly different from the solution above, we treat input spent as non-canonical.
-        // So an IFE is only canonical if all inputs of the in-flight tx are not double spent by competing tx or exit.
-        // see: https://github.com/omisego/plasma-contracts/issues/470
-        if (!exit.isCanonical || isAnyInputSpent(self.framework, exit, token)) {
+        // Also, slightly different from the solution above
+        // see: https://github.com/omisego/security-issues/issues/13
```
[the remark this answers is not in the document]
paulperegud: @pgebal please document this solution as the comment in the code. Answering the question: "why we are doing this?"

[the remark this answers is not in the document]
pgebal: @paulperegud  I put some comments explaining why it's done. Please take a look at them. Are they clear?
@boolafish I removed that second link, I hope my comment is enough. If it's not, I can either expand the comment or make the issue public after it's fixed.

PR #593, opened 2020-03-04
fix ife input spend challenge using wrong index

boolafish:
### Note

This is the fix for issue: omisego/security-issues#15

1. Add py tests to cover the scenerio
1. fix the contract using the right index

### Test

The test in this [commit](https://github.com/omisego/plasma-contracts/commit/5ee9e3936bb329347408b9db77273527a77c3bc3) would fail before making the contract change. `VM Exception while processing transaction: revert Output index out of bounds`


plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
```diff
@@ -20,6 +21,97 @@ def test_challenge_in_flight_exit_input_spent_should_succeed(testlang, period):
     assert not in_flight_exit.input_piggybacked(0)
 
 
+
+@pytest.mark.parametrize("num_outputs", [1, 2, 3, 4])
+def test_challenge_in_flight_exit_should_succeed_for_all_output_indexes(testlang, num_outputs):
```
boolafish: So this test actually would not fail even without the fix

plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
```diff
@@ -20,6 +21,97 @@ def test_challenge_in_flight_exit_input_spent_should_succeed(testlang, period):
...
+    testlang.piggyback_in_flight_exit_input(ife_tx_id, 0, double_spend_owner)
+    testlang.forward_to_period(1)
+    testlang.challenge_in_flight_exit_input_spent(ife_tx_id, challenge_tx_id, double_spend_owner)
+
+    in_flight_exit = testlang.get_in_flight_exit(ife_tx_id)
+    assert not in_flight_exit.input_piggybacked(double_spend_output_index)
+
+
+@pytest.mark.parametrize("challenge_index", [0, 1, 2, 3])
+def test_challenge_in_flight_exit_should_succeed_for_all_challenge_indexes(testlang, challenge_index):
```
boolafish: this is the test that would actually fail

plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
pdobacz: I think not that surprisingly:
 - the problem was with checking just the spending condition
 - the index of the double spend in the challenge tx is always 0, where owner[0] signs
 - this coincides with the input transaction having owner[0]'s output on position 0.

So even though the double spend would not necessarily be of owner[0] and at position 0, the spending condition would always pass.

pdobacz:
Suggesting a parametrized test for one of the other cases (`ChallengeIFEOutputSpent`) here https://github.com/omisego/plasma-contracts/pull/594, if this is useful

pdobacz:
On the fix itself: it looks on point, I'd however go for the other similar test cases, like the one I suggested (see https://github.com/omisego/plasma-contracts/pull/594). Such a test would fit for:
  - challenging IFE input (@pik694) working on this
  - challenging IFE output https://github.com/omisego/plasma-contracts/pull/594
  - challenging standard exits
  - challenging canonicity
  - responses to canonicity challenge

plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
boolafish: > this coincides with the input transaction having owner[0]'s output on position 0.

Yes, but other than 0, the index would mismatch. I was expecting this to break.

However because we are only using `outputType` from the output we retrieve, even the index mismatches, it would end up being same output type since we have only one type atm.

plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
```diff
@@ -1,11 +1,12 @@
...
 from eth_tester.exceptions import TransactionFailed
 from plasma_core.constants import NULL_ADDRESS
+from plasma_core.utils.transactions import decode_utxo_id, encode_utxo_id
 
 
 # should succeed even when phase 2 of in-flight exit is over
-@pytest.mark.parametrize("period", [1, 2, 4])
-def test_challenge_in_flight_exit_input_spent_should_succeed(testlang, period):
+@pytest.mark.parametrize("period", [1, 2, 3, 4])
+def test_challenge_in_flight_exit_input_spent_should_succeed_in_all_periods(testlang, period):
```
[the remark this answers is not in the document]
boolafish: I updated the test name a bit, can you check again?

> it forwards just to the second and then challenges

No, it actually forward to the period. Or at least, it forward to `period * 1/2 MIN_EXIT_PERIOD`

plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_input_spent.py
```diff
@@ -20,6 +21,96 @@ def test_challenge_in_flight_exit_input_spent_should_succeed(testlang, period):
...
+def test_challenge_in_flight_exit_should_succeed_for_all_output_indexes(testlang, num_outputs):
+    deposit_amount = 100
+    deposit_id = testlang.deposit(testlang.accounts[0], deposit_amount)
+
+    owners = []
+    outputs = []
+    tx_output_amount = deposit_amount // num_outputs
+    for i in range(0, num_outputs):
+        owners.append(testlang.accounts[i])
+        outputs.append((testlang.accounts[i].address, NULL_ADDRESS, tx_output_amount))
```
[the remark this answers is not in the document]
boolafish: I changed the way a bit, and also just realize that `owners` list is totally useless as the `spend_utxo` function would be only using the first owner when signing under the hood.

Please re-check whether your comment still hold?

PR #594, opened 2020-03-04
add challenge output big test

pdobacz:
This expands a related PR #593 with a (passing) test added.

TODO:
 - [ ] decide whether doing all the cases with `pytest.parametrize` is the good way. It produces 100 cases each lasting a few seconds. But it seems like a safer first bet, so optimizing the test experience might be a different story

cc/ @boolafish @pik694 @kevsul

### DIFF
--- a/plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_output_spent.py
+++ b/plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_output_spent.py
@@ -2,6 +2,7 @@ import pytest
 from eth_utils import keccak
 from eth_tester.exceptions import TransactionFailed
 from plasma_core.constants import NULL_ADDRESS
+from plasma_core.utils.transactions import decode_utxo_id, encode_utxo_id
 
 
 # challenge should succeed even when phase 2 of in-flight exit is over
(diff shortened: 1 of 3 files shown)


plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_output_spent.py
```diff
@@ -21,6 +22,57 @@ def test_challenge_in_flight_exit_output_spent_should_succeed(testlang, period):
...
+        [
+            (double_spend_owner.address, NULL_ADDRESS, tx_output_amount),
+        ],
+        force_invalid=True
+    )
+
+    testlang.challenge_in_flight_exit_output_spent(ife_tx_id, challenge_tx_id, double_spend_output_index, double_spend_owner)
+
+    in_flight_exit = testlang.get_in_flight_exit(ife_tx_id)
+    MAX_INDEX_SIZE = 4
```
pgebal: Please put constant definitions on top of the file.

pdobacz: yah, was copy-pasting from #593. I think this needs to be cleaned in bulk there, after this PR-to-a-PR lands

boolafish:
Hi, do you mind to PR this against master directly as there is no dependencies on the two PRs so we don't need to wait/block each other.

I do like the idea though, will play around in my PR too.

boolafish:
ah.....so just realize this PR itself is adding 30 mins to the py test job. I doubt is is acceptable to do this for all scenario we would like to cover?

Probably we could lower down to 2 dimensions, which is the num of inputs and outputs. The index always be `num - 1` which would cover all indexes as well.

pdobacz:
Agreed on that the test running time might be off the limits, however:

> Probably we could lower down to 2 dimensions

I'd not go there TBH. I see alternatives:
  - put the loop inside the test (@pik694 's suggestion)
  - speed up the test setup-teardown somehow, for example by snapshotting `ganache-cli`'s state. I imagine this could be possible.
  - paralelize tests in CircleCI with `-n` option maybe?

We can consider adding those long-running tests now and optimizing the testing experience later too.

pdobacz:
changed base to `master`, so need to merge #593 before this one

boolafish:
> put the loop inside the test (@pik694 's suggestion)

How does this help???

pdobacz:
> > put the loop inside the test (@pik694 's suggestion)
> 
> How does this help???

Hm, you setup/teardown once per all the parameter combinations, right @pik694 ? My understanding is that the setup (contract deployment) takes a significant portion of those 30 minutes

pdobacz:
> test would run for ~8 seconds, whereas the setup and teardown would take only ~2 seconds altogether

ah gotcha

PR #598, opened 2020-03-06
test: all indexes on IFE non canonical challenge and response

boolafish:


plasma_framework/python_tests/tests/contracts/root_chain/test_challenge_in_flight_exit_not_canonical.py
```diff
@@ -2,22 +2,56 @@
...
+    tx_output_amount = deposit_amount // PAYMENT_TX_MAX_OUTPUT_SIZE
+    outputs = [(alice.address, NULL_ADDRESS, tx_output_amount)] * PAYMENT_TX_MAX_OUTPUT_SIZE
+
+    input_tx_id = testlang.spend_utxo([deposit_id], [alice], outputs=outputs)
+    blknum, tx_index, _ = decode_utxo_id(input_tx_id)
+    double_spend_utxo = encode_utxo_id(blknum, tx_index, double_spend_output_index)
+
+    ife_output_amount = tx_output_amount
+    ife_tx_id = testlang.spend_utxo(
+        [double_spend_utxo],
```
pdobacz: I'd say that there still is one dimension to test out here - various indices of the double spent input in the `ife_tx` - for now you're always at `0`. I know that I might be sounding paranoid. At least test it in a separate test, if you're concerned about test running time.

Same comment for the `respond to non-canonical challenge` version of this test

boolafish: Can you take a look on `_indexes_for_test()`. Now I loops through the 3 indexes with this function which loop through each dimension with 2 other indexes being fixed.

boolafish:
If we are okay with this approach, I think we can decide we want to DRY the code a bit or not. But would like to get a 👌 pass first for the strategy.

@pdobacz @pik694

pdobacz:
Ah yeah, this is nice.

Now looking at this, such tests in the CI + full brute-force test on a less frequent basis could give us a big deal of confidence and a better developer experience

PR #599, opened 2020-03-06
Parametrize ife input challenge tests to cover all cases

pik694:
The following changes parametrize ife input challenges in order to test all combinations.

Refers #593 #594


pdobacz:
Looks nice, but I now you got me convinced it's overkill. 35 minutes on CI and only 13% through. I think this counts as prohibitive :(. So let's maybe slice this huge space somehow, at least this test, where there are so many combinations.

I'll defer to you guys' judgement on how to approach this.

boolafish:
lol....the test timeout on circle CI limit of 5 hr XDD

PR #611, opened 2020-04-01
test: add test checking that inputs can not be exited from restarted ife after outputs finalized on processing the first ife

pgebal:

### DIFF
--- a/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
+++ b/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
@@ -1005,6 +1005,14 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
 
 
 def test_after_canonical_ife_is_finalized_inputs_are_not_exited_when_ife_is_restarted_and_non_canonical(testlang, plasma_framework, token):
+    """
+    1. Alice and Bob send a canonical transaction transfering their funds to Alice.
+    2. Alice starts an in-flight exit and exits her output.
+    3. Bob creates a competing transaction spending his input.
+    4. In-flight exit is restarted and is non-canonical as Bob produces a comepeting transaction.
+    5. Alice and Bob piggyback on their inputs and exit is processed.
+    6. Funds from inputs are not withdrawn as they were marked as finalized when Alice exited her output in step 2.
+    """
     alice, bob, deposit_amount = testlang.accounts[0], testlang.accounts[1], 100
     alice_deposit_id = testlang.deposit(alice, deposit_amount)
     bob_deposit_id = testlang.deposit(bob, deposit_amount)
--- a/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
+++ b/plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
@@ -1005,14 +1005,16 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
 
 
 def test_after_canonical_ife_is_finalized_inputs_are_not_exited_when_ife_is_restarted_and_non_canonical(testlang, plasma_framework, token):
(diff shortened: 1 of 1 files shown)


plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -1002,3 +1002,49 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
...
+    testlang.start_in_flight_exit(spend_id)
+    testlang.piggyback_in_flight_exit_output(spend_id, 0, alice)
+    testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
+    testlang.process_exits(NULL_ADDRESS, 0, 1)
+
+    # alice exits with her Eth output
+    alice_eth_balance = testlang.get_balance(alice)
+    assert alice_eth_balance == alice_eth_balance_before + amount * 2
+
+    restart_spend_id = testlang.spend_utxo(
```
boolafish, line 1025: why a `restart_spend_id`. Shouldn't we just startIFE again with `spend_id`?

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -1002,3 +1002,49 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
...
+    alice_eth_balance_before = testlang.get_balance(alice)
+
+    testlang.start_in_flight_exit(spend_id)
+    testlang.piggyback_in_flight_exit_output(spend_id, 0, alice)
+    testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
+    testlang.process_exits(NULL_ADDRESS, 0, 1)
+
+    # alice exits with her Eth output
+    alice_eth_balance = testlang.get_balance(alice)
+    assert alice_eth_balance == alice_eth_balance_before + amount * 2
```
boolafish, line 1023: might worth to put a var on `amount * 2` to clarify what that is (use that on the output when `spend_utxo` too)

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -1002,3 +1002,49 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
     # and does not get the tokens back
     alice_token_balance = token.balanceOf(alice.address)
     assert alice_token_balance == alice_token_balance_before
+
+
+def test_after_canonical_ife_is_finalized_inputs_are_not_exited_when_ife_is_restarted_and_non_canonical(testlang, plasma_framework, token):
```
boolafish, line 1007: Though I am generally not a fan of adding too many comments, but I think it might worth to write down the Alice Bob description for each test. It has become quite complex and the function name honestly would probably not possible to be clear enough if first seeing this without any background info.

plasma_framework/python_tests/tests/contracts/root_chain/test_process_exits.py
```diff
@@ -1002,3 +1002,49 @@ def test_challenged_standard_exit_does_not_block_ife_output_exit(testlang, plasm
...
+
+    testlang.piggyback_in_flight_exit_input(restart_spend_id, 0, alice)
+    testlang.piggyback_in_flight_exit_input(restart_spend_id, 1, bob)
+
+    eth_balance_before_processing_exits = testlang.get_balance(plasma_framework.eth_vault)
+
+    testlang.forward_timestamp(2 * MIN_EXIT_PERIOD + 1)
+    testlang.process_exits(NULL_ADDRESS, 0, 2)
+
+    # users didn't exit their inputs - no funds were withdrawn from the eth vault
```
kevsul, line 1048: Maybe add the reason e.g.
`no funds were withdrawn from the eth vault because when the first exit was processed, the inputs were flagged as finalized`

PR #612, opened 2020-04-02
feat: payment v2 initiate

boolafish:
### Why
We will use payment v2 to initiate some exchange related feature development.

### Changes
- copy `payment/` -> `paymentV2/` and rename the files, contract name...
- did not copy the `PaymentToPaymentSpendingCondition`. might just reuse.

closes: https://github.com/omisego/velocity/issues/19
### Test
unit tests are copied for payment v2 too, CI passed.


boolafish:
> we should keep payment v2 in a separate branch until the requirements themselves are more broadly understood and also how payment v2 will impact the existing system.

How does a separate branch help? Also, the idea is to make it 0 impact on existing system that is why copy-pasting to payment v2. So we can control the deployed contract by simply manipulating the migration script with some flag for our development need.

pik694:
Can we close it?

boolafish:
@pik694 
I would like to hang this for discussion a bit, especially for people outside our team. Also, this change does not requires to be in private repo. We can sync private repo up from here too.

pik694:
But do we want it to be in `plasma-contracts`? I think that it is not necessary.

boolafish:
why not? It is not necessary but it decrease one commit/PR to sync between the mirroring repo.

However, this is exactly what I want to discuss and want other people to jam in. Also want to hear if anyone got some other claver idea.

boolafish:
closing this in flavor of: https://github.com/omisego/plasma-contracts-private/pull/2

PR #618, opened 2020-05-04
all currency is token

InoMurko:


InoMurko:
maybe you guys know what this python error is?
```
ImportError while loading conftest '/home/circleci/repo/plasma_framework/python_tests/tests/conftest.py'.
 
tests/conftest.py:15: in <module>
 
    from tests.tests_utils.constants import (
 
E   ModuleNotFoundError: No module named 'tests.tests_utils'
 
```

kevsul:
Thanks for this @InoMurko, I thought we'd already done this 🤦 

Unfortunately, changing the eip712 hash will have an impact on any code that does signing, such as omg-js, etc. 
Let's discuss with the relevant owners and see how to best sync this change

pgebal:
Why didn't we stay with currency? Suits better. For example ETH is not a token.

InoMurko:
`Ethereum is a shit token.`
I have heard this many many times :)

InoMurko:
The reason I didn't go with currency is simply because of the impact. If I would rename token2currency it would have touched more files.

boolafish:
Oops, did not notice we use `currency` in vault. I did not change the wording in EIP712 as the reason mentioned above....but if we are scoping to make it truly consistent across all repo it will be 💯

InoMurko:
how do we want to proceed? @kevsul @boolafish ?

boolafish:
hmm....how about we do the change on the vault in this PR only as it would not break anything outside.

For the change to EIP712, we create another issue and pull that into gold team, I don't think any feature team has extra engineer budget at this moment? The change has to cover all repos: contract, elixir-omg, omg-js (anything more? BE and wallet??).

Alternatively we try to ask to pull this in next cycle. Though I think if no emergent issue, this is sth gold team can take care of.

kevsul:
I'm still a bit worried about the impact - this is a breaking change, which means all currently deployed envs will be incompatible. We'll have to redeploy staging, integration, etc.

InoMurko:
This is 70% true for any merge into plasma-contracts repo :)

For the mainnet release, are we bundling any other PRs? Something like plasma-contracts 2.0 perhaps? Then this is where this PR should go.

kevsul:
Agreed. 
I think it's too much to be handle by the gold team - let's try to bundle it into the next breaking release.

Issue #623, opened 2020-05-13
[fast exits] Start standard exit from a another contract


pik694:
## Issue Type 

```
[ ] bug report
[x] feature request
```

# Description 
Let's make some tests on starting a SE by a liquidity contract LC. 

The flow should be as following:
1. Alice transfers some value to LC  (utxo0);
1. Alice sends an Ethereum tx to LC with some proof.
1. LC starts a SE on Plasma Framework contracts with the proofs provided by Alice.
1. Required time passes.
1. Alice processes exit queue -> LC receives the funds.

## Additional requirements
- [ ] only the user that transferred the amount to LC can start the SE via LC.

conniemkwan:
Sorry @pik694 I accidentially closed it but have reverted it back to Review.

pik694:
@conniemkwan no problem :slightly_smiling_face:

PR #625, opened 2020-05-15
Fast Exits: Liquidity Contract without NFT v1

souradeep-das:
https://github.com/omisego/plasma-contracts/issues/623
**Pending**

- Logic for Challenged Exits and turning off withdrawals for the same
- Exit Bond payment returns with and without challenged exits
- Integrating an NFT
- Should probably move to **`plasma-contracts/fast_exits`**. Setting up truffle and CI

**Implemented**

- Starting the Exit
- Checking the authenticity of the Tx that created the output that was further spent by the contract
- Only the user that transferred the amount to LC can start the SE via LC.
- Basic Layout for the withdrawal after Process Exits (needs updates)
- Workflow tests to start the exit through LC


plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+        uint256 _utxoPosInput
+    ) internal {
+
+            PaymentTransactionModel.Transaction memory decodedFirstTx
+         = PaymentTransactionModel.decode(_rlpInputCreationTx);
+        uint16 firstTransactionOutputIndex = getOutputIndex(_utxoPosInput);
+
+            FungibleTokenOutputModel.Output memory outputFromFirstTransaction
+         = decodedFirstTx.outputs[firstTransactionOutputIndex];
+        address ownerFirstTxOutput = address(
```
boolafish, line 115: consider `PaymentTransactionModel.getOutputOwner` function instead

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+            uint160(outputFromFirstTransaction.outputGuard)
+        );
+        require(
+            ownerFirstTxOutput == msg.sender,
+            "First Tx Provided was not called by sender"
```
boolafish, line 120: perhaps "Was not called by the first tx owner"?

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+        );
+    }
+
+    /**
+     * @notice Verify the First Tx provided is valid
+     * @param _utxoPosInput position of the output that created the inputs for second transaction
+     * @param _rlpInputCreationTx RLP-encoded first transaction that transfers to this contract
+     * @param _inputCreationTxInclusionProof First transactions inclusion proofs
+    */
+    function verifyTxValidity(
```
boolafish, line 130: I think it might not be necessary to check whether the input tx is included to a block or not (?)
If you start exit from a tx that has input not included and succeed, that is an invalid tx in the block, and the child chain is byzantine so all user would mass exit to safe guard their fund.

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+     * @param _utxoPosToExit position of the output which the contract has to exit
+     * @param _rlpOutputTxToContract RLP-encoded transaction that creates the output for the contract
+     * @param _outputTxToContractInclusionProof Second Transaction's inclusion proof
+    */
+    function runExit(
+        uint256 _utxoPosToExit,
+        bytes memory _rlpOutputTxToContract,
+        bytes memory _outputTxToContractInclusionProof
+    ) internal {
+        PaymentStandardExitRouterArgs.StartStandardExitArgs memory s;
```
boolafish, line 162: minor: slight preference of using the following style instead:
```
PaymentStandardExitRouterArgs.StartStandardExitArgs({
    utxoPos: _utxoPosToExit,
    rlpOutputTx: _rlpOutputTxToContract,
    ....: ...
})
```

reason is that this style would force you to provide information for all fields and avoid some manual mistakes.

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+        s.utxoPos = _utxoPosToExit;
+        s.rlpOutputTx = _rlpOutputTxToContract;
+        s.outputTxInclusionProof = _outputTxToContractInclusionProof;
+        peg.startStandardExit.value(msg.value)(s);
+    }
+
+    function getExitId(bytes memory _txBytes, uint16 _outputIndex)
```
boolafish, line 169: consider to use the module `ExitId`?

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+        uint160 croppedExitId = uint160(exitId);
+        require(uint256(croppedExitId) == exitId, "ExitId overflows");
+        return croppedExitId;
+    }
+
+    /**
+     * @dev Get Amount from contract after exit is processed - (to be updated)
+     * @param _exitId The exit id
+    */
+    function getWithdrawl(uint160 _exitId) public {
```
boolafish, line 185: can we move this public api in place earlier in the code?

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+        require(
+            exitIdtoUser[_exitId] == msg.sender,
+            "Only the exitId owner can get the withdrawl"
+        );
+        uint160[] memory _exitIdList = new uint160[](1);
+        _exitIdList[0] = _exitId;
+        PaymentExitDataModel.StandardExit[] memory exits = peg.standardExits(
+            _exitIdList
+        );
+        if (exits[0].utxoPos == 0) {
```
boolafish, line 195: this is quite interesting solution of verifying the exit has been processed. 
However, I *think* that if the exit got challenge, it would be the same like this as it will be deleted.

Worth to see if you can find some good solution to recognize an exit is finalized, if you find out that making change to existing contracts is required or would be better solution, now (this cycle) is the time to shout it out for discussion!

boolafish:
> Logic for Challenged Exits and turning off withdrawals for the same

What does this mean? Do we need to proxy challenge in the liquidity contract too? 🤔

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: Oh, I think validating the input tx also proves that you are exiting the tx that was meant for you through the LC, since I could provide for a fake tx and could get away starting someone else's exit through the LC ?

boolafish: kinda lost, isn't what you describe done in `verifyOwnership`?
function here is proofing the input tx is indeed included in a block?

pik694: @boolafish We have to check both these conditions.

Scenario:
1. Alice owns on plasma an `utxo1` - created by `tx1`;
1. Alice transfers the `utxo1` to the liquidity contract (LC)  - `tx2` and `utxo2`;
1. Alice starts a SE via LC.

Let's consider what happens without these assertions.
Since `utxo2` is owned by the LC -> anybody (provided he/she has proofs) can start an exit via LC - as the action on plasma framework must be triggered by LC, which in this case can be triggered by anybody. The contract does not have any additional data, so it assumes that the owner of the funds is the `msg.sender`.

To mitigate this, we assumed that we should check who is the owner of `utxo1` - input to `tx2`. Now, we should provide `tx1` to check whether the owner of `utxo1` was indeed `msg.sender`.

But what if Bob provides `tx1'`, where `utxo1'` is differs from `utxo1` just by the owner - we have to check which tx is the valid - this is why we also check inclusion proof.

---
I was thinking about some other approach - do not make such validations at the beginning, but provide a challenge mechanism. Bob could start the SE from `utxo2` - and initially, he owns it, but Alice can then provide proof that she is the real owner -> the NFT is transferred right to Alice, Bob has lost his bonds.

plasma_framework/contracts/src/fast_exits/liquidity.sol
pik694: @boolafish this is a very good point and sth that is actually a blocker - otherwise users' funds may be easily stolen.

pik694:
Refers #623

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: yes yes, and following from what @pik694 said, so `verifyOwnership` checks if you are really the owner of `utxo1` which was an input to `tx2`. And `verifyTxValidity` checks if the `tx1` which you provided to prove the previous statement, is actually valid.

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @dev Call this func to start the exit on Rootchain contract
+     * @param _utxoPosToExit position of the output which the contract has to exit
+     * @param _rlpOutputTxToContract RLP-encoded transaction that creates the output for the contract
+     * @param _outputTxToContractInclusionProof Second Transaction's inclusion proof
+     * @param _rlpInputCreationTx RLP-encoded first transaction that transfers to this contract
+     * @param _inputCreationTxInclusionProof First transactions inclusion proofs
+     * @param _utxoPosInput position of the output that created the inputs for second transaction
+    */
+    function startExitOnRootchainContract(
```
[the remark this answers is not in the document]
souradeep-das: Sure. PlasmaContract? or what should be suitable?

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: And, the challenge mechanism sure is a great idea. But will restricting the problem at the roots be the easier way from a user's perspective?

pik694: > But will restricting the problem at the roots be the easier way from a user's perspective?

That is right, this is why I created a whole issue ( #627) to discuss which path we should take.

plasma_framework/contracts/src/fast_exits/liquidity.sol
pik694: Let's move the discussion to #628

plasma_framework/contracts/src/fast_exits/liquidity.sol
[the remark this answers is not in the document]
pik694: maybe just 'startExit`?

plasma_framework/contracts/src/fast_exits/liquidity.sol
boolafish: oh, thanks for explanation. So the goal is to verify that the `input` is indeed pointing to that specific `tx` provided. Can you rename the function name or try to make the intention more easy to understand?

souradeep-das: Sure! Any suggestions?

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+import "../exits/payment/PaymentExitGame.sol";
+import "../framework/PlasmaFramework.sol";
+import "../transactions/PaymentTransactionModel.sol";
+import "../utils/PosLib.sol";
+import "../framework/models/BlockModel.sol";
+import "../utils/Merkle.sol";
+import "../exits/payment/routers/PaymentStandardExitRouter.sol";
+
+/** @title Liquidity Contract. */
+contract liquidity {
```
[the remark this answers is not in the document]
souradeep-das: fixed

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+
+import "../exits/payment/PaymentExitGame.sol";
+import "../framework/PlasmaFramework.sol";
+import "../transactions/PaymentTransactionModel.sol";
+import "../utils/PosLib.sol";
+import "../framework/models/BlockModel.sol";
+import "../utils/Merkle.sol";
+import "../exits/payment/routers/PaymentStandardExitRouter.sol";
+
+/** @title Liquidity Contract. */
```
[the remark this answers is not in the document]
souradeep-das: done

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+contract liquidity {
+    PaymentExitGame peg;
```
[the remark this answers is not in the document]
souradeep-das: changed to `paymentExitGame`

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+    address targetPaymentExitGameContract;
+
+    PlasmaFramework pf;
```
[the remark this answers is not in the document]
souradeep-das: :+1:

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+import "../transactions/PaymentTransactionModel.sol";
+import "../utils/PosLib.sol";
+import "../framework/models/BlockModel.sol";
+import "../utils/Merkle.sol";
+import "../exits/payment/routers/PaymentStandardExitRouter.sol";
+
+/** @title Liquidity Contract. */
+contract liquidity {
+    PaymentExitGame peg;
+    address targetPaymentExitGameContract;
```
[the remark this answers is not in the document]
souradeep-das: removed

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    PlasmaFramework pf;
+    address targetPlasmaFrameworkContract;
```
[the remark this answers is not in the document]
souradeep-das: removed

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    mapping(address => uint160[]) userExitIds;
+    mapping(uint160 => address) exitIdtoUser;
+    mapping(uint160 => uint256) exitIdtoAmount;
+
+    /**
+     * @notice generates instances of omg-contracts
+    */
+    constructor(address _plasmaFrameworkContract) public {
```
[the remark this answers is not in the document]
souradeep-das: true, fixed

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+
+    function getCurrentBondSize() public view returns (uint128) {
+        return peg.startStandardExitBondSize();
+    }
+
+    /**
+     * @dev gets the index of the output from the utxo position
+     * @param _utxoPos position of the output
+    */
+    function getOutputIndex(uint256 _utxoPos) internal pure returns (uint16) {
```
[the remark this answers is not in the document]
souradeep-das: changed

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: used

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: yes, definitely once the withdrawal method is finalised

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+     * @param _utxoPosInput position of the output that created the inputs for second transaction
+    */
+    function verifyOwnership(
+        bytes memory _rlpInputCreationTx,
+        uint256 _utxoPosInput
+    ) internal {
+
+            PaymentTransactionModel.Transaction memory decodedFirstTx
+         = PaymentTransactionModel.decode(_rlpInputCreationTx);
+        uint16 firstTransactionOutputIndex = getOutputIndex(_utxoPosInput);
```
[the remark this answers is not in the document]
souradeep-das: changed

plasma_framework/contracts/src/fast_exits/liquidity.sol
```diff
@@ -0,0 +1,202 @@
...
+/** @title Liquidity Contract. */
+contract liquidity {
+    PaymentExitGame peg;
+    address targetPaymentExitGameContract;
+
+    PlasmaFramework pf;
+    address targetPlasmaFrameworkContract;
+
+    mapping(address => uint160[]) userExitIds;
+    mapping(uint160 => address) exitIdtoUser;
```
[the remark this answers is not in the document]
souradeep-das: done. using ExitId lib `c3d1e17`

plasma_framework/contracts/src/fast_exits/liquidity.sol
souradeep-das: `c3d1e17`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+        verifyOwnership(rlpInputCreationTx, utxoPosInput);
+
+        PaymentTransactionModel.Transaction memory decodedSecondTx
+        = PaymentTransactionModel.decode(rlpOutputTxToContract);
+        require(
+            decodedSecondTx.inputs[0] == bytes32(utxoPosInput),
+            "Wrong utxoPosInput provided"
+        );
+
+        verifyTxValidity(
```
pik694, line 64: this could function could return a boolean and we could `assert` here instead on asserting lower

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+            utxoPosInput,
+            rlpInputCreationTx,
+            inputCreationTxInclusionProof
+        );
+        runExit(
```
pik694, line 69: same here  - let's assert that we actually did start an exit

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+        uint160 exitId = paymentExitGame.getStandardExitId(false, rlpOutputTxToContract, utxoPosToExit);
+        userExitIds[msg.sender].push(exitId);
+        exitIdtoUser[exitId] = msg.sender;
+
+        // associate the amount exiting to the exitId
+
+        FungibleTokenOutputModel.Output memory outputFromSecondTransaction
+        = decodedSecondTx.outputs[0];
+        uint256 amount = outputFromSecondTransaction.amount;
+        exitIdtoAmount[exitId] = amount;
```
pik694, line 85: this part of the code could be extracted into a function that mints a token

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @notice Check if the person calling is the same person who created the tx to the contract
+     * @param rlpInputCreationTx RLP-encoded first transaction that transfers to this contract
+     * @param utxoPosInput position of the output that created the inputs for second transaction
+    */
+    function verifyOwnership(
+        bytes memory rlpInputCreationTx,
+        uint256 utxoPosInput
+    ) internal {
```
pik694, line 96: this method is rather private

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+     * @notice Verify the First Tx provided is valid
+     * @param utxoPosInput position of the output that created the inputs for second transaction
+     * @param rlpInputCreationTx RLP-encoded first transaction that transfers to this contract
+     * @param inputCreationTxInclusionProof First transactions inclusion proofs
+    */
+    function verifyTxValidity(
+        uint256 utxoPosInput,
+        bytes memory rlpInputCreationTx,
+        bytes memory inputCreationTxInclusionProof
+    ) internal {
```
pik694, line 122: also this one

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+     * @param utxoPosInput position of the output that created the inputs for second transaction
+    */
+    function verifyOwnership(
+        bytes memory rlpInputCreationTx,
+        uint256 utxoPosInput
+    ) internal {
+
+        PaymentTransactionModel.Transaction memory decodedFirstTx
+        = PaymentTransactionModel.decode(rlpInputCreationTx);
+        PosLib.Position memory position = PosLib.decode(utxoPosInput);
```
pik694, line 100: we decode this twice - here and in function that verifies tx validity - probably there is also some more redundant decoding made - please consider refactoring these functions so that they use as many common variables as possible (decoding higher and passing values lower)

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+    */
+    function verifyTxValidity(
+        uint256 utxoPosInput,
+        bytes memory rlpInputCreationTx,
+        bytes memory inputCreationTxInclusionProof
+    ) internal {
+        PosLib.Position memory utxoDecoded = PosLib.decode(utxoPosInput);
+        utxoDecoded.outputIndex = 0;
+        (bytes32 root, ) = plasmaFrameworkInstance.blocks(utxoDecoded.blockNum);
+        require(root != bytes32(""), "Failed to get root of the block");
```
pik694, line 126: is there any other way to compare empty `bytes32`?

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+    */
+    function runExit(
+        uint256 utxoPosToExit,
+        bytes memory rlpOutputTxToContract,
+        bytes memory outputTxToContractInclusionProof
+    ) private {
+        PaymentStandardExitRouterArgs.StartStandardExitArgs memory s;
+        s.utxoPos = utxoPosToExit;
+        s.rlpOutputTx = rlpOutputTxToContract;
+        s.outputTxInclusionProof = outputTxToContractInclusionProof;
```
pik694, line 153: maybe just merge these lines into a constructor

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+        paymentExitGame.startStandardExit.value(msg.value)(s);
+    }
+
+    function getContractBalance() public view returns(uint256) {
+        return address(this).balance;
+    }
```
pik694, line 159: I believe that we do not need this helper

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
```diff
@@ -0,0 +1,265 @@
...
+        };
+
+        describe('Given contracts deployed, exit game and ETH vault registered', () => {
+            before(setupContracts);
+
+            it('should not allow to call processExit from outside of exit game controller contract', async () => {
+                await expectRevert(
+                    this.exitGame.processExit(0, config.registerKeys.vaultId.eth, constants.ZERO_ADDRESS),
+                    'Caller address is unauthorized.',
+                );
```
pik694, line 106: why do we check this?

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
```diff
@@ -0,0 +1,265 @@
...
+                    const aliceBalanceAfterDeposit = new BN(await web3.eth.getBalance(alice));
+                    const ethVaultBalanceAfterDeposit = new BN(await web3.eth.getBalance(this.ethVault.address));
+                    const expectedAliceBalance = this.aliceBalanceBeforeDeposit
+                        .sub(new BN(DEPOSIT_VALUE))
+                        .sub(await spentOnGas(this.aliceDepositReceipt));
+                    const expectedEthVaultBalance = this.ethVaultBalanceBeforeDeposit.add(new BN(DEPOSIT_VALUE));
+
+                    expect(aliceBalanceAfterDeposit).to.be.bignumber.equal(expectedAliceBalance);
+                    expect(ethVaultBalanceAfterDeposit).to.be.bignumber.equal(expectedEthVaultBalance);
+                });
```
pik694, line 127: I believe that we do not want to test it here - we should only test the code that we are actually testing and we own - we should not test Ethereum VM and also Plasma Framework - Plasma Framework in this case is a dependency and has its own test.

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
pik694: we could also emit an event when a token is minted.

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+        );
+        if (exits[0].utxoPos == 0) {
+            exitIdtoUser[exitId] = 0x0000000000000000000000000000000000000000;
+            msg.sender.transfer(exitIdtoAmount[exitId]);
+        } else {
+            revert("Not processed exit");
+        }
+    }
+
+    function() external payable {
```
boolafish, line 183: minor: can we have a comment/document on this function and state that we need the fallback function to allow transfer fund to this smart contract. (This is the reason of this empty function here right?)

plasma_framework/contracts/src/fast_exits/liquidity.sol
boolafish: probably sth like `verifyTxMatchThePosition` ?

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
```diff
@@ -0,0 +1,265 @@
...
+                            await time.increase(time.duration.weeks(2).add(time.duration.seconds(1)));
+
+                            this.LCBalanceBeforeProcessExit = new BN(await this.liquidity.getContractBalance());
+
+                            await this.framework.processExits(config.registerKeys.vaultId.eth, ETH, 0, 1, {
+                                from: alice,
+                            });
+                        });
+
+                        it('should return the output amount plus standard exit bond to the Liquidity Contract', async () => {
```
boolafish, line 252: shouldn't the exit bond be returned to Alice instead?
However, I do think this would need either change to existing contract or add some extra feature in the liquidity contract, could/should be out of scope of this PR. Left a comment on the issue tracking that somewhere though.

souradeep-das: Yes @boolafish, we could also return it to Alice after the contract receives it.(Alice pulls it)

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+import "../../src/framework/models/BlockModel.sol";
+import "../../src/utils/Merkle.sol";
+import "../../src/exits/payment/routers/PaymentStandardExitRouter.sol";
+
+/**
+ * @title Liquidity Contract
+ * Implementation Doc - https://github.com/omisego/research/blob/master/plasma/simple_fast_withdrawals.md
+*/
+contract Liquidity {
+    PaymentExitGame private paymentExitGame;
```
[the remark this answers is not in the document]
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    PlasmaFramework private plasmaFrameworkInstance;
```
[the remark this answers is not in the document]
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+
+    mapping(address => uint160[]) public userExitIds;
+    mapping(uint160 => address) public exitIdtoUser;
+    mapping(uint160 => uint256) public exitIdtoAmount;
```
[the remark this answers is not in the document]
souradeep-das: removing these mappings on the next pr with the nft

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * @title Liquidity Contract
+ * Implementation Doc - https://github.com/omisego/research/blob/master/plasma/simple_fast_withdrawals.md
+*/
+contract Liquidity {
+    PaymentExitGame private paymentExitGame;
+
+    PlasmaFramework private plasmaFrameworkInstance;
+
+    mapping(address => uint160[]) public userExitIds;
```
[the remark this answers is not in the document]
souradeep-das: true, removed them on the next pr, also removing in this one

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -0,0 +1,186 @@
...
+     * @notice provide PlasmaFramework contract-address when deploying the contract
+    */
+    constructor(address plasmaFrameworkContract) public {
+        plasmaFrameworkInstance = PlasmaFramework(plasmaFrameworkContract);
+        paymentExitGame = PaymentExitGame(plasmaFrameworkInstance.exitGames(1));
+    }
+
+    function getCurrentBondSize() public view returns (uint128) {
+        return paymentExitGame.startStandardExitBondSize();
+    }
```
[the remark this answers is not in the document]
souradeep-das: yes, we can use the one from main contract 
removed `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: let's do that for the next pr( nft minting) ?

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: Umm, any idea?

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done `8d760f9`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: true. removed `8d760f9`

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
souradeep-das: removed `8d760f9`

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
souradeep-das: true removed `c4c18de`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: done

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
pik694: `bytes32(0x00)` should work and I think I prefer it this way

souradeep-das: Aah cool, thanks! :+1: I'll make it a point to switch to this in the next PR

PR #631, opened 2020-05-20
New exitId format

### LINKED ISSUE #605: Increase test coverage for exit priority
## Issue Type 

```
[x ] bug report
[ ] feature request
```

## Current Behavior
Three values are wrapped up into the exit priority. Based on the documentation the following assumptions are made on now long the scheme can be used for the exitable timestamp as well as the block numbers.

```
    *   42 bits  - timestamp in seconds (exitable_at); we can represent dates until year 141431
    *   54 bits  - blknum * CHILD_BLOCK_INTERVAL * 10^5 + txindex; 54 bits represent all transactions for 85 years. 
        We are assuming CHILD_BLOCK_INTERVAL = 1000.
    *   160 bits - exit id
```

I could not find a test that confirms the values indicated above.

## Expected Behavior
Implement tests that confirm that the exitable timestamp and the blocknumber can be used for the time durations indicated. Additionally confirm with the tests that neither exitable timestamp or the blocknumber overflow or overwrite the other values packed into exit priority once the value reaches the upper bounds of its limits.

kevsul:
Change exitId and exitPriority

Closes https://github.com/omisego/security-issues/issues/12

Closes #605

### DIFF
--- a/plasma_framework/contracts/poc/fast_exits/Liquidity.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Liquidity.sol
@@ -24,7 +24,7 @@ contract Liquidity is ERC721Full {
         uint256 exitAmount;
     }
 
-    mapping(uint160 => ExitData) private exitData;
+    mapping(uint168 => ExitData) private exitData;
 
     /**
      * @notice provide PlasmaFramework contract-address when deploying the contract
@@ -154,8 +154,7 @@ contract Liquidity is ERC721Full {
      * @param decodedSecondTx decoded second transaction
     */
     function mintNFT(bytes memory rlpOutputTxToContract, uint256 utxoPosToExit, PaymentTransactionModel.Transaction memory decodedSecondTx) private {
-        //change the return type of exitId once the pr to change it has been merged
-        uint160 exitId = paymentExitGame.getStandardExitId(false, rlpOutputTxToContract, utxoPosToExit);
+        uint168 exitId = paymentExitGame.getStandardExitId(false, rlpOutputTxToContract, utxoPosToExit);
         super._mint(msg.sender, exitId);
 
         FungibleTokenOutputModel.Output memory outputFromSecondTransaction
@@ -167,12 +166,12 @@ contract Liquidity is ERC721Full {
      * @dev Get Amount from contract after exit is processed - (to be updated)
      * @param exitId The exit id
     */
-    function getWithdrawal(uint160 exitId) public {
+    function getWithdrawal(uint168 exitId) public {
         require(
             super.ownerOf(exitId) == msg.sender,
             "Only the NFT owner of the respective exit can get the withdrawal"
         );
-        uint160[] memory exitIdList = new uint160[](1);
+        uint168[] memory exitIdList = new uint168[](1);
         exitIdList[0] = exitId;
         PaymentExitDataModel.StandardExit[] memory exits = paymentExitGame.standardExits(
             exitIdList
--- a/plasma_framework/contracts/src/exits/utils/ExitId.sol
+++ b/plasma_framework/contracts/src/exits/utils/ExitId.sol
@@ -5,28 +5,28 @@ import "../../utils/PosLib.sol";
 
 library ExitId {
     using PosLib for PosLib.Position;
-    using Bits for uint160;
+    using Bits for uint168;
     using Bits for uint256;
+
+    uint8 constant internal FIRST_BIT = 167;
 
     /**
      * @notice Checks whether exitId is a standard exit ID
      */
-    function isStandardExit(uint160 _exitId) internal pure returns (bool) {
-        return _exitId.getBit(151) == 0;
+    function isStandardExit(uint168 _exitId) internal pure returns (bool) {
+        return _exitId.getBit(FIRST_BIT) == 0;
     }
 
     /**
      * @notice Given transaction bytes and UTXO position, returns its exit ID
      * @dev Computation of a deposit ID is different to any other tx because txBytes of a deposit tx can be a non-unique value
-     * @notice Output index must be within range 0 - 255
      * @param _isDeposit Defines whether the tx for the exitId is a deposit tx
      * @param _txBytes Transaction bytes
      * @param _utxoPos UTXO position of the exiting output
      * @return _standardExitId Unique ID of the standard exit
      *     Anatomy of returned value, most significant bits first:
-     *     8-bits - output index
      *     1-bit - in-flight flag (0 for standard exit)
-     *     151-bits - hash(tx) or hash(tx|utxo) for deposit
+     *     167-bits - hash(tx) or hash(tx|utxo) for deposit
      */
     function getStandardExitId(
         bool _isDeposit,
@@ -35,14 +35,14 @@ library ExitId {
     )
         internal
         pure
-        returns (uint160)
+        returns (uint168)
     {
         if (_isDeposit) {
             bytes32 hashData = keccak256(abi.encodePacked(_txBytes, _utxoPos.encode()));
-            return _computeStandardExitId(hashData, _utxoPos.outputIndex);
+            return _computeStandardExitId(hashData);
         }
 
-        return _computeStandardExitId(keccak256(_txBytes), _utxoPos.outputIndex);
+        return _computeStandardExitId(keccak256(_txBytes));
     }
 
     /**
@@ -50,20 +50,11 @@ library ExitId {
     * @param _txBytes Transaction bytes
     * @return Unique in-flight exit ID
     */
-    function getInFlightExitId(bytes memory _txBytes) internal pure returns (uint160) {
-        return uint160((uint256(keccak256(_txBytes)) >> 105).setBit(151));
+    function getInFlightExitId(bytes memory _txBytes) internal pure returns (uint168) {
+        return uint168((uint256(keccak256(_txBytes)) >> (256 - FIRST_BIT)).setBit(FIRST_BIT));
     }
 
-    function _computeStandardExitId(bytes32 _txhash, uint16 _outputIndex)
-        private
-        pure
-        returns (uint160)
-    {
-        uint256 exitId = (uint256(_txhash) >> 105) | (uint256(_outputIndex) << 152);
-        uint160 croppedExitId = uint160(exitId);
-
-        require(uint256(croppedExitId) == exitId, "ExitId overflows");
-
-        return croppedExitId;
+    function _computeStandardExitId(bytes32 _txhash) private pure returns (uint168) {
+        return uint168((uint256(_txhash) >> (256 - FIRST_BIT)));
     }
 }
--- a/plasma_framework/contracts/src/exits/utils/ExitId.sol
+++ b/plasma_framework/contracts/src/exits/utils/ExitId.sol
@@ -9,51 +9,51 @@ library ExitId {
     using Bits for uint168;
     using Bits for uint256;
 
-    uint8 constant internal FIRST_BIT = 167;
+    uint8 constant private FIRST_BIT_LOCATION = 167;
 
     /**
      * @notice Checks whether exitId is a standard exit ID
      */
-    function isStandardExit(uint168 _exitId) internal pure returns (bool) {
-        return _exitId.getBit(FIRST_BIT) == 0;
+    function isStandardExit(uint168 exitId) internal pure returns (bool) {
+        return exitId.getBit(FIRST_BIT_LOCATION) == 0;
     }
 
     /**
      * @notice Given transaction bytes and UTXO position, returns its exit ID
      * @dev Computation of a deposit ID is different to any other tx because txBytes of a deposit tx can be a non-unique value
-     * @param _isDeposit Defines whether the tx for the exitId is a deposit tx
-     * @param _txBytes Transaction bytes
-     * @param _utxoPos UTXO position of the exiting output
-     * @return _standardExitId Unique ID of the standard exit
+     * @param isDeposit Defines whether the tx for the exitId is a deposit tx
+     * @param txBytes Transaction bytes
+     * @param utxoPos UTXO position of the exiting output
+     * @return standardExitId Unique ID of the standard exit
      *     Anatomy of returned value, most significant bits first:
      *     1-bit - in-flight flag (0 for standard exit)
      *     167-bits - hash(tx) or hash(tx|utxo) for deposit
      */
     function getStandardExitId(
-        bool _isDeposit,
-        bytes memory _txBytes,
-        PosLib.Position memory _utxoPos
+        bool isDeposit,
+        bytes memory txBytes,
+        PosLib.Position memory utxoPos
     )
         internal
         pure
         returns (uint168)
     {
         bytes32 outputId;
-        if (_isDeposit) {
-            outputId = OutputId.computeDepositOutputId(_txBytes, _utxoPos.outputIndex, _utxoPos.encode());
+        if (isDeposit) {
+            outputId = OutputId.computeDepositOutputId(txBytes, utxoPos.outputIndex, utxoPos.encode());
         } else {
-            outputId = OutputId.computeNormalOutputId(_txBytes, _utxoPos.outputIndex);
+            outputId = OutputId.computeNormalOutputId(txBytes, utxoPos.outputIndex);
         }
 
-        return uint168((uint256(outputId) >> (256 - FIRST_BIT)));
+        return uint168((uint256(outputId) >> (256 - FIRST_BIT_LOCATION)));
     }
 
     /**
     * @notice Given transaction bytes, returns in-flight exit ID
-    * @param _txBytes Transaction bytes
+    * @param txBytes Transaction bytes
     * @return Unique in-flight exit ID
     */
-    function getInFlightExitId(bytes memory _txBytes) internal pure returns (uint168) {
-        return uint168((uint256(keccak256(_txBytes)) >> (256 - FIRST_BIT)).setBit(FIRST_BIT));
+    function getInFlightExitId(bytes memory txBytes) internal pure returns (uint168) {
+        return uint168((uint256(keccak256(txBytes)) >> (256 - FIRST_BIT_LOCATION)).setBit(FIRST_BIT_LOCATION));
     }
 }
--- a/plasma_framework/contracts/src/exits/utils/ExitId.sol
+++ b/plasma_framework/contracts/src/exits/utils/ExitId.sol
@@ -2,6 +2,7 @@ pragma solidity 0.5.11;
 
 import "../../utils/Bits.sol";
 import "../../utils/PosLib.sol";
+import "./OutputId.sol";
 
 library ExitId {
     using PosLib for PosLib.Position;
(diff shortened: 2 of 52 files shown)


plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -2,31 +2,32 @@ pragma solidity 0.5.11;
...
 import "../../utils/PosLib.sol";
+import "./OutputId.sol";
 
 library ExitId {
     using PosLib for PosLib.Position;
-    using Bits for uint160;
+    using Bits for uint168;
     using Bits for uint256;
 
+    uint8 constant internal FIRST_BIT = 167;
```
boolafish, line 12: I guess this can go completely private as it is a hidden logic abstracted by the API already
consider rename this as `FIRST_BIT_LOCATION` instead. When looking at shifting I have always thought 256 - FIRST_BIT is like 256 - 1.

plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -2,31 +2,32 @@ pragma solidity 0.5.11;
...
      * @param _txBytes Transaction bytes
      * @param _utxoPos UTXO position of the exiting output
      * @return _standardExitId Unique ID of the standard exit
      *     Anatomy of returned value, most significant bits first:
-     *     8-bits - output index
      *     1-bit - in-flight flag (0 for standard exit)
-     *     151-bits - hash(tx) or hash(tx|utxo) for deposit
+     *     167-bits - hash(tx) or hash(tx|utxo) for deposit
      */
     function getStandardExitId(
```
boolafish, line 32: minor: If you don't mind to sneak some small refactoring change, we can as well clean up the var name to be without underscore here.

plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -35,35 +36,24 @@ library ExitId {
...
         if (_isDeposit) {
-            bytes32 hashData = keccak256(abi.encodePacked(_txBytes, _utxoPos.encode()));
-            return _computeStandardExitId(hashData, _utxoPos.outputIndex);
+            outputId = OutputId.computeDepositOutputId(_txBytes, _utxoPos.outputIndex, _utxoPos.encode());
+        } else {
+            outputId = OutputId.computeNormalOutputId(_txBytes, _utxoPos.outputIndex);
         }
 
-        return _computeStandardExitId(keccak256(_txBytes), _utxoPos.outputIndex);
+        return uint168((uint256(outputId) >> (256 - FIRST_BIT)));
```
boolafish, line 48: hmmm....shouldn't we first cast to size 168 then shift?

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,36 @@ library ExitPriority {
 
     using PosLib for PosLib.Position;
 
+    uint8 constant internal SIZEOF_TIMESTAMP = 32;
```
boolafish, line 9: same, preference of private var

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,36 @@ library ExitPriority {
...
+    uint8 constant internal SIZEOF_EXITID = 168;
+
     /**
      * @dev Returns an exit priority for a given UTXO position and a unique ID.
      * The priority for Plasma M(ore)VP protocol is a combination of 'exitableAt' and 'txPos'.
-     * Since 'exitableAt' only provides granularity of block, add 'txPos' to provide priority for a transaction.
+     * Since 'exitableAt' only provides granularity of block, adding 'txPos' gives higher priority to transactions added earlier in the same block.
      * @notice Detailed explanation on field lengths can be found at https://github.com/omisego/plasma-contracts/pull/303#discussion_r328850572
```
boolafish, line 16: consider to move the exitId discussion from security-issue to plasma-contract repo and link it here for future code reader too

plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -35,35 +36,24 @@ library ExitId {
...
-        returns (uint160)
-    {
-        uint256 exitId = (uint256(_txhash) >> 105) | (uint256(_outputIndex) << 152);
-        uint160 croppedExitId = uint160(exitId);
-
-        require(uint256(croppedExitId) == exitId, "ExitId overflows");
-
-        return croppedExitId;
+    function getInFlightExitId(bytes memory _txBytes) internal pure returns (uint168) {
+        return uint168((uint256(keccak256(_txBytes)) >> (256 - FIRST_BIT)).setBit(FIRST_BIT));
```
boolafish, line 57: same, first cast then shift?

plasma_framework/test/src/framework/utils/ExitPriority.test.js
```diff
@@ -116,8 +116,8 @@ contract('ExitPriority', () => {
             expect(parsedExitableAt).to.be.bignumber.equal(new BN(exitableAt));
         });
```
thec00n, line 118: This is more of a general comment towards exit priority tests. Can we add a function to `ExitPriority` that returns the tx position from the exit priority and assert in all tests that compute the exit priority that all 3 input parameters are the same than what is extracted from the exit priority number.

plasma_framework/contracts/src/exits/utils/ExitId.sol
kevsul: No, I think this is correct.
We cast to uint256 to convert from bytes32 to an integer. You can't cast a bytes256 directly to uint168.
Then we shift it to get the _first_ 167 bits of the hash

Do you mean that we could use the _last_ 167 bytes of the hash instead? e.g.
```
return uint168(uint256(outputId)) >> 1;
```

plasma_framework/contracts/src/exits/utils/ExitId.sol
kevsul: This is the way we've always done it to get the first 167 bits (previously 159 bits). See comment above

plasma_framework/contracts/src/exits/utils/ExitId.sol
boolafish: ah.....okay now I see that is indeed a correct implementation to get 167 bits out.

and yes, `uint168(uint256(outputId)) >> 1` was indeed what in my mind of getting 167 bits out from the outputId that's why I thought it is a wrong implementation.

plasma_framework/test/src/framework/utils/ExitPriority.test.js
```diff
@@ -96,6 +98,10 @@ contract('ExitPriority', () => {
             const priority = await this.contract.computePriority(exitableAt, txPos, exitId);
             const parsedExitableAt = await this.contract.parseExitableAt(priority);
             expect(parsedExitableAt).to.be.bignumber.equal(new BN(exitableAt));
+            const parsedExitId = await this.contract.parseExitId(priority);
```
boolafish, line 101: Not a fan of just sneaking these extra tests inside the existing ones. It breaks the test description.

I would probably just put this into a new describe, probably `describe('exit priority keeps the exitId and txPos info')` and put a few interested test case there.

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,42 @@ library ExitPriority {
...
+    function computePriority(uint64 exitableAt, PosLib.Position memory txPos, uint168 exitId)
         internal
         pure
         returns (uint256)
     {
-        return (uint256(exitableAt) << 214) | (txPos.getTxPositionForExitPriority() << 160) | uint256(exitId);
+        return (uint256(exitableAt) << (256 - SIZEOF_TIMESTAMP)) | (txPos.getTxPositionForExitPriority() << SIZEOF_EXITID) | uint256(exitId);
     }
 
     function parseExitableAt(uint256 priority) internal pure returns (uint64) {
```
thec00n, line 34: Wouldn't it be useful for a client to be able to call the parsing functions?

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -17,9 +17,10 @@ library PosLib {
     uint256 constant internal TX_OFFSET = 10000;
     
     uint256 constant internal MAX_OUTPUT_INDEX = TX_OFFSET - 1;
-    // since we are using merkle tree of depth 16, max tx index size is 2^16 - 1
+    // Since we are using a merkle tree of depth 16, max tx index size is 2^16 - 1
     uint256 constant internal MAX_TX_INDEX = 2 ** 16 - 1;
-    // in ExitPriority, only 54 bits are reserved for both blockNum and txIndex
+    // The maximum block number we can handle depends on ExitPriority. ExitPriority only uses 54 bits for blockNum + txIndex
+    // This is 180143985094 blocks, which works out to be 85 years of blocks (given a block every 15 seconds)
```
thec00n, line 23:
```suggestion
    // This is 180143985094 blocks, which works out to be 85 years of blocks (given a block occurs every 15 seconds)
```

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
thec00n: In ExitPriorityWrapper

boolafish: meanwhile, actually I am slightly on the fence whether we should have this function at all or not as no production code is using this at this moment. If we only want it for testing, we can put this logic into `ExitPriorityWrapper` as alternative.

but I am not sure which is better actually.

plasma_framework/test/src/framework/utils/ExitPriority.test.js
```diff
@@ -5,6 +5,8 @@ const { expect } = require('chai');
```
thec00n, line 5: There is typo here https://github.com/omisego/plasma-contracts/blob/8e21daf3752e47a537d745fc3d51df5057f96e86/plasma_framework/test/src/framework/utils/ExitPriority.test.js#L15

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,42 @@ library ExitPriority {
...
     /**
      * @dev Returns an exit priority for a given UTXO position and a unique ID.
      * The priority for Plasma M(ore)VP protocol is a combination of 'exitableAt' and 'txPos'.
-     * Since 'exitableAt' only provides granularity of block, add 'txPos' to provide priority for a transaction.
+     * Since 'exitableAt' only provides granularity of block, adding 'txPos' gives higher priority to transactions added earlier in the same block.
      * @notice Detailed explanation on field lengths can be found at https://github.com/omisego/plasma-contracts/pull/303#discussion_r328850572
      * @param exitId Unique exit identifier
      * @return An exit priority
      *   Anatomy of returned value, most significant bits first
-     *   42 bits  - timestamp in seconds (exitable_at); we can represent dates until year 141431
```
thec00n, line 17: Creating the sum of bits over timestamp (42) and blocknum (54) and exit id (160) before was 256. With the new scheme it's 254?

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,42 @@ library ExitPriority {
...
+    function parseExitId(uint256 priority) internal pure returns (uint168) {
+        // Exit ID uses only 168 least significant bits
+        return uint168(priority);
     }
 
-    function parseExitId(uint256 priority) internal pure returns (uint160) {
-        // Exit ID uses only 160 least significant bits
-        return uint160(priority);
+    function parseTxPos(uint256 priority) internal pure returns (uint256) {
+        uint256 pos = (priority >> SIZEOF_EXITID) & (2 ** SIZEOF_TXPOS - 1);
```
thec00n, line 44: Hmm why not left shift after the right shift?

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
```diff
@@ -6,32 +6,42 @@ library ExitPriority {
...
-     *   42 bits  - timestamp in seconds (exitable_at); we can represent dates until year 141431
-     *   54 bits  - blocknum * 10^5 + txindex; 54 bits represent all transactions for 85 years. Be aware that child chain block number jumps with the interval of CHILD_BLOCK_INTERVAL, which would be 1000 in production.
-     *   160 bits - exit id
+     *   32 bits  - The exitable_at timestamp (in seconds); can represent dates until the year 2106
+     *   54 bits  - The transaction position. Be aware that child chain block number jumps with the interval of CHILD_BLOCK_INTERVAL (usually 1000).
+                    blocknum * (BLOCK_OFFSET / TX_OFFSET) + txindex; 54 bits can represent all transactions for 85 years.
+     *   168 bits - The exit id
      */
-    function computePriority(uint64 exitableAt, PosLib.Position memory txPos, uint160 exitId)
+    function computePriority(uint64 exitableAt, PosLib.Position memory txPos, uint168 exitId)
```
thec00n, line 26: `exitableAt` can represent 32 bit timestamp. We should change the datatype everywhere to unsure static type safety.

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
thec00n: I am ok with that as well.

kevsul: `parseExitableAt()` is used in `ExitGameController`
 
https://github.com/omgnetwork/plasma-contracts/blob/30741a715e6acd9c95f2cc3cf4a8f77666f00950/plasma_framework/contracts/src/framework/ExitGameController.sol#L168

The 2 new parse functions aren't used, but I put them in here for consistency.

They are all available to clients in the `ExitPriorityWrapper`.

Not sure what the issue is - do you want to move just `parseExitId` and `parseTxPos` to `ExitPriorityWrapper`?

boolafish: > do you want to move just parseExitId and parseTxPos to ExitPriorityWrapper?

yeah. Not putting unused code in production as general good practice. Though I am not really strong on this. I will leave this to your decision.

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
kevsul: Dammit, you're right 🤦 
txpos should be 56 instead of 54

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
kevsul: Sure, another shift works as well as a mask. Do you think this is clearer/more efficient? 
```
uint256 pos = ((priority << SIZEOF_TIMESTAMP) >> SIZEOF_EXITID + SIZEOF_TIMESTAMP);
```

plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -2,68 +2,58 @@ pragma solidity 0.5.11;
...
-            return _computeStandardExitId(hashData, _utxoPos.outputIndex);
+        bytes32 outputId;
+        if (isDeposit) {
+            outputId = OutputId.computeDepositOutputId(txBytes, utxoPos.outputIndex, utxoPos.encode());
+        } else {
+            outputId = OutputId.computeNormalOutputId(txBytes, utxoPos.outputIndex);
         }
 
-        return _computeStandardExitId(keccak256(_txBytes), _utxoPos.outputIndex);
+        return uint168((uint256(outputId) >> (256 - FIRST_BIT_LOCATION)));
```
thec00n, line 48: Does this functionally still return the same exit id? Is this meant as a code clean up?

plasma_framework/contracts/src/utils/PosLib.sol
```diff
@@ -17,9 +17,10 @@ library PosLib {
...
     uint256 constant internal MAX_OUTPUT_INDEX = TX_OFFSET - 1;
-    // since we are using merkle tree of depth 16, max tx index size is 2^16 - 1
+    // Since we are using a merkle tree of depth 16, max tx index size is 2^16 - 1
     uint256 constant internal MAX_TX_INDEX = 2 ** 16 - 1;
-    // in ExitPriority, only 54 bits are reserved for both blockNum and txIndex
+    // The maximum block number we can handle depends on ExitPriority. ExitPriority only uses 54 bits for blockNum + txIndex
+    // This is 180143985094 blocks, which works out to be 85 years of blocks (given a block occurs every 15 seconds)
     uint256 constant internal MAX_BLOCK_NUM = ((2 ** 54 - 1) - MAX_TX_INDEX) / (BLOCK_OFFSET / TX_OFFSET);
 
     /**
```
thec00n, line 26: getTxPositionForExitPriority() should return uint56?

plasma_framework/contracts/src/framework/utils/ExitPriority.sol
thec00n: I think we use shift operations pretty consistently and i honestly can read them better :)

plasma_framework/test/src/framework/utils/ExitPriority.test.js
```diff
@@ -134,23 +136,94 @@ contract('ExitPriority', () => {
             expect(parsedExitableAt).to.be.bignumber.equal(new BN(exitableAt));
         });
 
-        it('should be able to parse the "exitableAt" from priority given max txPos of uint54', async () => {
+        it('should be able to parse the "exitableAt" from priority given max txPos of uint56', async () => {
             const exitId = (new BN(2)).pow(new BN(160)).sub(new BN(1)); // 2^160 - 1
```
thec00n, line 140: should 2^168?

plasma_framework/contracts/src/exits/utils/ExitId.sol
kevsul: Yes.
Previously `_computeStandardExitId()` was called twice.
Now that code is one line that's not repeated, so no need to split it out into a separate function

thec00n:
Could we get another set of eyes on this before merging?

thec00n:
Closes https://github.com/omgnetwork/plasma-contracts/issues/605

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -167,12 +166,12 @@ contract Liquidity is ERC721Full {
...
      * @param exitId The exit id
     */
-    function getWithdrawal(uint160 exitId) public {
+    function getWithdrawal(uint168 exitId) public {
         require(
             super.ownerOf(exitId) == msg.sender,
             "Only the NFT owner of the respective exit can get the withdrawal"
         );
-        uint160[] memory exitIdList = new uint160[](1);
+        uint168[] memory exitIdList = new uint168[](1);
```
pgebal, line 174: I know it's not part of your task, but that was a great opportunity to rename of this variable to `exitIds` :+)

plasma_framework/contracts/src/exits/utils/ExitId.sol
```diff
@@ -2,68 +2,58 @@ pragma solidity 0.5.11;
...
-     * @return _standardExitId Unique ID of the standard exit
+     * @param isDeposit Defines whether the tx for the exitId is a deposit tx
+     * @param txBytes Transaction bytes
+     * @param utxoPos UTXO position of the exiting output
+     * @return standardExitId Unique ID of the standard exit
      *     Anatomy of returned value, most significant bits first:
-     *     8-bits - output index
      *     1-bit - in-flight flag (0 for standard exit)
-     *     151-bits - hash(tx) or hash(tx|utxo) for deposit
+     *     167-bits - hash(tx) or hash(tx|utxo) for deposit
```
pgebal, line 30: It's  not  `hash(tx)` anymore. Now it's `outputId`.

Issue #632, opened 2020-05-21
Add sanity check that contract can deploy on Ropsten


boolafish:
##  Issue Type 
```
[ ] bug report
[x] feature request
```

## Current Behavior

From Devops' feedback, contract deployment via truffle migrate from time to time fails and requires manual retry. Would be nice to have automated sanity check testing the contract deployment on real network.

PS. currently it is only tested against local ganache, which the response time is fast and does not need to wait on the transaction to go through the network + mining....etc

## Motivation

Make devops happy

boolafish:
done, have the deployment to Rinkeby test

PR #637, opened 2020-05-29
feat: deploy to Rinkeby in ci

boolafish:
### Note
- update docker file and move it to outer layer of folder because we need to copy the `.git` file now (we need to parse git hash when deploying with semver)
- CI add check with deployment to Rinkeby with docker on master branch
- auto check CI faucet address balance. If not enough, submit GitHub issue to alarm people.
- add `GAS_PRICE` env var control on deployment.
- update README with the env vars.
- Sneak the rebrand change!

closes: https://github.com/omisego/devops/issues/242, https://github.com/omisego/plasma-contracts/issues/632
### Test
- contract deployment job: [here](https://app.circleci.com/pipelines/github/omisego/plasma-contracts/2158/workflows/fff12689-9be9-406e-9318-10bd8d983972/jobs/11241/steps)
- check balance job: [here](https://app.circleci.com/pipelines/github/omisego/plasma-contracts/2165/workflows/ce37349e-3008-4ca6-8bc3-b9357123efcc/jobs/11278)


.circleci/config.yml
```diff
@@ -171,7 +171,39 @@ jobs:
...
+            -e MAINTAINER_PRIVATEKEY=${FAUCET_PRIV_KEY} \
+            -e AUTHORITY_PRIVATEKEY=${FAUCET_PRIV_KEY} \
+            -e REMOTE_URL="https://rinkeby.infura.io/v3/${INFURA_API_TOKEN}" \
+            omisego/plasma-contracts \
+            npx truffle migrate --reset \
+            --network remote
+      - run:
+          name: check faucet balance
+          command: |
+            python .circleci/scripts/check_faucet_balance_and_warn.py
```
pik694, line 206: maybe let's move it up - even before building a docker image

plasma_framework/docs/deploying/deploying.md
```diff
@@ -12,6 +12,11 @@ Truffle migration scripts run in the order of the numbered prefix of the filenam
 - 2000: Future extensions, new exit games, etc.
 
 ## ENV VAR settings
+- DEPLOYER_PRIVATEKEY: Private key of the deployer. Required when deploying with `--network remote`.
```
pik694, line 15: `PRIVATEKEY` or `PRIVATE_KEY`?

.circleci/scripts/check_faucet_balance_and_warn.py
```diff
@@ -0,0 +1,63 @@
+import json
+import os
+from urllib import request, parse
+
+THRESHOLD = 3000000000000000000 # 3 ETH
+FAUCET_ADDRESS = '0x179E20BA056FF73A8161d3E3bbC1627CB12e88cC'
```
pik694, line 6: this should probably be configurable?

.circleci/config.yml
boolafish: I put it here so that after using some eth from the faucet account, we can make sure the remaining is still enough.

.circleci/scripts/check_faucet_balance_and_warn.py
boolafish: hmmm so you are proposing this to be as env var and set in CircleCI config instead?

plasma_framework/docs/deploying/deploying.md
boolafish: this is how it is used for now 🤷‍♂️. I agree I don't like the var name, but I don't want to introduce change now.
Or actually if DevOps deployment flow is not too much impacted we can change that.... cc. @jbunce @arthurk 

https://github.com/omgnetwork/plasma-contracts/blob/804a4f054bab7d46644e57ef0069b775fc645264/plasma_framework/truffle-config.js#L30

.circleci/scripts/check_faucet_balance_and_warn.py
pik694: yeah, probably

boolafish: late to come back but updated

PR #638, opened 2020-06-03
[fast_exits] Pay exit bond to exit initiator

souradeep-das:
Closes https://github.com/omgnetwork/plasma-contracts/issues/629

### DIFF
--- a/plasma_framework/contracts/poc/fast_exits/Liquidity.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Liquidity.sol
@@ -22,7 +22,6 @@ contract Liquidity is ERC721Full {
     struct ExitData {
         uint256 exitBondSize;
         address exitInitiator;
-        bool bondReturned;
         uint256 exitAmount;
     }
 
@@ -162,7 +161,7 @@ contract Liquidity is ERC721Full {
 
         FungibleTokenOutputModel.Output memory outputFromSecondTransaction
         = decodedSecondTx.outputs[0];
-        exitData[exitId] = ExitData(msg.value, msg.sender, false, outputFromSecondTransaction.amount);
+        exitData[exitId] = ExitData(msg.value, msg.sender, outputFromSecondTransaction.amount);
     }
 
     /**
@@ -174,18 +173,10 @@ contract Liquidity is ERC721Full {
             super.ownerOf(exitId) == msg.sender,
             "Only the NFT owner of the respective exit can get the withdrawal"
         );
-        uint160[] memory exitIdList = new uint160[](1);
-        exitIdList[0] = exitId;
-        PaymentExitDataModel.StandardExit[] memory exits = paymentExitGame.standardExits(
-            exitIdList
-        );
-        if (exits[0].utxoPos == 0) {
-            super._burn(msg.sender, exitId);
-            // possibly a separate function to pull the exit bond
-            msg.sender.transfer(exitData[exitId].exitAmount);
-        } else {
-            revert("Not processed exit");
-        }
+        
+        require(isExitProcessed(exitId), "Exit not Processed");
+        super._burn(msg.sender, exitId);
+        msg.sender.transfer(exitData[exitId].exitAmount);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -21,6 +21,8 @@ contract Liquidity is ERC721Full {
 
     struct ExitData {
         uint256 exitBondSize;
+        address exitInitiator;
+        bool bondReturned;
```
pik694, line 25: instead of having such a flag we could set `exitInitiator` to `address(0)`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -186,6 +188,28 @@ contract Liquidity is ERC721Full {
         }
     }
 
+    /**
+     * @dev Get Exit bond back - to be called by exit intitiator
+     * @param exitId The exit id
+    */
+    function getExitBond(uint160 exitId) public {
+        require(exitData[exitId].bondReturned == false, "Exit Bond Already returned");
+        require(exitData[exitId].exitInitiator != address(0), "ExitId is invalid");
```
pik694, line 197: I believe that this check is not necessary as the check blow is essential

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -186,6 +188,28 @@ contract Liquidity is ERC721Full {
...
+        require(msg.sender == exitData[exitId].exitInitiator, "Only the Exit Initiator can claim the bond");
+        
+        uint160[] memory exitIdList = new uint160[](1);
+        exitIdList[0] = exitId;
+        PaymentExitDataModel.StandardExit[] memory exits = paymentExitGame.standardExits(
+            exitIdList
+        );
+        if (exits[0].utxoPos == 0) {
```
pik694, line 205: maybe `assert` here. this would eliminate need to use `if else` structure

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: yes @pik694, but in that case we wouldn't be able to differentiate between a wrong/invalid exitId provided and an already claimed bond in case we wan't to throw distinct messages for the users? :thinking:

pik694: but should we need that? it is just an invalid claim

souradeep-das: Lets remove it if we dont need to give out the specific reason. Your call :)

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
souradeep-das: ohh if we use assert here, we cant return back an error message? Do we want that?

boolafish: > require in the middle might not be a good idea?

I don't think it matters. And `if-else + revert` is really just the same thing.

boolafish: ah...I know we are still handling the exit processed check in another issue.
However, can we put `exit.utxoPos == 0` into a function now (and the part in `getWithdraw` too) with naming like: (eg.) `checkExitProcessed` so make this a bit easier to read.

souradeep-das: >And if-else + revert is really just the same thing

Thats true! I'll change to require for better readability then
Also having a function here to `checkExitProcessed` will make it simpler to make changes later

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
```diff
@@ -250,10 +261,39 @@ contract(
 
                                 expect(actualLCBalanceAfterProcessExit).to.be.bignumber.equal(expectedLCBalance);
                             });
+                            describe('When Alice tries to claim the exit bond back', () => {
```
pik694, line 264: maybe let's add one additional test: alice starts an exit with exit bond of size 1, bond size changes, bob starts exit with new exit bond, etc...

souradeep-das: sure, noted. Merging this and will add that test in the next one?

pik694: is rather seems to be a quick change - if you have time, please add it now

souradeep-das: done `c5d4e11`

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -160,30 +161,48 @@ contract Liquidity is ERC721Full {
...
+        super._burn(msg.sender, exitId);
+        msg.sender.transfer(exitData[exitId].exitAmount);
+    }
+
+    /**
+     * @dev Get Exit bond back - to be called by exit intitiator
+     * @param exitId The exit id
+    */
+    function withdrawExitBond(uint160 exitId) public {
+        require(exitData[exitId].exitInitiator != address(0), "Exit Bond does not exist or is already claimed");
```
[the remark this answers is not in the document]
souradeep-das: Thanks @pik694, doing it on the next one.

PR #645, opened 2020-06-17
fix: let deployer deploy not maintainer

InoMurko:
https://github.com/omgnetwork/plasma-contracts/issues/578


InoMurko:
@boolafish @kevsul @thec00n https://github.com/omgnetwork/plasma-contracts/pull/645/files#diff-d5745788fcba9134f1624cfd7bfc37b4L31


are we okay with `spendingConditionRegistry.registerSpendingCondition` spending conditions being registered by the deployer?

boolafish:
It is a bit complex here, but the `deployer` would/should only have a short lived ownership for registering the spending condition before the `SpendingConditionRegistry` get injected to the exit game contract. As a result, we renounce the ownership right after all spending conditions are registered.

https://github.com/omgnetwork/plasma-contracts/blob/master/plasma_framework/migrations/99_renounce_spending_condition_registry.js#L12

We also check that the spending condition registry ownership is renounced at the time of exit game construction: https://github.com/omgnetwork/plasma-contracts/blob/master/plasma_framework/contracts/src/exits/payment/PaymentExitGame.sol#L30

thec00n:
I wonder if we could rewrite the migration scripts in such a way that that we don't need the Maintainer at all during deployment. We could look into moving relevant things into the constructor, do you guys think think this could be done?

boolafish:
I think we can at best move all the registration to later part of migration, or move it out of migration script and become sth standalone. but I am not 100% sure if there is any contract check that requires the maintainer to show up on contract deployment (I doubt there is or there should be, need to check).

InoMurko:
Alright, I think what you're saying is that my concerns are valid, but they're handled already!

What I'm doing right now is to move all maintainer operations into one migration file that'll work both with the secure enclave and without it (so that we keep development intact).

InoMurko:
Closed for https://github.com/omgnetwork/plasma-contracts/pull/649/files

PR #647, opened 2020-06-22
feat: emit an event indicating that in-flight exit has finalized

pgebal:
Closes https://github.com/omgnetwork/plasma-contracts/issues/646


plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
```diff
@@ -123,6 +127,8 @@ library PaymentProcessInFlightExit {
             if (!success) {
                 emit InFlightBondReturnFailed(exit.bondOwner, exit.bondSize);
             }
+
+            emit InFlightExitFinalized(exitId);
```
InoMurko, line 131: whoa! how was this one missed. cool we finally cought it

InoMurko:
@kevsul @souradeep-das should we do this too?

souradeep-das (approved):
Let's also generate solidoc before merge if we can?

plasma_framework/contracts/src/exits/payment/controllers/PaymentProcessInFlightExit.sol
souradeep-das: Can we swap L141 and L142 🙈 if we add another commit

PR #651, opened 2020-06-26
fix: truffle test failure

### LINKED ISSUE #652: fix truffle test failure
## Issue Type 

```
[x] bug report
[ ] feature request
```

## Current Behavior
CI is failing on truffle testing

boolafish:
Seems like there is some dependency trouble. Try blindly solving it with npm audit fix.

closes #652

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -235,8 +235,8 @@ workflows:
 
   Deployment checks:
     jobs:
-      - Deploy to Rinkeby:
-          filters:
-            branches:
-              only:
-                - master
+      - Deploy to Rinkeby
+          # filters:
+          #   branches:
+          #     only:
+          #       - master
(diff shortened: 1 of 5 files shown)


kevsul:
Quick update before I give up for now:
The problem is that sometimes calling `eth_getBlockByNumber` returns a null block (and no error).
It appears to only happen when the provider is `remote` i.e. using Infura

I think it's an Infura timing issue - we get a notification that a new block has been produced, but when we try to get it, it returns null.

I can "fix" it by hacking a `sleep` call into the Web3ProviderEngine code:
```
Web3ProviderEngine.prototype.start = function(cb = noop){
  const self = this

  // trigger start
  self._ready.go()

  // on new block, request block body and emit as events
  self._blockTracker.on('latest', async (blockNumber) => {

    await sleep(1000) // Come on Infura!

    // get block body
    self._getBlockByNumber(blockNumber, (err, block) => {
```

but obviously that's horrible.

I don't know why this is only showing up now though...

boolafish:
okay, seems like the new HDWalletProvider would fix this.

Issue #661, opened 2020-07-21
[Process Exit Bounty] Implement bounty for processing standard exits


souradeep-das:
https://github.com/omgnetwork/plasma-contracts/issues/607
## Issue Type 

```
[ ] bug report
[x] feature request
```

## High level Checklist

- Accept bounty when starting a standard exit
- Ensure size of bounty is correct
- Transfer bounty to challenger for challenged exits
- Transfer bounty to user for processing exits

PR #665, opened 2020-07-28
[Process Exit Bounty] Add process exit bounty for SE

souradeep-das:
#661 
#### Overview 
This PR adds the PE bounty for Standard Exits only
#### Out of scope for this PR
1. Make the bounty updatable
2. anti-Front running fix


plasma_framework/contracts/src/exits/utils/ExitBounty.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity 0.5.11;
+
+library ExitBounty {
+
+    /**
+     * @notice Returns the Process Exit Bounty size for standard exits
+     */
+    function processStandardExitBountySize() internal view returns (uint256) {
+        return 107000 * tx.gasprice;
```
pgebal, line 9: Why that formula? 
It would be nice if we had an explanation in the issue or in a comment here.

plasma_framework/test/endToEndTests/FastExit.e2e.test.js
```diff
@@ -192,7 +200,11 @@ contract(
                                 rlpDepositTx,
                                 depositInclusionProof,
                                 depositUtxoPos,
-                                { from: bob, value: this.startStandardExitBondSize },
+                                {
+                                    from: bob,
+                                    value: this.startStandardExitBondSize.add(this.processExitBountySize),
```
pgebal, line 205: This addition can be done in `setupContracts`, named `startStandardExitTxValue` or other, better name.

plasma_framework/test/endToEndTests/PaymentStandardExit.e2e.test.js
```diff
@@ -238,14 +261,18 @@ contract('PaymentExitGame - Standard Exit - End to End Tests', ([_deployer, _mai
 
                         this.bobBalanceBeforeProcessExit = new BN(await web3.eth.getBalance(bob));
 
-                        await this.framework.processExits(config.registerKeys.vaultId.eth, ETH, 0, 1);
+                        this.bobtx = await this.framework.processExits(config.registerKeys.vaultId.eth, ETH, 0, 1, {
```
pgebal, line 264: `bobTx` or `processTx`

souradeep-das:
The python tests could have worked without changes, ( by assuming the process exit bounty to have been 0), but i have modified them to account for the bounty while starting a standard exit. 
Please let me know if we should scrap the changes to the python tests, and move with the assumption of 0 bounty for them or if we feel the need to add any other py tests.

plasma_framework/contracts/poc/fast_exits/Liquidity.sol
```diff
@@ -9,6 +9,7 @@ import "../../src/framework/models/BlockModel.sol";
 import "../../src/utils/Merkle.sol";
 import "../../src/exits/payment/routers/PaymentStandardExitRouter.sol";
 import "openzeppelin-solidity/contracts/token/ERC721/ERC721Full.sol";
+import "../../src/exits/utils/ExitBounty.sol";
```
boolafish, line 12: minor: ordering of this import should be before openzeppelin ones

plasma_framework/contracts/src/exits/payment/PaymentExitGame.sol
```diff
@@ -41,10 +41,11 @@ contract PaymentExitGame is IExitProcessor, OnlyFromAddress, PaymentStandardExit
      * @notice Callback processes exit function for the PlasmaFramework to call
      * @param exitId The exit ID
      * @param token Token (ERC20 address or address(0) for ETH) of the exiting output
+     * @param processor The processExit initiator
      */
-    function processExit(uint168 exitId, uint256, address token) external onlyFrom(address(paymentExitGameArgs.framework)) {
+    function processExit(uint168 exitId, uint256, address token, address payable processor) external onlyFrom(address(paymentExitGameArgs.framework)) {
```
boolafish, line 46: is there a reason passing in this instead of using `msg.sender` to prevent auto frontrunning bot?

Also, do we want to check the processor is really the msg.sender or are we fine to be different people? If so, can we rename it to sth else, probably `exitProcessCompoentionTarget` (bear my bad English, I believe there is better wording)

plasma_framework/contracts/src/framework/ExitGameController.sol
```diff
@@ -170,7 +170,7 @@ contract ExitGameController is ExitGameRegistry {
             queue.delMin();
             processedNum++;
 
-            processor.processExit(exitId, vaultId, token);
+            processor.processExit(exitId, vaultId, token, msg.sender);
```
boolafish, line 173: hmm, in another comment I thought it is for front-running, but if it is so, the protection should be in this function instead right?

souradeep-das:
@boolafish Yes, I have the "make bounty updatable" and "anti- front running" on our board. Planning to do both of them together at the end after i land the IFE bounty PR

plasma_framework/contracts/src/exits/payment/PaymentExitGame.sol
souradeep-das: Oh so the processor here is taken to award him the bounty. Since there is an internal call from framework contract to the exit game contract, passing in the address of the caller and awarding the bounty. Thinking about doing the anti-front running at the end in a different task

pgebal:
> @boolafish Yes, I have the "make bounty updatable" and "anti- front running" on our board. Planning to do both of them together at the end after i land the IFE bounty PR

Do you have an idea how to prevent front running? I'm asking, because if we come up with it later it may require a very major refactor, that could for example revert this PR in a big way.

souradeep-das:
Mostly using the sender Data to prevent simple front-running by the bots was the idea. Similar to how we do for the challenge exits

pgebal:
> @pgebal The python tests could have worked without changes, ( by assuming the process exit bounty to have been 0), but i have modified them to account for the bounty while starting a standard exit.
> Please let me know if we should scrap the changes to the python tests, and move with the assumption of 0 bounty for them or if we feel the need to add any other py tests.( cc @boolafish )

I don't think we need more python tests.

pgebal:
@boolafish I think we should have a requirement for 2 reviewers to approve a PR before merging. I think from all our repos that's the one that deserves it the most. What do you think?

plasma_framework/contracts/src/exits/payment/PaymentExitGame.sol
boolafish: okay. now I rethink it might be better to explicit pass in args vs using `msg.sender` as a clearer dependency declaration and decoupling. So ignore the part why not `msg.sender`.

plasma_framework/contracts/src/exits/utils/ExitBounty.sol
```diff
@@ -0,0 +1,14 @@
...
+
+library ExitBounty {
+
+    /**
+     * @notice Returns the Process Exit Bounty size for standard exits
+     * @dev See https://github.com/omgnetwork/plasma-contracts/issues/658 for discussion about size
+     * 107000 is the approx gas usage for calling processExit()
+     */
+    function processStandardExitBountySize() internal view returns (uint256) {
+        return 107000 * tx.gasprice;
```
boolafish, line 11: - As for this implementation and design, it might be better to pass in gas price as args instead of using `tx.gasprice` for easier testing. I was actually surprise you can call a view function and pass in the gas price in the call to the Wrapper contract 😅 as it is not consuming any gas.
- However, if we are adding upgrade mechanism, I think it will be easier to be the same as what we do for bond. So we don't use `tx.gasprice` to do the calculation but hardcode one price there.
- Minor: slightly concern the abstract might break a bit as ExitBounty quite likely would need a storage when introducing upgrade mechanism. But would leave that as out of scope just a warning this seems to break in the future.

plasma_framework/contracts/mocks/exits/utils/ExitBountyWrapper.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity 0.5.11;
+
+import "../../../src/exits/utils/ExitBounty.sol";
+
+contract ExitBountyWrapper {
+
+    function processStandardExitBountySize() public view returns (uint256) {
```
boolafish, line 7: I think we need this function in the exit game contract directly. Users would need to know what bounty they should pass into. (Same as what we did for bond)

plasma_framework/python_tests/tests/tests_utils/plasma_framework.py
```diff
@@ -343,6 +343,9 @@ def childBlockInterval(self):
     def standardExitBond(self):
         return self.payment_exit_game.startStandardExitBondSize()
 
+    def processStandardExitBounty(self):
+        return 107000 * 100
```
boolafish, line 347: where does this 100 come from?? Also, I think we should make this the same as `standardExitBond` which gets data from contract directly

plasma_framework/test/src/exits/payment/controllers/PaymentProcessStandardExit.test.js
```diff
@@ -79,6 +94,7 @@ contract('PaymentProcessStandardExit', ([_, alice]) => {
             exitTarget,
             amount: web3.utils.toWei('3', 'ether'),
             bondSize: this.startStandardExitBondSize.toString(),
+            bountySize: this.processExitBountySize.toString(),
         });
 
         describe('when paying out bond fails', () => {
```
boolafish, line 100: I think we are missing the same test for bounty flow when the transfer fails

boolafish:
> I think we should have a requirement for 2 reviewers to approve a PR before merging. I think from all our repos that's the one that deserves it the most. What do you think?

lol okay....if other people feel like having more reviewers for approval I will change the setup then. Was thinking to loosen it for faster development speed for v2.

plasma_framework/contracts/src/exits/utils/ExitBounty.sol
souradeep-das: passing gas price in args will indeed be better 👍 . 
For the upgrade mechanism, i was thinking more in lines of updating the multiplier ["K"](https://github.com/omgnetwork/plasma-contracts/issues/658#issue-659410568) instead of a new value for the size, so we would not need to update the bounty size frequently and can accommodate frequent gas changes? What do you think?

pgebal:
> > I think we should have a requirement for 2 reviewers to approve a PR before merging. I think from all our repos that's the one that deserves it the most. What do you think?
> 
> lol okay....if other people feel like having more reviewers for approval I will change the setup then. Was thinking to loosen it for faster development speed for v2.

Maybe I'm wrong on that. That was just an idea. If you think 1 approval is enough, then let's stay with 1.

plasma_framework/contracts/mocks/exits/utils/ExitBountyWrapper.sol
souradeep-das: 👍 will take the gas price as an arg and add this func to exit game

plasma_framework/test/src/exits/payment/controllers/PaymentProcessStandardExit.test.js
souradeep-das: yess, thanks for catching

boolafish:
Just a reminder, we probably want to start a doc to track the breaking changes on V2.

I doubt you will implement this all the way to childchain, watcher, omg-js...etc so probably we should go back to @InoMurko's suggestion of documenting the integration changes.

PR #676, opened 2020-08-05
Inomurko/start vault

InoMurko:
splitting the PR https://github.com/omgnetwork/plasma-contracts/pull/649/files


MultiSigWalletOverride/docker/immutability/config/gencerts.sh
```diff
@@ -0,0 +1,92 @@
...
+EOF
+
+	openssl req \
+	-new \
+	-sha256 \
+	-newkey rsa:2048 \
+	-days 120 \
+	-nodes \
+	-x509 \
+	-subj "/C=US/ST=Maryland/L=Vault/O=My Company CA" \
```
souradeep-das, line 56: Use the same names as above

InoMurko: not sure I understand?

InoMurko: also, this is really not important for plasma-contracts specifically, vault just needs to standup and generate an authority account on demand. in an actual deployment this would be helm/kubernetess driven.

souradeep-das: True true. Just to ensure :)

MultiSigWalletOverride/docker/immutability/config/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+while [ "$count" -le 60 ]; do
+  if vault status; then break; fi
+  count=$((count+1))
+  sleep 0.5
+done
+
+vault status
+
+function configure_plugin {
+	plugin_file="immutability-eth-plugin"
```
boolafish, line 20: I know this is merged, but the space here seems off.

PR #678, opened 2020-08-05
Inomurko/vault+authority

InoMurko:


plasma_framework/truffle-config.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
             provider: () => new HDWalletProvider(
                 [
                     process.env.DEPLOYER_PRIVATEKEY || '0'.repeat(64),
-                    process.env.MAINTAINER_PRIVATEKEY || '0'.repeat(64),
+                    fs.readFileSync('../MultiSigWallet/build/multisig_instance') || process.env.MAINTAINER_PRIVATEKEY || '0'.repeat(64),
```
boolafish, line 27: I think this is misleading....you get deployed contract address from the file right (not private key...and there is no way to get a private key of the contract)?

This is one reason that is why I was suggesting a restructure of this comment: https://github.com/omgnetwork/plasma-contracts/pull/649#pullrequestreview-461371869

plasma_framework/migrations/1040_setup.js
```diff
@@ -0,0 +1,60 @@
...
+const config = require('../config.js');
+const pck = require('../package.json');
+
+module.exports = async (
+    deployer,
+    _,
+    // eslint-disable-next-line no-unused-vars
+    [deployerAddress, maintainerAddress, authorityAddress],
+) => {
+    const vault = process.env.VAULT || false;
```
boolafish, line 19: A discussion on `VAULT`. Since we have removed the activate childchain function, so there is actually no "vault" interaction anymore, right? Could we name the flag with better naming that does not confuse. Or is the Multisig contract considered as a form of vault too?

boolafish:
sorry not fully reviewed yet

plasma_framework/migrations/1040_setup.js
InoMurko: The idea is that you'd use the vault *with* multisig maintainer account, so they're both enabled/disabled under one flag. Perhaps if there's need, we could also have that optional...

- So you could use the vault, but no multisig maintainer.
- You could use multisig maintainer, but not vault.

Would that be better?

plasma_framework/truffle-config.js
InoMurko: yeah... I don't think this is used at all :D

InoMurko: lemme review what you said in #649

plasma_framework/migrations/1040_setup.js
boolafish: I am just wondering from the perspective of this repo, it does not make interaction to "vault" after we removed the activate childchain function. So it is not really a concept needed in this repo any more. So just wonder if there is better naming we can name this flag to...idk represent we are deploying with extra security care 😅

But yeah, I don't think we should complex this to 2 flag yet unless we see the need. I am just thinking if there can have a better naming for this concept but this comment could be minor. Not a blocker comment IMO

InoMurko: Not sure I understand fully ... but Vault is still used and interacted with: https://github.com/omgnetwork/plasma-contracts/pull/678/files#diff-6ea48dfd3d7c2965be57aa58597eb191R14

because it needs to create the wallet and authority account that's later used.

boolafish: oh! so actually we ... create the authority address via API during truffle migration?!
I misunderstood as we already have the authority account initiated/created beforehand.
I see

InoMurko: we could do that but then ... it needs to be automated somewhere so I thought It might as well be here :)

plasma_framework/migrations/1050_output_results.js
```diff
@@ -0,0 +1,52 @@
...
+const MerkleWrapper = artifacts.require('MerkleWrapper');
+const ERC20Mintable = artifacts.require('ERC20Mintable');
+module.exports = async (
+    deployer,
+    _,
+    // eslint-disable-next-line no-unused-vars
+    [deployerAddress, maintainerAddress, authorityAddress],
+) => {
+    const vault = process.env.VAULT || false;
+    if (!vault) {
```
souradeep-das, line 18: Maybe output doesn't show up for multisig?

.circleci/config.yml
```diff
@@ -106,6 +105,40 @@ jobs:
             npm install
             npx truffle migrate 0xa508dD875f10C33C52a8abb20E16fc68E981F186 1
           working_directory: ~/repo/MultiSigWallet
+      - run:
+          name: Install node 12
+          command: |
+            curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+            nvm install v12
+            node --version
+            node -v
```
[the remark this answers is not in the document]
InoMurko: I like em :D

plasma_framework/migrations/1050_output_results.js
InoMurko: yeah until the multisig actually executed all pending transactions we would have faulty outputs.

souradeep-das: Can we remove the vault check here? vault or no vault still just getting the deployed instances i mean 🤔

InoMurko: I think these three:
```
const ethVault = await plasmaFramework.vaults(config.registerKeys.vaultId.eth);
        const erc20Vault = await plasmaFramework.vaults(config.registerKeys.vaultId.erc20);
        const paymentExitGame = await plasmaFramework.exitGames(config.registerKeys.txTypes.payment);
```
would return faulty results since vaults aren't registered and exit games as well. is that what you meant?

souradeep-das: umm migration 1040 has register vault and exit game right?

souradeep-das: Maybe these values https://github.com/omgnetwork/plasma-contracts/pull/678/files#diff-a9446999f888af4dc205cab368614004R34

InoMurko:
```
1050_output_results.js
======================
{"authority_address":"0x7633fe8542c2218b5a25777477f63d395aa5afb4","eth_vault":"0x0000000000000000000000000000000000000000","erc20_vault":"0x0000000000000000000000000000000000000000","payment_exit_game":"0x0000000000000000000000000000000000000000","plasma_framework_tx_hash":"0x6fdc1e60ba970127239da7c3aa44121cd1a5f44e2a135434e686d44e38026c6b","plasma_framework":"0x2d697f6fb489326ecaabce7fc67678d3e8b838b1"}

- Saving migration to chain.
   > Saving migration to chain.
   -------------------------------------
   > Total cost:    
```
That's because https://github.com/omgnetwork/plasma-contracts/pull/678/files#diff-a9446999f888af4dc205cab368614004R54-R56 only puts the transactions into the multisig queue, not execute them. So techically they're waiting for the multisignature accounts to approve and call execute!

InoMurko: I mean yeah, I can leave it as is:
```
"eth_vault":"0x0000000000000000000000000000000000000000"
"erc20_vault":"0x0000000000000000000000000000000000000000"
"payment_exit_game":"0x0000000000000000000000000000000000000000"
```

Doesn't harm anyone 👍

souradeep-das: Aah i see. Let's see what we get here, if there is a problem we could probably just have whatever addresses are there logged? wdyt?

souradeep-das: awesome! we got what you said! 🚀

plasma_framework/truffle-config.js
InoMurko: okay... https://github.com/omgnetwork/plasma-contracts/pull/649#pullrequestreview-461371869
to regards to your comment. The comment I think isn't really related to the issue at hand and I'm not sure how it's a better approach regardless. It seems (the current way) is a fairly standard way how people provide accounts to truffle migration. Having them in config doesn't really change anything much. Or am I not understanding something?

InoMurko:
As a separate step I'll write a script that interacts with the gnosis multisig and executes all transactions.

MultiSigWalletOverride/truffle.js
```diff
@@ -22,6 +24,14 @@ module.exports = {
     remote: {
       skipDryRun: true,
       gasPrice: process.env.GAS_PRICE || 20000000000, // default 20 gwei
+      skipDryRun: true,
```
souradeep-das, line 27: extra :D

plasma_framework/truffle-config.js
boolafish: > and I'm not sure how it's a better approach regardless.

you do not need such `fs.readFileSync('../MultiSigWallet/build/multisig_instance')` thing (well...you do not need this line here too tbh) also all the [branching of getting vault address](https://github.com/omgnetwork/plasma-contracts/blob/bbdb558b4f61c54e34a7d8342c412ed45c13f3ec/plasma_framework/migrations/20_deploy_plasma_framework.js#L11-L17) could be hidden to a single place, reverting the dependency control of vault address to higher level of truffle migration script. In this case we have 2 options, `truffle-config.js` or `config.js`. I am not sure how to do that via `truffle-config.js` so I proposed `config.js`

> is a fairly standard way 

true, but current standard tooling (hd-wallet-provider) does not support what you need. That is what causing the complexity here.

hmmm...however, you have point me to a direction that it might be possible to just implement a new provider that would link with different way to "sending/approving transaction". Did a bit research and have not find enough documentation on this yet. So just throwing out idea at this moment. I will side track this direction differently and can/should be out of your effort of delivering this (unless it becomes obvious and it is super easy to do)

.circleci/config.yml
```diff
@@ -104,8 +103,41 @@ jobs:
...
+            node --version
+            npm install
+      - save_cache:
+          key: dependencies-machine-v1-{{ checksum "~/repo/plasma_framework/package-lock.json" }}
+          paths:
+            - ~/repo/plasma_framework/node_modules
+      - run:
+          name: Deploy Plasma Contracts
+          command: |
+            npx truffle migrate --network local
```
boolafish, line 133: I think DevOps has been asking for a docker to deploy things. We have a Dockerfile in the repo for this exact purpose and is tested on rinkeyby for each master merging (see the job `Deploy to Rinkeby`) though I don't think they have leveraged that as it is a bit new. Given our code would actually fail on random node version, I think it makes sense to support them with docker for a more stable deployer tool env setup.

Can we add docker support for them? So the ideal situation is this line is run inside a docker with the essential dependency installed already. Or we extend the existing `Deploy to Rinkeby` job with vault deployment.

MultiSigWalletOverride/migrations/2_deploy_contracts.js
```diff
@@ -4,25 +4,33 @@ const MultisigWalletFactory = artifacts.require('MultiSigWalletWithDailyLimitFac
...
+      accountsIndex = j;
+    }
+    if (process.argv[j] == '--confirmations') {
+      confirmationsIndex = j;
+    }
+  }
+  console.log(`Accounts: ${args[accountsIndex + 1].split(",")}`);
+  console.log(`Confirmations: ${args[confirmationsIndex + 1]}`);
+  if (accountsIndex === undefined || confirmationsIndex === undefined) {
+    console.log('ABORT. Use: --accounts 0xasdf,0xfdsa --confirmations 2');
```
boolafish, line 23: not sure I get what this means? Should the user retry with passing the account and confirmation using the data in this console.log? Also, might consider `console.error` instead
`Use: --accounts 0xasdf,0xfdsa --confirmations 2`

MultiSigWalletOverride/migrations/2_deploy_contracts.js
```diff
@@ -4,25 +4,33 @@ const MultisigWalletFactory = artifacts.require('MultiSigWalletWithDailyLimitFac
...
   const args = process.argv.slice();
-  if (process.env.DEPLOY_FACTORY){
-    deployer.deploy(MultisigWalletFactory);
-    console.log("Factory with Daily Limit deployed");
-  } else if (args.length < 5) {
-    console.error("Multisig with daily limit requires to pass owner " +
-      "list, required confirmations and daily limit");
-  } else if (args.length < 6) {
-    deployer.deploy(MultisigWalletWithoutDailyLimit, args[3].split(","), args[4]).then(function() {
+  let accountsIndex = undefined;
```
boolafish, line 9: probably you can use `indexOf` function here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

MultiSigWalletOverride/migrations/2_deploy_contracts.js
```diff
@@ -4,25 +4,33 @@ const MultisigWalletFactory = artifacts.require('MultiSigWalletWithDailyLimitFac
...
+  for (let j = 0; j < process.argv.length; j++) {
+    
+    if (process.argv[j] == '--accounts') {
+      accountsIndex = j;
+    }
+    if (process.argv[j] == '--confirmations') {
+      confirmationsIndex = j;
+    }
+  }
+  console.log(`Accounts: ${args[accountsIndex + 1].split(",")}`);
```
boolafish, line 20: minor: can we put `args[accountsIndex + 1].split(",")` into a variable as this is used afterward again to be more readable? same as the confirmation one.

MultiSigWalletOverride/migrations/2_deploy_contracts.js
InoMurko: It just explains how the migrations should be run, because I've overriden the way gnosis does it. Ill use console.error, tnx!

.circleci/config.yml
InoMurko: yeah we can def. do that. I'll do that in the next PR so that I don't make this too big.

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,171 @@
...
+        web3.eth.sendTransaction({ gas: 3000000, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier }, async function (error, transactonHash) {
+            console.log(`Submitted transaction with hash: ${transactonHash}`);
+            let transactionReceipt = null;
+            while (transactionReceipt == null) { // Waiting expectedBlockTime until the transaction is mined
+                /* eslint-disable no-await-in-loop */
+                transactionReceipt = await web3.eth.getTransactionReceipt(transactonHash);
+                /* eslint-disable no-await-in-loop */
+                await sleep(expectedBlockTime);
+            }
+            console.log(`Got the transaction receipt for ETH setDepositVerifier: ${transactionReceipt}`);
```
souradeep-das, line 69: looks like we wait for the receipt but it isn't logged at the end 🤔 Does anything silently fail?

souradeep-das:
Should we also enable the `Deploy to Rinkeby` workflow?

plasma_framework/migrations/1041_setup_vault.js
InoMurko: I'm not sure I understand. What isn't logged? Ah, do you think I should look at status from the transactionReceipt and abort if it's not true?

souradeep-das: Ohh the transactionReceipt body's aren't logged i think. Yeah I was also thinking about looking at the status and handling if a failed transaction, though not a necessity maybe but might be good to have

souradeep-das: Probably just logging out transaction receipts to show which failed or aborting

InoMurko: I think I'll abort. Because we insert 8 transactions into the gnosis multisig, all of them need to be executed "in order" they were submitted. So it's easier to know where things broke and where to continue.

InoMurko:
> Should we also enable the `Deploy to Rinkeby` workflow?

So that we deploy twice? Once with vault and the second time without?

souradeep-das:
yeah 👍

InoMurko:
I've just noticed `Deploy to Rinkeby` uses a docker approach. So I'll do that in the subsequent PR where I actually install all dependencies that we need to support vault and gnosis stuff.

.gitignore
```diff
@@ -51,3 +51,4 @@ tenderly.yaml
 
 MultiSigWallet
 MultiSigWalletOverride/ganache_data/
+MultiSigWalletOverride/docker/immutability/ca/
```
boolafish, line 54: really minor: GH dislike file without extra line 🤷‍♂️ always show up on review page

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,179 @@
...
+        const gnosisSetDepositVerifier = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [ethVault.address, 0, setDepositVerifier]);
+        let transaction = await web3.eth.sendTransaction({ gas: 3000000, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
+        console.log(`Submitted transaction with hash for ETH setDepositVerifier: ${transaction.transactionHash}`);
+        let transactionReceipt = null;
+        while (transactionReceipt == null) { // Waiting expectedBlockTime until the transaction is mined
+            transactionReceipt = await web3.eth.getTransactionReceipt(transaction.transactionHash);
+            if (transactionReceipt != null && transactionReceipt.status === true) {
+                console.log('Got a success transaction receipt for ETH setDepositVerifier');
+            } else {
+                await sleep(expectedBlockTime);
```
boolafish, line 69: probably at least add a log or print sth before sleeping so we would know what we are waiting and the code is still running if it gets slow....same for other places

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,179 @@
...
+            ],
+            name: 'submitTransaction',
+            outputs: [{ name: 'transactionId', type: 'uint256' }],
+            payable: false,
+            type: 'function',
+            signature: '0xc6427474' };
+        // ethVault.setDepositVerifier
+        const setDepositVerifier = web3.eth.abi.encodeFunctionCall(ethVault.abi.find(o => o.name === 'setDepositVerifier'), [ethDepositVerifier.address]);
+        const gnosisSetDepositVerifier = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [ethVault.address, 0, setDepositVerifier]);
+        let transaction = await web3.eth.sendTransaction({ gas: 3000000, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
```
boolafish, line 61: omg, it is so great it is not callback now compare to the first version 🙇‍♂️ it was hard

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,179 @@
...
+        console.log(`Submitted transaction with hash for ETH setDepositVerifier: ${transaction.transactionHash}`);
+        let transactionReceipt = null;
+        while (transactionReceipt == null) { // Waiting expectedBlockTime until the transaction is mined
```
boolafish, line 64: I believe js prefer `===` instead of `==` for stricter comparison though I forgot what exactly the difference is 😅 But in general it is better to forgot `==` exists.

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,179 @@
...
+            transactionReceipt = await web3.eth.getTransactionReceipt(transaction.transactionHash);
+            if (transactionReceipt != null && transactionReceipt.status === true) {
+                console.log('Got a success transaction receipt for ETH setDepositVerifier');
+            } else {
+                await sleep(expectedBlockTime);
+            }
+        }
+        console.log(`Transaction receipt for ETH setDepositVerifier: ${util.inspect(transactionReceipt, { showHidden: false, depth: null })}`);
+        // plasmaFramework.registerVault
```
boolafish, line 73: minor: I would appreciate if you could break these steps into smaller functions and call it here for better readability

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,179 @@
...
+        const registerVault = web3.eth.abi.encodeFunctionCall(plasmaFramework.abi.find(o => o.name === 'registerVault'), [config.registerKeys.vaultId.eth, ethVault.address]);
+        const gnosisRegisterVault = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [plasmaFramework.address, 0, registerVault]);
+        transaction = await web3.eth.sendTransaction({ gas: 3000000, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisRegisterVault });
+        console.log(`Submitted transaction with hash for ETH registerVault: ${transaction.transactionHash}`);
+        transactionReceipt = null;
+        while (transactionReceipt == null) { // Waiting expectedBlockTime until the transaction is mined
+            transactionReceipt = await web3.eth.getTransactionReceipt(transaction.transactionHash);
+            if (transactionReceipt != null && transactionReceipt.status === true) {
```
boolafish, line 81: same here `!==` is preferred. Apply to all the  other `if` inside `while` I think

MultiSigWalletOverride/migrations/2_deploy_contracts.js
```diff
@@ -4,25 +4,25 @@ const MultisigWalletFactory = artifacts.require('MultiSigWalletWithDailyLimitFac
...
+  const confirmationsIndex = args.indexOf('--confirmations');
+  const numberOfConfirmations = args[confirmationsIndex + 1];
+  const accounts = args[accountsIndex + 1].split(",");
+  if (accountsIndex === -1 || confirmationsIndex === -1) {
+    console.error('ABORTED. Use: --accounts 0xasdf,0xfdsa --confirmations 2');
+    process.exit(1); 
+  } else {
+    console.log(`Accounts: ${accounts}`);
+    console.log(`Confirmations: ${numberOfConfirmations}`);
+    deployer.deploy(MultisigWalletWithoutDailyLimit, accounts, numberOfConfirmations).then(function() {
```
kevsul, line 19: Minor: Using `await` here might be better / easier to read.
As it is, if the deploy fails you'll still get a `Wallet deployed` log, which could be confusing.

Either use `await` or move the `console.log("Wallet deployed");` into the `then` block

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,187 @@
...
+        const ethDepositVerifier = await EthDepositVerifier.deployed();
+        const ethVault = await EthVault.deployed();
+        const erc20DepositVerifier = await Erc20DepositVerifier.deployed();
+        const erc20Vault = await Erc20Vault.deployed();
+        const MORE_VP = config.frameworks.protocols.moreVp;
+        const PAYMENT_TX_TYPE = config.registerKeys.txTypes.payment;
+        const FEE_TX_TYPE = config.registerKeys.txTypes.fee;
+        const paymentExitGame = await PaymentExitGame.deployed();
+        const feeExitGame = await FeeExitGame.deployed();
+        const buildDir = path.resolve(__dirname, '../../MultiSigWallet/build/multisig_instance');
```
kevsul, line 63: Minor. `buildDir` doesn't seem the right name for this variable.

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -0,0 +1,187 @@
...
+const EthDepositVerifier = artifacts.require('EthDepositVerifier');
+const Erc20DepositVerifier = artifacts.require('Erc20DepositVerifier');
+const util = require('util');
+const childProcess = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const config = require('../config.js');
+const pck = require('../package.json');
+
+const expectedBlockTime = 1000;
```
kevsul, line 19: expectedBlockTime is one second?

plasma_framework/migrations/1041_setup_vault.js
kevsul: I agree. There seems to be quite a bit of duplicated code that could be extracted

plasma_framework/migrations/1041_setup_vault.js
InoMurko: it's more like how frequently we're willing to pull for results

kevsul (approved):

PR #679, opened 2020-08-05
rem child chain activation

InoMurko:
some background: https://github.com/omgnetwork/plasma-contracts/pull/649#discussion_r465524336


boolafish:
cc @thec00n to add to your list of review when you're back. You can probably scan everything again as a whole instead to see if we are missing anything.

PR #686, opened 2020-08-19
Enrich contract events to contain data needed by child-chain and watcher

pgebal:
Update data stored in events so it contains everything needed by child-chain and watcher, so they do not have to get if from call data.

Motivation for the changes are:
- we can't rely on call data
- simplicity in elixir services

Changes come with a cost of increased gas.

References https://github.com/omgnetwork/plasma-contracts/issues/659

Helpful link:
https://omisego.atlassian.net/wiki/spaces/FT/pages/667746350/Minimal+Contract+Events+Review


kevsul (approved):

plasma_framework/package.json
```diff
@@ -14,10 +14,11 @@
     "solidoc": "^1.0.5"
   },
   "devDependencies": {
-    "elliptic": ">=6.5.3",
     "@codechecks/client": "^0.1.10",
+    "@openzeppelin/test-helpers": "^0.5.6",
```
souradeep-das, line 18: The older version of the test helpers package could be uninstalled

thec00n (commented):
@pgebal the tests do not pass I would like to see a gas report and compare the impact

pgebal:
@kevsul please take a look at this commit: https://github.com/omgnetwork/plasma-contracts/pull/686/commits/f31f72b078971fd3e53827e9f8089523b6af224f

I deleted that check because web3py can't handle more complex event (I guess it's `bytes` field). I tried to update web3py to see if newer version is better but there is a lot of dependencies to untangle. I can get back to that in the last week of the cycle to not block developments in elixir-omg now.

Removing this check does not invalidates the python test.
We have e2e and unit truffle tests that check if a proper event is emitted.

pgebal:
> @pgebal the tests do not pass I would like to see a gas report and compare the impact

yes, I've been playing with it, fixing now

plasma_framework/package.json
pgebal: Thanks, I decided to stay with the old version and remove the new one. I noticed I didn't use the new version.

kevsul:
> @kevsul please take a look at this commit: [f31f72b](https://github.com/omgnetwork/plasma-contracts/commit/f31f72b078971fd3e53827e9f8089523b6af224f)
> 
> I deleted that check because web3py can't handle complex events (I guess because of `bytes` field). I tried to update web3py hoping that newer version handles that well, but there is a lot of dependencies to untangle. I can get back to that in the last week of the cycle to not block developments in elixir-omg now.
> 
> Removing this check does not invalidate the python test.
> We have e2e and unit truffle tests that check if a proper event is emitted.

I think that's fine.

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartStandardExit.sol
```diff
@@ -43,7 +43,9 @@ library PaymentStartStandardExit {
 
     event ExitStarted(
         address indexed owner,
-        uint168 exitId
+        uint168 exitId,
+        uint256 utxoPos,
+        bytes rlpOutputTx
```
InoMurko, line 48: hmmm, do we need `rlpOutputTx `? I don't think we do. I thin it's only `utxoPos `

InoMurko: https://github.com/omgnetwork/elixir-omg/blob/306246168240d90f9f4cad63092a8f5ea243ea57/apps/omg_watcher/lib/omg_watcher/exit_processor.ex#L717

InoMurko: `rlpOutputTx` gets renamed into:
```
{"outputTxInclusionProof", :output_tx_inclusion_proof},
      {"rlpOutputTx", :output_tx},
      {"utxoPos", :utxo_pos}
```

so it's `output_tx`.

pgebal: We need it in our current code @Ino. Check `fields.ex` and exit processor core. We could work around that by getting the tx from rocksdb storage.

InoMurko: yeah fileds is a mapping module, it's not really *used* anywhere.

plasma_framework/package.json
souradeep-das: @pgebal Can we do a clean uninstall then? looks like it's still there in package-lock

plasma_framework/contracts/src/exits/payment/controllers/PaymentStartStandardExit.sol
pgebal: @InoMurko how about here:   https://github.com/omgnetwork/elixir-omg/blob/a7b4188c938390b13d069f40d941507ca79faca3/apps/omg_watcher/lib/omg_watcher/exit_processor/exit_info.ex#L66
Isn't it the exiting tx bytes from call data?
Called here: https://github.com/omgnetwork/elixir-omg/blob/a7b4188c938390b13d069f40d941507ca79faca3/apps/omg_watcher/lib/omg_watcher/exit_processor/core.ex#L188

pgebal:
During a call with @thec00n it was decided to remove: `owner` and `rlpOutputTx` fields from `ExitStarted` event and `inputTxs` from `InFlightExitStarted`.
I decided to keep `owner` field because it was an indexed field and may be beneficial for a user to track her own exits. We have analogous fields in some other exit events: `InFlightExitStarted`, piggyback events, etc.
@thec00n is it ok for you if we leave `owner` field? And please do a CR :+)

thec00n (approved):
I discussed this with @pgebal and we covered my concerns in terms of the increased gas usage. In general I think we should plan for targeted feature team working reducing gas costs before v2 goes live.

PR #690, opened 2020-08-19
support docker deploy for gnosis and vault

InoMurko:


kevsul (approved):

.circleci/config.yml
```diff
@@ -260,11 +328,21 @@ jobs:
             pyenv versions
             pyenv global 3.7.0
       - run:
-          name: build docker
+          name: Pull Submodules
```
souradeep-das, line 331: Why are we pulling submodules here? Just an idea, If it is because of the dockerfile, will using conditional logic in dockerfile be better or clumsy?

.circleci/config.yml
```diff
@@ -315,3 +393,9 @@ workflows:
             branches:
               only:
                 - master
+      - Deploy to Rinkeby with Gnosis and Vault:
+          filters:
+            branches:
+              only:
+                - master
```
souradeep-das, line 400: Since `v2.0.0` is the default branch, should we check deployments in the branch too

.circleci/config.yml
InoMurko: yeah so I decided not to complicate things but yeah sure I can add that!

.circleci/config.yml
InoMurko: uh uh, forgot about that!

.circleci/config.yml
souradeep-das: makes sense, lets keep this open for now to get another opinion

InoMurko: sorted!

souradeep-das (approved):
👍

PR #704, opened 2020-09-17
bump Vault

InoMurko:
Bumping the Vault version to the latest release.


.circleci/config.yml
```diff
@@ -96,12 +105,11 @@ jobs:
           name: Apply Overrides
           command: make init_multisig
           working_directory: ~/repo/MultiSigWalletOverride
+      # node symlink missing if there's not nodejs-legacy
+      - run: sudo apt-get update && sudo apt-get install -y build-essential libhidapi-dev libudev-dev libusb-1.0-0 libusb-1.0-0-dev nodejs-legacy
```
boolafish, line 109: what's happening to the original nvm installation?

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -40,45 +40,55 @@ const waitForReceipt = async (whatLog, transaction) => {
...
     const transaction = await web3.eth.sendTransaction({ gas: gas, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisFeeRegisterExitGame });
     console.log(`Submitted transaction with hash for ${whatLog}: ${transaction.transactionHash}`);
+    console.log(`Full log for ${whatLog}: ${transaction}`);
     await waitForReceipt(whatLog, transaction);
 };
 
 const setVersion = async (whatLog, plasmaFramework, sha, gnosisMultisigAbi, gnosisMultisigAddress, deployerAddress) => {
     const setVersionCall = web3.eth.abi.encodeFunctionCall(plasmaFramework.abi.find(o => o.name === 'setVersion'), [`${pck.version}+${sha}`]);
     const gnosisSetVersion = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [plasmaFramework.address, 0, setVersionCall]);
     const gas = await web3.eth.estimateGas({ to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetVersion });
```
boolafish, line 87: do we not need `*2` for this one?

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -40,45 +40,55 @@ const waitForReceipt = async (whatLog, transaction) => {
 const setDepositVerifier = async (whatLog, ethVault, ethDepositVerifier, gnosisMultisigAbi, gnosisMultisigAddress, deployerAddress) => {
     const setDepositVerifierCall = web3.eth.abi.encodeFunctionCall(ethVault.abi.find(o => o.name === 'setDepositVerifier'), [ethDepositVerifier.address]);
     const gnosisSetDepositVerifier = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [ethVault.address, 0, setDepositVerifierCall]);
-    const gas = await web3.eth.estimateGas({ to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
+    const gas = await web3.eth.estimateGas({ to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier }) * 2;
```
boolafish, line 43: nitpick: the `await web3.eth.estimateGas()` part is quite long, reading wise the `*2` is not that obvious at my first glance. Might be slightly clearer with the following

```js
const estimateGas = await web3.eth.estimateGas({ to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier })

const transaction = await web3.eth.sendTransaction({ gas: estimateGas * 2, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
```

MultiSigWalletOverride/docker/immutability/config/entrypoint.sh
```diff
@@ -1,20 +1,123 @@
...
+	-out "$TLS_CERT"
+
+	openssl x509 -in "$TLS_CERT" -noout -text
+
+	# rm openssl.cnf
+
+#  chown -R nobody:nobody $CONFIG_DIR && chmod -R 777 $CONFIG_DIR
+}
+
+gencerts
```
boolafish, line 109: I am not a script expert so I don't know what should be best practice, but shall we run commands after all function declaration so we don't mix reading command and function declaration together ??

plasma_framework/migrations/1041_setup_vault.js
InoMurko: makes sense!

also,

all of this is really not optimal at the moment, but it onblocks me on some other things. I will come back to this and make some adjustments - (like to be able to retry a certain transaction by maintaining some sort of state (like truffle migration does)). just FYI.

MultiSigWalletOverride/docker/immutability/config/entrypoint.sh
InoMurko: this is really just a copy/clone of *what comes with vault*. and it's purpose is to be able to standup and test the "migration integration" really works!

.circleci/config.yml
InoMurko: the intial problem was that ubuntu 14 (which is what cirleci uses if you don't define an image) uses a really old gcloud and docker and docker compose. that didn't allow me to authenticate towards our GCS registry. so I had to bump to ubuntu 16. `image: ubuntu-1604:201903-01`
now that broke helluva lot of things with gnosis 1.6 tag so I had to install `nodejs-legacy` to even be able to `npm install`.

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -41,44 +41,54 @@ const setDepositVerifier = async (whatLog, ethVault, ethDepositVerifier, gnosisM
     const setDepositVerifierCall = web3.eth.abi.encodeFunctionCall(ethVault.abi.find(o => o.name === 'setDepositVerifier'), [ethDepositVerifier.address]);
     const gnosisSetDepositVerifier = web3.eth.abi.encodeFunctionCall(gnosisMultisigAbi, [ethVault.address, 0, setDepositVerifierCall]);
     const gas = await web3.eth.estimateGas({ to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
-    const transaction = await web3.eth.sendTransaction({ gas: gas, to: gnosisMultisigAddress, from: deployerAddress, data: gnosisSetDepositVerifier });
+    console.log(`The amount of gas used for ${whatLog}: ${gas}`);
```
souradeep-das, line 44: Can we log the correct value of gas sent here too ( *2 ) and for all the logs

MultiSigWalletOverride/docker/immutability/config/entrypoint.sh
```diff
@@ -26,26 +129,43 @@ function configure_plugin {
...
+    vault audit enable file file_path=stdout
+    vault status
+    vault secrets list
+    test_banner
+fi
 
 # Don't exit until vault dies
-wait $VAULT_PID
+
+wait $VAULT_PID
```
souradeep-das, line 171: a newline if you want

souradeep-das (approved):

plasma_framework/migrations/1041_setup_vault.js
InoMurko: not sure, but at the moment even if it's reverted, there's no impact. the version I'll build after this PR will handle all these cases.

plasma_framework/migrations/1041_setup_vault.js
InoMurko: uuuuups!!!

boolafish (approved):

PR #716, opened 2020-11-12
Fix plasma if/else branching.

InoMurko:


souradeep-das (approved):
😅

plasma_framework/migrations/1040_setup.js
```diff
@@ -17,7 +17,7 @@ module.exports = async (
     [deployerAddress, maintainerAddress, authorityAddress],
 ) => {
     const vault = process.env.VAULT || false;
-    if (!vault) {
+    if (vault === false) {
```
souradeep-das, line 20: wait shouldn't this be if (vault !== true)

kevsul:
Good catch @InoMurko but I don't think the fix does what we want.
Because `VAULT=false` in the env will be converted to a string `"false"`, which will evaluate as true.

I believe the usual way to deal with booleans in env vars is:

const vault = process.env.VAULT === 'true';

That way `VAULT=true` evaluates to  `true` and anything else evaluates to `false`
It's not pretty, but I think it's as good as it gets...

plasma_framework/migrations/1040_setup.js
kevsul: That doesn't work either, because if you set `VAULT=true` in the env, javascript `vault !== true` resolves to true.
Which is kind of unexpected...

souradeep-das: omg yes 🤦

plasma_framework/migrations/1041_setup_vault.js
```diff
@@ -103,8 +103,8 @@ module.exports = async (
     // eslint-disable-next-line no-unused-vars
     [deployerAddress, maintainerAddress, authorityAddress],
 ) => {
-    const vault = process.env.VAULT || false;
-    if (vault) {
+    const vault = process.env.VAULT === 'true';
+    if (vault === true) {
```
kevsul, line 107: Nitpick: I prefer the way it was.
`vault` is definitely a boolean now, and IMO `if (vault)` is better than `if (vault === true)`
It's not a strong opinion though, so leave it if you think `if (vault === true` is clearer

kevsul (approved):

PR #723, opened 2020-12-02
Quasar Fast Exit: Implement withdrawal the happy path

souradeep-das:
### Overview

This includes-

1. Happy Withdrawal path with challenges
2. Updating Quasar Parameters

Yet to implement-

1. Byzantine condition and IFE-Claims
2. ERC-20 support
3. Anti front-running and service provider fee

### DIFF
--- a/plasma_framework/contracts/poc/fast_exits/Quasar.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Quasar.sol
@@ -95,7 +95,7 @@ contract Quasar {
         uint256 tokenAmount = ticketData[utxoPos].reservedAmount;
         ticketData[utxoPos].reservedAmount = 0;
         ticketData[utxoPos].validityTimestamp = 0;
-        tokenUsableCapacity[ticketData[utxoPos].token] += tokenAmount;
+        tokenUsableCapacity[ticketData[utxoPos].token] = tokenUsableCapacity[ticketData[utxoPos].token].add(tokenAmount);
         bondReserve = bondReserve.add(ticketData[utxoPos].bondValue); 
     }
 
@@ -103,7 +103,7 @@ contract Quasar {
      * @dev Add Eth Liquid funds to the quasar
     */
     function addEthCapacity() public payable onlyQuasarMaintainer() {
-        tokenUsableCapacity[address(0x0)] += msg.value;
+        tokenUsableCapacity[address(0x0)] = tokenUsableCapacity[address(0x0)].add(msg.value);
         emit QuasarTotalEthCapacityUpdated(tokenUsableCapacity[address(0x0)]);
     }
 
@@ -122,6 +122,7 @@ contract Quasar {
             uint256 residualAlmount = amount.sub(bondReserve);
             bondReserve = 0;
             tokenUsableCapacity[token] = tokenUsableCapacity[token].sub(residualAlmount);
+            emit QuasarTotalEthCapacityUpdated(tokenUsableCapacity[address(0x0)]);
         }
         msg.sender.transfer(amount);
     }
@@ -160,8 +161,8 @@ contract Quasar {
         require(outputData.amount != 0, "The reserved amount cannot be zero");
         require(msg.value == bondValue, "Bond Value incorrect");
 
-        tokenUsableCapacity[outputData.token] -= outputData.amount;
-        ticketData[utxoPos] = Ticket(msg.sender, block.timestamp + 14400, outputData.amount, outputData.token, msg.value, false);
+        tokenUsableCapacity[outputData.token] = tokenUsableCapacity[outputData.token].sub(outputData.amount);
+        ticketData[utxoPos] = Ticket(msg.sender, block.timestamp.add(14400), outputData.amount, outputData.token, msg.value, false);
     }
 
     // for simplicity fee has to be from a seperate input in the tx to quasar
@@ -209,7 +210,7 @@ contract Quasar {
         require(ticketData[utxoPos].reservedAmount == outputData.amount, "Wrong Amount Sent to Quasar Owner");
         require(ticketData[utxoPos].token == outputData.token, "Wrong Token Sent to Quasar Owner");
 
-        claimData[utxoPos] = Claim(rlpTxToQuasarOwner, block.timestamp + waitingPeriod, true);
+        claimData[utxoPos] = Claim(rlpTxToQuasarOwner, block.timestamp.add(waitingPeriod), true);
         ticketData[utxoPos].isClaimed = true;
     }
 
--- a/plasma_framework/contracts/poc/fast_exits/Quasar.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Quasar.sol
@@ -254,10 +254,10 @@ contract Quasar {
         claimData[utxoPos].isValid = false;
         Ticket memory ticket = ticketData[utxoPos];
         address token = ticket.token;
-        uint256 bondValue = ticket.bondValue;
+        uint256 ticketBondValue = ticket.bondValue;
         uint256 fundsReserved = ticket.reservedAmount;
         tokenUsableCapacity[token] = tokenUsableCapacity[token].add(fundsReserved);
-        msg.sender.transfer(bondValue);
+        msg.sender.transfer(ticketBondValue);
     }
 
     /**
@@ -278,7 +278,7 @@ contract Quasar {
      * @param output Output Data
      * @param expectedOutputOwner expected owner of the output
     */
-    function verifyOwnership(FungibleTokenOutputModel.Output memory output, address expectedOutputOwner) private view {
+    function verifyOwnership(FungibleTokenOutputModel.Output memory output, address expectedOutputOwner) private pure {
         address outputOwner = PaymentTransactionModel.getOutputOwner(output);
         require(outputOwner == expectedOutputOwner, "Was not called by the Output owner");
     }
--- a/plasma_framework/contracts/poc/fast_exits/Quasar.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Quasar.sol
@@ -4,6 +4,7 @@ pragma experimental ABIEncoderV2;
 import "../../src/framework/PlasmaFramework.sol";
 import "../../src/utils/PosLib.sol";
 import "../../src/utils/Merkle.sol";
+import "../../src/utils/SafeEthTransfer.sol";
 import "../../src/transactions/PaymentTransactionModel.sol";
 import "../../src/transactions/GenericTransaction.sol";
 import "../../src/exits/utils/MoreVpFinalization.sol";
@@ -133,7 +134,7 @@ contract Quasar {
             tokenUsableCapacity[token] = tokenUsableCapacity[token].sub(residualAlmount);
             emit QuasarTotalEthCapacityUpdated(tokenUsableCapacity[address(0x0)]);
         }
-        msg.sender.transfer(amount);
+        SafeEthTransfer.transferRevertOnError(msg.sender, amount, 2300);
     }
 
     ////////////////////////////////////////////	
@@ -241,7 +242,7 @@ contract Quasar {
         uint256 ticketBondValue = ticket.bondValue;
         uint256 fundsReserved = ticket.reservedAmount;
         tokenUsableCapacity[token] = tokenUsableCapacity[token].add(fundsReserved);
-        msg.sender.transfer(ticketBondValue);
+        SafeEthTransfer.transferRevertOnError(msg.sender, ticketBondValue, 2300);
     }
 
     /**
@@ -254,7 +255,7 @@ contract Quasar {
         address payable outputOwner = ticketData[utxoPos].outputOwner;
         uint256 totalAmount = ticketData[utxoPos].reservedAmount.add(ticketData[utxoPos].bondValue);
         claimData[utxoPos].isValid = false;
-        outputOwner.transfer(totalAmount);
+        SafeEthTransfer.transferRevertOnError(outputOwner, totalAmount, 2300);
     }
 
     ////////////////////////////////////////////	
--- a/plasma_framework/contracts/poc/fast_exits/Quasar.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Quasar.sol
@@ -1,0 +1,305 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "../../src/framework/PlasmaFramework.sol";
+import "../../src/utils/PosLib.sol";
+import "../../src/utils/Merkle.sol";
+import "../../src/transactions/PaymentTransactionModel.sol";
+import "../../src/exits/utils/MoreVpFinalization.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
+
+/**
+ * @title Quasar Contract
+ * Implementation Doc - https://github.com/omgnetwork/research-workshop/blob/master/Incognito_fast_withdrawals.md
+*/
+contract Quasar {
+    using SafeERC20 for IERC20;
+    using SafeMath for uint256;
+    using PosLib for PosLib.Position;
+
+    PlasmaFramework public plasmaFramework;
+
+    address public quasarOwner;
+    address public quasarMaintainer;
+    uint256 public safePlasmaBlockNum;
+    uint256 public waitingPeriod;
+    uint256 public bondValue;
+    // bond is added to the reserve only when tickets are flushed, bond is returned every other time
+    uint256 private bondReserve;
+
+    struct Ticket {
+        address payable outputOwner;
+        uint256 validityTimestamp;
+        uint256 reservedAmount;
+        address token;
+        uint256 bondValue;
+        bool isClaimed;
+    }
+
+    struct Claim {
+        bytes rlpClaimTx;
+        uint256 finalizationTimestamp;
+        bool isValid;
+    }
+
+    mapping (address => uint256) public tokenUsableCapacity;
+    mapping (uint256 => Ticket) public ticketData;
+    mapping (uint256 => Claim) private claimData;
+
+    event QuasarTotalEthCapacityUpdated(uint256 balance);
+    
+    modifier onlyQuasarMaintainer() {
+        require(msg.sender == quasarMaintainer, "Only the Quasar Maintainer can invoke this method");
+        _;
+    }
+
+    /**
+     * @dev Constructor, takes params to set up quasar contract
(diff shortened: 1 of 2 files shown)


souradeep-das:
#### Funds flow in/out the contract
##### TokenCapacity

--> addEthCapacity() increases capacity
----> obtainTicket() decreases capacity
------> flushExpiredTicket() can increase the capacity back
------> challengeClaim() can increase the capacity back
else, withdrawal successful, capacity decreased

##### Bond

--> obtainTicket() adds bond value to contract
----> flushExpiredTicket() adds bond to bond reserve (stays in the contract)
----> challengeClaim() returns bond to challenger
----> processClaim() returns bond to output owner

bondReserve is collective amount of bonds that is not blocked by any claim. can be withdrawn

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,285 @@
...
+        require(newSafePlasmaBlockNum > safePlasmaBlockNum, "New limit should be higher than older limit");
+        safePlasmaBlockNum = newSafePlasmaBlockNum;
+    }
+
+    /**
+     * @dev Flush an expired ticket to free up reserved space
+     * @notice Only an unclaimed ticket can be flushed, bond amount is added to bondReserve
+     * @param utxoPos pos of the output, which is the ticket identifier
+    */
+    function flushExpiredTicket(uint256 utxoPos) public {
```
souradeep-das, line 90: Alternatively, we can push tickets to queue, to flush expired tickets together according to timestamp, but guessing there might not be many expired unclaimed tickets because of the bond

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,285 @@
...
+        uint256 fundsReserved = ticket.reservedAmount;
+        tokenUsableCapacity[token] = tokenUsableCapacity[token].add(fundsReserved);
+        msg.sender.transfer(bondValue);
+    }
+
+    /**
+     * @dev Process the Claim to get liquid funds
+     * @param utxoPos pos of the output, which is the ticket identifier
+    */
+    function processClaim(uint256 utxoPos) public {
```
souradeep-das, line 266: Alternatively, we can maintain a queue for processing multiple claims together

souradeep-das (commented):
Also, there is a scope to remove `utxoPos` from the args required for `claim()` and `challengeClaim()`. I have kept it for now for simplicity and readability but this might increase gas.
Any suggestions about these design decisions?

InoMurko (commented):
1337

InoMurko (commented):
I noticed I didn't look at the correct PR.

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+    function flushExpiredTicket(uint256 utxoPos) public {
+        uint256 expiryTimestamp = ticketData[utxoPos].validityTimestamp;
+        require(!ticketData[utxoPos].isClaimed, "The UTXO has already been claimed");
+        require(block.timestamp > expiryTimestamp && expiryTimestamp != 0, "Ticket still valid or doesn't exist");
+
+        uint256 tokenAmount = ticketData[utxoPos].reservedAmount;
+        ticketData[utxoPos].reservedAmount = 0;
+        ticketData[utxoPos].validityTimestamp = 0;
+        tokenUsableCapacity[ticketData[utxoPos].token] = tokenUsableCapacity[ticketData[utxoPos].token].add(tokenAmount);
+        bondReserve = bondReserve.add(ticketData[utxoPos].bondValue);
```
InoMurko, line 108:
```suggestion
        tokenUsableCapacity[ticketData[utxoPos].token] = tokenUsableCapacity[ticketData[utxoPos].token].add(tokenAmount);
        bondReserve = bondReserve.add(ticketData[utxoPos].bondValue); 
```
this is basically returning the reserved amount back to the Quasar and claim the bond?

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+
+        verifyClaimTxCorrectlyFormed(utxoPos, rlpTxToQuasarOwner);
+
+        PosLib.Position memory utxoQuasarOwnerDecoded = PosLib.decode(utxoPosQuasarOwner);
+        require(MoreVpFinalization.isStandardFinalized(
+            plasmaFramework,
+            rlpTxToQuasarOwner,
+            utxoQuasarOwnerDecoded.toStrictTxPos(),
+            txToQuasarOwnerInclusionProof
+        ), "Provided Tx doesn't exist");
```
InoMurko, line 204: If I understand this correctly, the output needs to be transfered to the quasar owner on the actuall plasma chain and it has to be included in the block, correct?

This means, from the UX perspective:
1. initate an exit on plasma framework contracts
2. get ticket from quasar for the exit
3. after the ticket from the quasar, use the same data to make a transfer via Watcher to ChildChain and transfer that output to the quasar owner and wait for a block to be submitted 
4. now claim

is that correct?

is there a way to merge any of these steps?

perhaps:
quasar emits an event that marks the output from alice that wants to exit as spent 
and quasar emits an event that "makes a fake deposit" that basically transfers the ownership of the output to the quasar owner?

maybe that's actually one event emmited. this events would be tracked by the childchain and watcher.

plasma_framework/contracts/poc/fast_exits/Quasar.sol
souradeep-das: Yes! It's just adding to the amount of bonds from expired tickets, that the quasar owner can withdraw.(failed bonds)
I'll redirect you to this for the flow https://github.com/omgnetwork/plasma-contracts/pull/723#issuecomment-737103755

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+        ticketData[utxoPos].reservedAmount = 0;
+        ticketData[utxoPos].validityTimestamp = 0;
+        tokenUsableCapacity[ticketData[utxoPos].token] = tokenUsableCapacity[ticketData[utxoPos].token].add(tokenAmount);
+        bondReserve = bondReserve.add(ticketData[utxoPos].bondValue); 
+    }
+
+    /**
+     * @dev Add Eth Liquid funds to the quasar
+    */
+    function addEthCapacity() public payable onlyQuasarMaintainer() {
```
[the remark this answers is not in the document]
souradeep-das: ohh I'm yet to add ERC20 support. I had it planned in next PRs along with quasar fees

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "../../src/framework/PlasmaFramework.sol";
+import "../../src/utils/PosLib.sol";
+import "../../src/utils/Merkle.sol";
+import "../../src/transactions/PaymentTransactionModel.sol";
+import "../../src/transactions/GenericTransaction.sol";
+import "../../src/exits/utils/MoreVpFinalization.sol";
```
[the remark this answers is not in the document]
souradeep-das: Yes 👍 , these contracts are libraries. more like embedded libraries, and are used here to perform some operations. While the PlasmaFramework contract is a contract and we rely on it as an external contract

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+
+        if (amount <= bondReserve) {
+            bondReserve = bondReserve.sub(amount);
+        } else {
+            uint256 residualAlmount = amount.sub(bondReserve);
+            bondReserve = 0;
+            tokenUsableCapacity[token] = tokenUsableCapacity[token].sub(residualAlmount);
+            emit QuasarTotalEthCapacityUpdated(tokenUsableCapacity[address(0x0)]);
+        }
+        msg.sender.transfer(amount);
```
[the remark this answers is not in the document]
souradeep-das: yeah, I should change this. Thanks for catching!

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+    // Exit procedure	
+    ////////////////////////////////////////////
+    /**
+     * @dev Obtain a ticket from the Quasar
+     * @notice Ticket is valid for four hours, pay bond here for obtaining ticket
+     * @param utxoPos Output that will be spent to the quasar later, is the ticket identifier
+     * @param rlpOutputCreationTx RLP-encoded transaction that created the output
+     * @param outputCreationTxInclusionProof Transaction inclusion proof
+    */
+    function obtainTicket(uint256 utxoPos, bytes memory rlpOutputCreationTx, bytes memory outputCreationTxInclusionProof) public payable {
```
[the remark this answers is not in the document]
souradeep-das: Yeah, I think what we followed in the past is to move away from underscore prefixes unless there is a var name collision, probably because of possible confusion of it being an unused var 😄

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,334 @@
...
+            utxoPosDecoded.toStrictTxPos(),
+            outputCreationTxInclusionProof
+        ), "Provided Tx doesn't exist");
+
+        require(outputData.amount <= tokenUsableCapacity[outputData.token], "Requested amount exceeds the Usable Liqudity");
+        require(outputData.amount != 0, "The reserved amount cannot be zero");
+        require(msg.value == bondValue, "Bond Value incorrect");
+
+        tokenUsableCapacity[outputData.token] = tokenUsableCapacity[outputData.token].sub(outputData.amount);
+        ticketData[utxoPos] = Ticket(msg.sender, block.timestamp.add(14400), outputData.amount, outputData.token, msg.value, rlpOutputCreationTx, false);
```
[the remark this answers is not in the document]
souradeep-das: Sure, we definitely need more events.

plasma_framework/contracts/poc/fast_exits/Quasar.sol
souradeep-das: Yes correct, for the standard way of doing the claim, you have to spend the output to the quasar owner and it has to be included in a block, (there is IFEClaim for tx not included, which is more of a way to recover funds)
The initiate exit on plasma is not required, instead just spending the output 
so the UX will be -
1. get ticket from quasar with an output
2. after the ticket from the quasar, make a transfer via Watcher to ChildChain and spend that same output to the quasar owner and wait for a block to be submitted
3. Now Claim (there is no processClaim if there is no waiting period on quasar, so they can be merged)

Merging some steps on the UX would be interesting to look at. I think looking at it from the core, the 1st step is contract, 2nd is Plasma and 3rd is contract again. I was thinking if 1 and 2 could be one step, doing the second on a success receipt of 1, but still would require two signing?

yeah, these are good thoughts to explore 👍

InoMurko: is it safe to merge step 1 and 2 by emitting an event? so that the childchain and watcher spend the output and there's no need to manually send a transaction from alice to quasar owner?

souradeep-das: umm this is tricky, currently the watcher would expect the tx to be correclty signed by alice after the block is submitted and also if there is a need to start exit on this tx later, it won't happen if it isn't signed by alice, so I think merging the steps from a ux perspective should be good as long as we get alice to sign both times?

souradeep-das: for 3, I feel it should be fine for anyone to call the `claim` func with the parameters. so maybe maybe there is a chance here that we could have a bot automate 3 after 2?

InoMurko: yes absolutely

InoMurko: on the topic of 1 and 2.
what if we make the transaction between alice and quasar - like a deposit event?

plasma_framework/contracts/poc/fast_exits/Quasar.sol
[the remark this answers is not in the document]
InoMurko: now I see your inital comment, sry!

plasma_framework/contracts/poc/fast_exits/Quasar.sol
souradeep-das: interesting! one thing i think will stand in the way is - the need of alice's output to be an input in the tx to the quasar owner in order to verify the claim. in case we mark the output as spent and send a deposit event, we won't have inputs on the deposit tx, and still thinking if there is a way to verify alice spent her output. but worth exploring.
Also to mark alice's output as spent we need a tx that spends that output? since this tx will be used to challenge alice exiting that output. So we would still need a signed tx from alice anyway?

InoMurko:
any particular obstacles if we merge `flushExpiredTicket` (by iterating through the list of tickets and checking  `block.timestamp > expiryTimestamp && expiryTimestamp != 0`)?

souradeep-das:
Oh do you mean to collectively flush all expired tickets? we can totally do that

InoMurko:
> Oh do you mean to collectively flush all expired tickets? we can totally do that

Yeah, but also, maybe we can do that while someone calls `obtainTicket/3`? Or would that be too expensive?

souradeep-das:
ohh so attempting to flush expired tickets from the list everytime a ticket obtained? that's a possibility. Yeah that might be expensive, and we might expect very less tickets to go expired beacuse of the bond ( we should call it `unclaimed` instead of `expired` ticket) only the tickets that weren't claimed will need to be flushed

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+        );
+        require(isSpent, "Spending condition failed");
+    }
+
+    /**
+     * @dev Verify the validity of the ticket
+     * @param utxoPos pos of the output, which is the ticket identifier
+    */
+    function verifyTicketValidityForClaim(uint256 utxoPos) private {
+        require(!ticketData[utxoPos].isClaimed, "Already Claimed");
```
kevsul, line 307:
```suggestion
        require(!ticketData[utxoPos].isClaimed, "Already claimed");
```

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+
+        // first input should be the Utxo with which ticket was obtained
+        require(decodedTx.inputs[0] == bytes32(utxoPos), "Wrong Tx provided");
+
+        // first output should be the utxo for Quasar owner
+        FungibleTokenOutputModel.Output memory outputData
+        = PaymentTransactionModel.getOutput(decodedTx, 0);
+
+        // verify output to Quasar Owner
+        verifyOwnership(outputData, quasarOwner);
```
kevsul, line 330: The error message if this fails doesn't make sense - see comment on `verifyOwnership()` above

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+        // considering fee as a separate input
+        require(ticketData[utxoPos].reservedAmount == outputData.amount, "Wrong Amount Sent to Quasar Owner");
```
kevsul, line 332:
```suggestion
        require(ticketData[utxoPos].reservedAmount == outputData.amount, "Wrong amount sent to quasar owner");
```

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+        require(ticketData[utxoPos].token == outputData.token, "Wrong Token Sent to Quasar Owner");
```
kevsul, line 333:
```suggestion
        require(ticketData[utxoPos].token == outputData.token, "Wrong token sent to quasar owner");
```

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+     * @dev Verify the claim Tx is properly formed
+     * @param utxoPos pos of the output, which is the ticket identifier
+     * @param claimTx the Claim Tx to the quasar owner
+    */
+    function verifyClaimTxCorrectlyFormed(uint256 utxoPos, bytes memory claimTx) private {
+        PaymentTransactionModel.Transaction memory decodedTx
+        = PaymentTransactionModel.decode(claimTx);
+
+        // first input should be the Utxo with which ticket was obtained
+        require(decodedTx.inputs[0] == bytes32(utxoPos), "Wrong Tx provided");
```
kevsul, line 323:
```suggestion
        require(decodedTx.inputs[0] == bytes32(utxoPos), "The claim transaction does not spend the correct output");
```

kevsul (commented):
In general looks good.
I've updated many of the error messages for readability, but feel free to modify my suggestions.

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+    PlasmaFramework public plasmaFramework;
+    SpendingConditionRegistry public spendingConditionRegistry;
+
+    address public quasarOwner;
+    address public quasarMaintainer;
+    uint256 public safePlasmaBlockNum;
+    uint256 public waitingPeriod;
+    uint256 public bondValue;
+    // bond is added to the reserve only when tickets are flushed, bond is returned every other time
+    uint256 private bondReserve;
```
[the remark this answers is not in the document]
souradeep-das: changed to `unclaimedBonds`

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -0,0 +1,335 @@
...
+        require(newSafePlasmaBlockNum > safePlasmaBlockNum, "New limit should be higher than older limit");
+        safePlasmaBlockNum = newSafePlasmaBlockNum;
+    }
+
+    /**
+     * @dev Flush an expired ticket to free up reserved space
+     * @notice Only an unclaimed ticket can be flushed, bond amount is added to bondReserve
+     * @param utxoPos pos of the output, which is the ticket identifier
+    */
+    function flushExpiredTicket(uint256 utxoPos) public {
```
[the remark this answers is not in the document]
souradeep-das: added an event here. should we have a queue instead?

InoMurko (approved):

kevsul (approved):

PR #735, opened 2021-01-15
feat: safe block margin

kevsul:
Changes `safePlasmaBlockNum` to `safeBlockMargin` - the Quasar will not accept exits for outputs younger than the current plasma block minus `safeBlockMargin`.

### DIFF
--- a/plasma_framework/contracts/poc/fast_exits/Quasar.sol
+++ b/plasma_framework/contracts/poc/fast_exits/Quasar.sol
@@ -73,12 +73,12 @@ contract Quasar {
      * @param _bondValue bond to obtain tickets
     */
     constructor (
-      address plasmaFrameworkContract, 
-      address spendingConditionRegistryContract, 
-      address _quasarOwner, 
-      uint256 _safeBlockMargin, 
-      uint256 _waitingPeriod, 
-      uint256 _bondValue
+        address plasmaFrameworkContract, 
+        address spendingConditionRegistryContract, 
+        address _quasarOwner, 
+        uint256 _safeBlockMargin, 
+        uint256 _waitingPeriod, 
+        uint256 _bondValue
     ) public {
         plasmaFramework = PlasmaFramework(plasmaFrameworkContract);
         spendingConditionRegistry = SpendingConditionRegistry(spendingConditionRegistryContract);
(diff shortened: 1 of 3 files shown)


souradeep-das (approved):
looks great!

souradeep-das:
also curious about the tradeoffs between a block margin and timestamp margin?
I definitely also like the block margin more for our use case, since a timestamp margin would force someone to wait. just wondering what you think.

kevsul:
> also curious about the tradeoffs between a block margin and timestamp margin?
> I definitely also like the block margin more for our use case, since a timestamp margin would force someone to wait. just wondering what you think.

My concern with a timestamp margin is that if no blocks are being produced then you could hit a situation where the time has passed but there's no real safety margin at all.

For example, if due to a bug in the child chain, the operator produces an invalid block which also contains Alice's output and then stops producing blocks, then Alice will be able to start a fast exit after the time margin has passed.
With a block margin, she won't be able to do that. 

Maybe that's a contrived example though, because if the operator stops producing blocks, then Alice won't be able to transfer the output to the Quasar owner anyway...


I suppose that difference is that a block margin forces a user to wait until the output to be exited has a certain block height which could take a variable amount of time, whereas a time margin forces the user to wait a fixed amount of time whether or not any blocks are produced in that time.

I can see how a time margin might be easier to understand for the user, but I believe that when dealing with blockchains block height is more secure than time passed.

InoMurko:
```
I can see how a time margin might be easier to understand for the user, but I believe that when dealing with blockchains block height is more secure than time passed.
```

@kevsul is this what you had in mind: https://ethereum.stackexchange.com/questions/413/can-a-contract-safely-rely-on-block-timestamp/428#428

plasma_framework/contracts/poc/fast_exits/Quasar.sol
```diff
@@ -159,7 +174,7 @@ contract Quasar {
 
         PosLib.Position memory utxoPosDecoded = PosLib.decode(utxoPos);
 
-        require(utxoPosDecoded.blockNum <= safePlasmaBlockNum, "The UTXO is from a block later than the safe limit");
+        require(utxoPosDecoded.blockNum <= getLatestSafeBlock(), "The UTXO is from a block later than the safe limit");
```
InoMurko, line 177: this is actually rly cool! good idea

kevsul:
> ```
> I can see how a time margin might be easier to understand for the user, but I believe that when dealing with blockchains block height is more secure than time passed.
> ```
> 
> @kevsul is this what you had in mind: https://ethereum.stackexchange.com/questions/413/can-a-contract-safely-rely-on-block-timestamp/428#428

Yeah, pretty much. Miners put in the block timestamp - there are some restrictions on what they can set it to, but it's best not to trust it too much.

PR #741, opened 2021-01-25
feat: remove waiting period for standard claims

kevsul:


souradeep-das (approved):
Looks good!
What do you think about renaming `claimData`, `challengeClaim()` and `processClaim()` to something like `processIfeClaim()` to show their relation to only ife claims? will that be easier to follow?
I am not really sure if we need it. Whatever you feel is right

souradeep-das (approved):

PR #757, opened 2021-03-18
feat: add quasar migration scripts

souradeep-das:


souradeep-das:
Put this along with the main scripts. If this looks unclear, we could move it to a subdir.

kevsul (approved):
Nice

PR #771, opened 2021-05-27
Kevin/quasar improvements

kevsul:
Some tweaks resulting from audit:
- Remove bondValue from the Ticket struct
- Change functions from public to external where possible
- Add ReentrancyGuard to QuasarPool functions
- Make safeBlockMargin a Timelocked value.
- The contract bytecode is very close to the EIP-170 0x6000 size limit. To be able to make the previous modifications, I've had to reduce the size of the `require` messages


souradeep-das (approved):
Looks wonderful!! 🚀

